PR #3, opened 2016-04-27
OEP-2: Repository Metadata

cpennington:


oeps/oep-2.rst
```diff
@@ -0,0 +1,101 @@
+===================================
+OEP-2: Accordance to Best Practices
```
nedbat: This title is really broad.  Perhaps "Repository metadata"?

oeps/oep-2.rst
```diff
@@ -0,0 +1,101 @@
...
+| Type          | Best Practice                             |
++---------------+-------------------------------------------+
+| Created       | 2016-04-21                                |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+Proposes a standard format for repository metadata to self-report on
+accordance to all Best Practice OEPs.
```
nedbat: This document already covers more than just OEPs, so maybe we shouldn't focus on that.  Conformance to best practice OEPs is just one kind of information in this data.

oeps/oep-2.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Specification
+=============
+
+Each repo in the Open edX system will include a file ``.edx.yml``, with the
```
nedbat: I don't like the name.  First, why should it be a hidden file (I am torn about this myself). Second, "edx" is not the right word. At least "openedx", but maybe just "repo.yml".

oeps/oep-2.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Abstract
+========
+
+Proposes a standard format for repository metadata for Open edX projects.
+
+Specification
+=============
+
+Each repo in the Open edX system will include a file ``openedx.yml``, with the
```
nedbat: "repoinfo.yml"?

oeps/oep-2.rst
```diff
@@ -0,0 +1,100 @@
...
+By providing a standard format for metadata about accordance to Best Practice
+OEPs, tools can be developed in the future to report on the current status of
+all of the Open edX repositories, and how well they follow the Best Practices
+listed in the OEPs.
+
+
+Rationale
+=========
+
+The keys in ``.edx.yml`` were derived from existing repository metadata collected
```
nedbat: Change the file name here once we settle on a name...

oeps/oep-2.rst
```diff
@@ -0,0 +1,100 @@
...
+``area``: string
+    Used to group repositories by in reporting tools. Sample areas: core,
+    mobile analytics.
+
+``oeps``: dictionary
+    A list of keys corresponding to Best Practice OEPs (in the format
+    ``oep-n``). Each value would be one of ``True``, ``False``, ``auto`` or a
+    reason dictionary. ``True`` and ``False`` would indicate accordance (or
+    lack thereof) with the specified OEP, and ``auto`` would specify that the
+    tooling should auto-detect accordance with the OEP. A reason dictionary
```
nedbat: "auto" confuses me.  What kind of tooling are we envisioning? If it could determine it automatically, why does it need to be told to do that?

cpennington: Yeah, I was thinking the same.

oeps/oep-2.rst
cpennington: I feel like this is being specified as part of the Open edX system, and should be specific about that. Just like travis or coveralls are specific about what is instigating the metadata there.

nedbat: It feels like it should have "repo" in the file name, or is that silly?

cpennington: I don't see why it needs "repo" in the name. It makes sense for "repos.yml", because that's talking about a collection of repos, but for this file, it's already implicitly about the repo, because of its location (IMO).

nedbat: Working from the title, "metadata.yml"?

cpennington: Fundamentally, I think this file is Open edX specific, and thus should have that in the name. Once it's there, I'm not sure there's much to be gained from making the name longer.

nedbat: Maybe we can get some opinions from other people.  Ideally, file names would describe what's in the file.

oeps/oep-0002.rst
```diff
@@ -0,0 +1,99 @@
...
+Motivation
+==========
+
+One of the primary goals for Best Practice OEPs is to standardize techniques
+and tools across the Open edX repositories. However, without tooling, it will
+be difficult to move all of the repositories in the same direction (and hard
+to validate which ones have been updated). This OEP exists to fill that gap.
+By providing a standard format for metadata about accordance to Best Practice
+OEPs, tools can be developed in the future to report on the current status of
+all of the Open edX repositories, and how well they follow the Best Practices
```
antoviaque: Will that tool be only for repos on the edX org, or will it includes all the community's repositories? How will the tool discover/track the repositories?

oeps/oep-0002.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    If an OEP isn't listed in the ``oeps`` dictionary, then it is assumed to be
+    ``False``, unless the reporting tool can auto-detect accordance.
+
+
+Motivation
+==========
+
+One of the primary goals for Best Practice OEPs is to standardize techniques
+and tools across the Open edX repositories. However, without tooling, it will
```
antoviaque: How do you define "Open edX repositories", precisely? Does that include XBlocks, for examples? Would there be categories? I imagine not all OEPs would apply to all types of repos?

oeps/oep-0002.rst
cpennington: My intention was for this to cover edx.org repos. However, the tooling will be opensource, and will be reading from a config file, so I'll make sure to make it possible to add new orgs/repos to those configs for folks that want to track the status of non-edx.org repos.

oeps/oep-0002.rst
cpennington: That's an interesting point, and one that I hadn't considered. You're right, there are likely to be some OEPs that cover subsets of repos... I think for now, I'd say that any repo that isn't covered should just use the "rationale" field to explain why, but we may want further refinement down the road to more clearly define the scopes of best-practices.

cpennington:
@efagin: Can you assign an Arbiter for this OEP?

efagin:
@cpennington I can arbit this one. (I'm aware that "arbit" is likely not a verb but it sounded cool.)

oeps/oep-0002.rst
```diff
@@ -0,0 +1,101 @@
...
+Each repo in the Open edX system will include a file ``openedx.yml``, with the
+following fields:
+
+``owner``: string
+    The github user/team responsible for the repository.
+
+``nick``: string
+    A short-name for the repository used by reporting tools.
+
+``area``: string
```
efagin: @cpennington: this might be more flexible as `tags`, which could be a list of strings, and would better handle the case where a repository covers multiple "areas." What do you think?

oeps/oep-0002.rst
```diff
@@ -0,0 +1,101 @@
...
+    ``oep-n``). Each value would be one of ``True``, ``False``, or a
+    reason dictionary. ``True`` and ``False`` would indicate accordance (or
+    lack thereof) with the specified OEP. A reason dictionary
+    must have ``state`` and ``reason`` keys. The ``state`` key is a boolean,
+    and ``reason`` is a string that should explain why it's ok that the
+    ``state`` is ``False`` (this will be displayed in reporting tools).
+
+    If an OEP isn't listed in the ``oeps`` dictionary, then it is assumed to be
+    ``False``, unless the reporting tool can auto-detect accordance.
+
```
efagin: @cpennington it would be helpful to the reader if you included a chunk of sample YAML that shows all of these fields in action, with obviously fake values. Or you could just show snippets for each sub-section, whatever works for you, I just think a few examples would help illustrate the data types and enumerations.

oeps/oep-0002.rst
```diff
@@ -0,0 +1,101 @@
...
+1. Adding a new OEP that few repositories will support, initially. We shouldn't
+   need to update all repositories to detect if they are in accordance, we should
+   be able to assume that they aren't, or automatically detect whether they are.
+2. Repositories may have specific requirements that force them to not implement a
+   best practice. The tools should be able to present that reasoning to anyone
+   looking across repositories, and the reasons should be documented in the repositories
+   themselves.
+3. As much as possible, Best Practices should be autodetected, but because they will
+   often involve a judgement call, autodetection shouldn't be mandatory.
+
```
efagin: I'd love to see a reference implementation, i.e., an example `openedx.yml` file and an example script that can parse said file and generate metrics from it. I think the former should be added on the sooner side (see my above note about YAML snippets), while the latter can emerge once you've gathered some initial feedback on this proposal.

efagin:
@cpennington let's work together to get this OEP merged in and moving along. At this point, I think it's ready to at least move to `Accepted` given the lack of any feedback that would materially impact this proposal.

As noted in OEP-1, we should get some example YAML and a reference implementation in place and ironed out before moving this to Closed but I think we can at least merge it to master at this point.

oeps/oep-0002.rst
cpennington: Agreed.

oeps/oep-0002.rst
cpennington: Will do.

oeps/oep-0002.rst
cpennington: Yeah, I'd like to build up some tools that takes the data we have already, pushes it out into the repos we have, and then collects back the data from those repos.

cpennington:
@efagin: I've resolved all of your suggestions.

efagin:
Looks good, thanks. 👍 from me, this can be marked Approved on Wednesday (pending other feedback) to give the community a few more days to jump in.

efagin:
@cpennington just marked this as Approved. Go ahead and merge into master when ready.

PR #6, opened 2016-06-15
First draft of OEP-3: Asynchronous Task Management

jmbowman:
Not sure the exact level of detail we want, but this should serve as a springboard for further discussion.  In the meantime, I'll dig further into the existing related code and start a tentative implementation.


oeps/oep-0003.rst
```diff
@@ -0,0 +1,144 @@
...
+perspective with a meaningful progress indicator.
+
+For notifications, email is sufficient for the initial implementation (but
+other push notifications may be added later).  A task status page for each
+task will show any available information about how far along the task is, and
+allow retrieval of any output it may have generated.  The user will be
+redirected to this status page when the task is requested.  The initial
+implementation might not include real-time updates of the data on the page or
+a dashboard listing all tasks initiated by the user, but these will be added
+in time.  Similarly, support for tasks which can be monitored by multiple
```
macdiesel: I would like to see us firm up the language here. What can be expected in the initial delivery, what should be in phase 2 ,etc.

oeps/oep-0003.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Celery will be used to run the asynchronous tasks, but the design should be
+capable of adapting to a different task runner later if necessary.  Database
+models will be used to represent tasks as well as their input, output, and
+current status.  It will be possible to present a fan-out of Celery tasks
+instantiated to perform work in parallel as a single task from the user's
+perspective with a meaningful progress indicator.
+
+For notifications, email is sufficient for the initial implementation (but
+other push notifications may be added later).  A task status page for each
```
macdiesel: @efagin, @jmbowman @doctoryes @jbarciauskas I was wondering what other people's opinions are on this? What happens when the user has an out of date email address how will they find this download? Also what about emails going in to the spam filter?

I think it would behove us to implement a tracking page on the first iteration.

oeps/oep-0003.rst
```diff
@@ -0,0 +1,144 @@
...
+in time.  Similarly, support for tasks which can be monitored by multiple
+users with sufficient access priveleges is desired but may be added after the
+initial release.
+
+Course import and export will utilize the new system when it is released, and
+other long-running tasks will switch to use it as needed.
+
+Motivation
+==========
+
```
macdiesel: Should the motivation be at the top?

macdiesel:
Good first pass. @efagin question, should this document contain more technical detail on how this will be implemented? It would be good to see what the data structure would like like and perhaps a layout of the interfaces that will be implemented. 

If this information does not live in this document, where should it live and how should we link the two?

oeps/oep-0003.rst
efagin: @macdiesel I think @jmbowman's suggestion is reasonable as stated: start with e-mail, combined with a single-task status page. The e-mail only acts as an optional heads up that the single-task status page has new information, which the user could also find by just leaving the status page open and periodically refreshing. That seems like a very useful first cut, with future iterations potentially improving the user experience with a "show all my jobs" dashboard and real-time updating/progress bars and such.

Because e-mail is not technically required for a user to do their job here, I think the "out of date email" and "spam filter" use cases are probably just edge cases we leave out for now and possibly revisit them later. There are far more places in our system where we expect that a user's email is up to date and not sending system mail to spam. This could just be my cynicism but fixing the email issues feels like a hard problem to solve with a low return on that investment.

All that said, I don't think e-mail _by itself_ is sufficient for the first iteration, we really ought to have that single-task status page too.

oeps/oep-0003.rst
efagin: Agree with @macdiesel - a brief, but explicit, roadmap would be helpful. I imagine most readers are trying to mentally construct a linear roadmap while reading this paragraph. It's OK if it's just a suggested roadmap. And the sequencing (milestones) is far more important than the timing.

In addition, I'd advise explicitly splitting out a section on "initial implementation" and "not initial implementation" since you bring those points up a few times.

oeps/oep-0003.rst
efagin: Great question for @cpennington, who created this template with sections in the order you're seeing here. I had a similar question while thinking through my own OEP, and agree that some of the underlying "why do you care about this" stuff should bubble up closer to the abstract.

oeps/oep-0003.rst
```diff
@@ -0,0 +1,144 @@
...
+  updated to describe the new workflow.  The task status page should contain
+  enough information to be pretty self-explanatory, though.
+* There will be migrations to add new tables for task tracking.
+* There are a lot of asynchronous tasks already in the
+  `instructor task <https://github.com/edx/edx-platform/tree/master/lms/djangoapps/instructor_task>`_
+  package of the LMS, so particular care should be taken to produce something
+  compatible with that in both code and behavior.  The existing code is local
+  to the LMS and makes some assumptions about what kind of tasks can be run,
+  so it can't just be used as is.
+
```
efagin: @jmbowman I'd add a note that we'll likely need to reimplement services over time to fit the new workflow (is that true?), and work with various partners to update their integrations if it's required. I'd stress that this process is opt-in (we'll move things over as we have time) for each service.

oeps/oep-0003.rst
```diff
@@ -0,0 +1,144 @@
...
+redirected to this status page when the task is requested.  The initial
+implementation might not include real-time updates of the data on the page or
+a dashboard listing all tasks initiated by the user, but these will be added
+in time.  Similarly, support for tasks which can be monitored by multiple
+users with sufficient access priveleges is desired but may be added after the
+initial release.
+
+Course import and export will utilize the new system when it is released, and
+other long-running tasks will switch to use it as needed.
+
```
efagin: @jmbowman Could you add a paragraph or two regarding security and performance? I can imagine that some services might be asynchronous in nature but have a ton of tasks to run (e.g. grading requests for a large course), are there upper limits to the types of async tasks we should consider here? Which users are allowed to view which jobs?

oeps/oep-0003.rst
cpennington: Yeah, moving motivation to the top makes sense to me. Feel like moving the template around, @macdiesel?

cpennington: I fixed this in #8

jmbowman: Good question; I just followed the order from the template.  During the OEP review process, it's one of the most important sections and should probably near the top.  Once the OEP is accepted, I suspect most readers would prefer to get straight into the details of the specification.  @efagin , thoughts?

jmbowman: Gah, never mind, needed to update the page to see the newer comments...

efagin:
👍 merging now. thanks @jmbowman - great work!

nasthagiri:
Question: Any thoughts on allowing the entire course team (and/or others) to view/edit(cancel) tasks created for a course - not just the user who started the task?

jmbowman:
I mentioned that down in the "later enhancements" section:

> Support for tasks which can be monitored by multiple users with sufficient access priveleges. Different permissions could be created for status viewing, generated artifact access, the ability to cancel the task, and the ability to delete the task.

It's something we'll definitely want, and we shouldn't make any design decisions that make it difficult to implement, but discussion so far indicated that it wasn't a requirement for the initial implementation.  (Actually, the implementation I have so far just uses the Django authorization system and leaves it up to the including service to configure the object permission rules appropriately.)

PR #9, opened 2016-06-24
Add "Under review" state and announcement recommendations

jbarciauskas:
Add a state that communicates that an OEP is currently open and currently seeking feedback, and recommend appropriate channels for soliciting this feedback.


jbarciauskas:
@cpennington @efagin @nedbat

efagin:
@jbarciauskas please update the last-modified date of OEP-1 as well; 👍 from me otherwise. This process isn't yet mature enough to warrant a full revisioning of the OEP in my opinion, but in the future we should consider these types of changes to be more significant.

efagin:
Will wait for Cale and Ned to offer feedback before merging.

cpennington:
We should probably update the image of the state diagram, too. I'll share the original source image in lucidchart.

cpennington:
@jbarciauskas: You don't have a lucidchart account, afaict. You should be able to get one from support.

jbarciauskas:
@cpennington @efagin addressed this feedback. Not sure exactly the interaction between draft, under review and deferred but, seems good enough.

efagin:
Thanks, @jbarciauskas. I'm going to merge now. Recommend that authors of outstanding OEPs (@cpennington & @jmbowman) update their proposals to the new "Under Review" state as time permits - there are only three so I'm not concerned.

PR #10, opened 2016-06-24
OEP-4: Application Authorization (scopes)

efagin:
Status: Under Review. Feedback requested. Please see https://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0001.html for context on the process.


efagin:
@ormsbee @cpennington @nasthagiri @jcdyer @clintonb @e0d please review!

oeps/oep-0004.rst
```diff
@@ -0,0 +1,177 @@
...
+  enforcement, etc)
+
+* Access tokens must be reasonably well-contained `JSON Web Tokens`_, signed by
+  the central authorization service, that contain scope information
+
+* Scopes should cover all RESTful web API endpoints [#REST]_ within a service,
+  and may also cover other endpoints as desired by the service team
+
+* Resources should not leak implementation details (such as ``otto_cart``) or,
+  on the other hand, be too broad to be useful (such as ``courseware``).
```
pdpinch: I don't understand the meaning of these parenthetical asides. Are they counter-examples?

oeps/oep-0004.rst
```diff
@@ -0,0 +1,177 @@
...
+  metadata and definitions
+
+* Centralize application and token management but decentralize token
+  enforcement
+
+* Centralize common implementation patterns as we encounter them (logging,
+  enforcement, etc)
+
+* Access tokens must be reasonably well-contained `JSON Web Tokens`_, signed by
+  the central authorization service, that contain scope information
```
pdpinch: Are there alternatives to JSON Web Tokens that you have decided against? Is it helpful to have some of the reasoning here?

oeps/oep-0004.rst
```diff
@@ -0,0 +1,177 @@
...
+* Resources should not leak implementation details (such as ``otto_cart``) or,
+  on the other hand, be too broad to be useful (such as ``courseware``).
+
+.. _`JSON Web Tokens`: https://jwt.io
+
+
+Open questions
+--------------
+* Where, exactly, are we going to maintain the list of all scopes, and which
+  scopes should we start with?
```
pdpinch: While I agree that a central list of all scopes is desirable, I think it would also be good to have scopes included with the API docs presented by DRF. That's the moment when you need to know the scope.

oeps/oep-0004.rst
efagin: @pdpinch the quick version of this is: we're looking to create a separate OEP for the JWT format, it was something "decided" on before we had this framework that probably should be re-captured. I'll propose that I remove the JWT reference here and replace it with less implementation-specific language like `Access tokens may have one or more scopes associated to them by the central authorization service.` That doesn't specify the _option_ to pass that information out as a JWT as an optimization, but also doesn't prevent it from happening. What do you think about that updated language?

oeps/oep-0004.rst
efagin: @pdpinch yeah, sorry, those were meant to be examples of what not to do. I'll clarify.

oeps/oep-0004.rst
efagin: @pdpinch I agree that there should be an additional note that the central scope authority should transparently provide the full scope list and will add this, thanks. I'm not sure what implementation you're suggesting though: 
- returned as data in a real-time API call like `GET /oauth2/v1/scopes`?
- a comment in the DRF response to `/oauth2/v1/access_token` listing the available scopes?
- a github README or read-the-docs page that's easy for developers to reference?
- or something else?

efagin: I added a line item to the Governance section: `The central documented list of scopes must be readily available to application developers and service owners, and kept up to date`. I think that addresses the broader concern.

efagin:
@cpennington point of order, I put `0004` as the value of the `OEP` field in the header table, what you called "the OEP number" in OEP-1. I wasn't entirely sure what to put there - a link? an ID? With what format? Left pad? Etc. It wasn't clear to me initially. I'll update OEP-1 once we've clarified this point.

and @jbarciauskas: I updated the state to `Under Review` as per your update to OEP-1, thanks for doing that!

cpennington:
@efagin: I think I'd be fine with either just the number (`4`) or the short-string (`oep-4`). I don't like the left-padding, because we don't use that in the index of OEPs (and, assuming we can get the sorting to cooperate, I'd rather not use the left-padded version as part of any of the official ids [filesystem notwithstanding])

efagin:
@cpennington feedback addressed.

clintonb:
👍

oeps/oep-0004.rst
ormsbee: @efagin: +1 for removing the JWT reference and moving to the more generic language. I think that the important thing to get across is that the access tokens will be functionally opaque to the clients, and we can separately discuss whether the services accepting that token need to do some lookup to an auth service, interpret the JWT, etc.

efagin: Done. @ormsbee I liked your "functionally opaque" language and borrowed it. Look good?

ormsbee:
👍

oeps/oep-0004.rst
efagin: Got it. I'd like to leave that out of the OEP just because it's an implementation detail, while the overall governance note addresses the fact that we need _something_ for developers to use. I expect long term we'll end up with a run-time way to get relevant scopes, but that feels like something we can work towards over time. The minimum baseline I will suggest for implementation is that the list of relevant scopes lives side by side with the Swagger documentation for the API, so that at least it's machine-parseable and in a well-known spec. And the more we can push it into the runtime environment over time, the better.

oeps/oep-0004.rst
```diff
@@ -0,0 +1,181 @@
...
+* Access tokens may have one or more scopes associated to them by the central
+  authorization service; tokens should be functionally opaque to applications
+
+* Scopes should cover all RESTful web API endpoints [#REST]_ within a service,
+  and may also cover other endpoints as desired by the service team
+
+* Resources should not leak implementation details or internal names, i.e.,
+  instead of ``otto_cart``, consider just ``cart`` or ``shopping_cart``
+
+* Resources should, on the other hand, be too broad to be useful, i.e.,
```
mulby: I think this sentence is missing a "not"

Resources should, on the other hand, *not* be too broad to be useful, i.e.,

efagin: Oops. Fixed, thanks.

PR #12, opened 2016-06-30
OEP-5: Pre-built Developer Environments

cpennington:


aphanse:
👍

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+   is modified.
+7. Open edX Service containers should be built during CI, and published so that any
+   developer can download the latest container. They should also be tagged for
+   each versioned release.
+
+Technology Selection
+--------------------
+
+The prototype implementation was done using `Docker`_ for the containers, and
+`Docker Compose`_ to manage the networking between services.
```
antoviaque: How does that relate/match with the production environment setup? Docker has many specificities, like the requirement to run a single process, or the use of pseudo-virtualization - as a result, it can't use a stock Ubuntu distribution. So does that mean that the devstack environment will differ from the production setup? Or that the production setup will switch to docker?

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+
+Rationale
+=========
+
+Choice of `Docker`_ over `Vagrant`_
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While `Vagrant`_ has worked relatively well for the current monolithic Devstack
+implementation, it would be too resource heavy for running many services simultaneously.
+`Docker`_ has a much lower per-service overhead.
```
antoviaque: Is that really so blocking? With one VM per service, each of the services VMs would have a smaller memory and CPU footprint, not all services would have to be run at the same time, and RAM is pretty cheap nowadays. I would be interested in looking at the actual figures - do we know what resources it would actually take in both cases? I assume you are comparing it to Vagrant multi-machine setups, ie https://www.vagrantup.com/docs/multi-machine/ ? Converting to docker has a high development time price tag, in addition to having to deal with its drawbacks mentioned above (single process orientation & pseudo virtualization).

cpennington: I'm not sure that converting to docker has a significantly higher development time than converting to vagrant multi-machine would. The biggest difficulties are just in the setup of devstack to run as multiple services with network boundaries, rather than docker specific issues (I believe).

As far as resource usage goes, I think the advantage that Docker has in a development environment is that, unlike a pure virtualization solution like virtualbox, Docker *doesn't* have resource allocation guarantees, which means that the all of the services can share the resources on the host box, rather than having to explicitly parcel them out to each of the VMs. That said, I haven't done an apples-to-apples comparison on an edx service in particular to see how the strategies compare.

oeps/oep-0005.rst
cpennington: Can you point to the places where the Docker guest can't use a stock Ubuntu image? I hadn't heard that (although I may simply have missed it). Even if it's not a stock image, I think there's still a question of whether the changes to make it compatible with Docker matter for our development environments.

I know that @e0d and @efagin have expressed interest in seeing the production stack move to Docker. However, that's not a specific goal of this OEP.

Re: single process: my understanding is that Docker requires you to have a single entry-point for container, but not a single process. So, we can `supervisor` to run a number of processes, for instance (which I believe parallels our current production usage).

antoviaque: @cpennington For the stock ubuntu image - the fact that it needs to use a different init process is already a difference with the stock image. Then there are also the stock components that don't run correctly with pseudo virtualization - I know syslog has been a long running pain point in this, although I do not know if that's still the case with the latest Ubuntu releases.

You could use supervisor to run a number of processes - but you still lose the ability to launch processes the standard way, and not everything is run through supervisor. nginx comes to mind as an example.

oeps/oep-0005.rst
antoviaque: > I'm not sure that converting to docker has a significantly higher development time than converting to vagrant multi-machine would.

I used to think that - but if my past experience is any indication, pseudo-virtualization quirks do tend to make things more complicated than using a native VM.

> The biggest difficulties are just in the setup of devstack to run as multiple services with network boundaries, rather than docker specific issues (I believe).

I hope you're right : )

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+Motivation
+==========
+
+Open edX and the edx-platform repository have gone through several iterations
+of developer setup instructions.
+
+The first set of instructions detailed how to install edx-platform and all of
+its dependencies on your local workstation. As the number of developers and
+developer environments increased, so did the workload of maintaining instructions
+to work around all of the ideosyncacies in those environments. These instructions
```
jzoldak: typo ideosyncacies

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+3. The monolithic `Vagrant`_ image isn't particularly representative of production
+   environments, because all of the services are installed on the same machine,
+   which makes it harder for developers to be confident that their changes will
+   work when deployed.
+4. The shared services dependencies on things like MySQL and Elasticsearch mean
+   that all of the services in Devstack have to be upgraded at the same time,
+   or have to maintain compatability with multiple database versions until all
+   services are ready to update. This hinders independent development on the
+   services.
+5. Because all of the services are installed on the same image, the image eiher
```
jzoldak: s/eiher/either

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+1. Services that aren't part of the Open edX ecosystem must be run using
+   their own published containers (where available).
+2. Open edX containers must run all of their processes via `supervisor`_ to
+   provide a single entrypoint even when the container has multiple processes
+   (such as `nginx`_ and a `Django`_ service running under `gunicorn`_).
+3. Open edX services running in containers must log to ``stdout`` and ``stderr``
+   so that those logs can be captured by `supervisor`_.
+4. Dependent services (such as databases) should be configured to mount their
+   storage directories in ``.devstack/`` inside the source code directory (this
+   should be added to `.gitignore` and `.dockerignore`_. This will allow data from
```
jzoldak: missing close )

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+   ``devstack`` to be persist across container executions.
```
jzoldak: to persist

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+Interface
+~~~~~~~~~
+
+The primary developer interaction with ``devstack`` will be through ``make``
+shortcuts. The following targets will be standard::
+
+    make devstack.watch  # Start the devstack services (if they aren't already)
+                         # and display their logs on stdout
+
+    make devstack.ssh    # Start the devstack services (if they aren't already)
```
jzoldak: Because I'm not familiar with how you would do this in Docker, I wanted to double-check that it would be possible to ssh from the host system into the container. I'm thinking specifically of the use case of remote debugging (with a remote interpreter) in PyCharm.

cpennington: Actually, looking back at this, `ssh` is the wrong term. Basically, you can run another process in the same container. That process can be `/bin/bash`, which gives you a shell (much like sshing into a VM). From that shell, you can interact with the other processes that are running in the container.

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+Rationale
+=========
+
+Choice of `Docker`_ over `Vagrant`_
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While `Vagrant`_ has worked relatively well for the current monolithic Devstack
+implementation, it would be too resource heavy for running many services simultaneously.
+`Docker`_ has a much lower per-service overhead.
+
```
jzoldak: If a developer already has a docker environment installed and running, would it be faster to upgrade to latest than it currently is with Vagrant? 
Do you only need to download the new changes to layer on top of what you have, rather than with Vagrant where you need to download a whole new huge .box file. That would be a huge advantage, especially for people with slow internet connections.

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+newcomers working in that system, each service will provide a prebuilt environment
+that is packaged, published, and ready to be used for development.
+
+Motivation
+==========
+
+Open edX and the edx-platform repository have gone through several iterations
+of developer setup instructions.
+
+The first set of instructions detailed how to install edx-platform and all of
```
benpatterson: Not sure if you want to include too much historical context...

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+the infrastructure that was built out for installing Open edX in production.
+However, there are several problems with the current monolithic `Vagrant`_ image
+which we would like to address:
+
+1. As new services are added to the Open edX ecosystem, the `Vagrant`_ image
+   grows in size, requiring more system memory from the host machine.
+2. `Vagrant`_ imposes noticable overhead during filesystem access, which especially
+   hurts any development that it doing file-heavy operations (such as manipulating
+   static assets).
+3. The monolithic `Vagrant`_ image isn't particularly representative of production
```
benpatterson: I think this is a bit of an overstatement. My understanding is that it is representative of production based on system requirements, but notsomuch on system footprint.

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+   environments, because all of the services are installed on the same machine,
+   which makes it harder for developers to be confident that their changes will
+   work when deployed.
+4. The shared services dependencies on things like MySQL and Elasticsearch mean
+   that all of the services in Devstack have to be upgraded at the same time,
+   or have to maintain compatability with multiple database versions until all
+   services are ready to update. This hinders independent development on the
```
benpatterson: plus-1 on that

oeps/oep-0005.rst
```diff
@@ -0,0 +1,182 @@
...
+The prototype implementation was done using `Docker`_ for the containers, and
+`Docker Compose`_ to manage the networking between services.
+
+Rationale
+=========
+
+Choice of `Docker`_ over `Vagrant`_
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While `Vagrant`_ has worked relatively well for the current monolithic Devstack
```
benpatterson: I'd personally go light on the `vs` discussion and just point out what we're trying to get out of docker (e.g., taking advantage of docker's _x_ or _y_ as opposed to _docker is better because of x or y_)

benpatterson:
I'd like to see discussion of the test capabilities. For example, reading this `OEP`, I get the sense that I'll have something that is production-like, but I don't get the sense it'll contain everything I need to actually run tests. I don't think that's what you want (right?), but I think it's a must for any given development environment.

I also suggest you have some info about a rollout plan. You do point out the first case, which is great, but I'd like to know what you propose for platform in particular, as the most complicated case, or whether or not we include things like the analytics devstack in this work.

oeps/oep-0005.rst
cpennington: I think having it in there, as a way of describing the things that haven't worked in the past, is useful.

oeps/oep-0005.rst
cpennington: Yeah, I'll qualify the concern a bit more.

cpennington:
@benpatterson: I've added something specific about making sure that the devstack can run tests when it's spun up.

I don't think rollout plan is appropriate for the OEP. The idea with OEPs is that they are essentially the description of our target state, and then we can measure as we go along which IDAs have implemented it.

oeps/oep-0005.rst
```diff
@@ -84,10 +84,10 @@ shortcuts. The following targets will be standard::
     make devstack.watch  # Start the devstack services (if they aren't already)
                          # and display their logs on stdout
 
-    make devstack.ssh    # Start the devstack services (if they aren't already)
-                         # and ssh into the primary service (as the user that
-                         # owns the service, in the directory that has the
-                         # service source code)
+    make devstack.shell  # Start the devstack services (if they aren't already)
+                         # and access a shell on into the primary container
```
benpatterson: nit: s/on into/on

oeps/oep-0005.rst
```diff
@@ -131,18 +133,21 @@ The prototype implementation was done using `Docker`_ for the containers, and
 Rationale
 =========
 
-Choice of `Docker`_ over `Vagrant`_
```
benpatterson: ty

benpatterson:
@cpennington thanks for addressing my comments.

Ah...well, I think you are right about rollout strategy not-being germane to the OEP. But, where will it live? I'd like to have a clear rollout strategy with milestones to shoot for. Could you put that together somewhere else and let folks know about it offline, then?

cpennington:
I think https://openedx.atlassian.net/browse/EV-19 will be the place to coordinate the build-plan.

jzoldak:
👍

oeps/oep-0005.rst
```diff
@@ -0,0 +1,187 @@
...
+of developer setup instructions.
+
+The first set of instructions detailed how to install edx-platform and all of
+its dependencies on your local workstation. As the number of developers and
+developer environments increased, so did the workload of maintaining instructions
+to work around all of the idiosyncrasies in those environments. These instructions
+weren't useful for production installations, which were standardized on a
+particular release of the Ubuntu operating system.
+
+The second, current, method of development is referred to as Devstack. This is
```
stvstnfrd: @cpennington  While I'm generally +1 on this proposal, I'm -2 on the naming convention.

Here, you remind us that "devstack" means something very specific in the OpenEdX context;
it refers to the current, Vagrant-based virtual machine.. 

To redefine this term seems needlessly ripe for confusion.
If we're going to advocate a new development pattern, it needs to have a different name.

I can already see new developers trying to troubleshoot this:
1. $devstack breaks
2. $user googles "fix openedx devstack"
3. $user wades through _years_ of edx-code posts, read-the-docs, and Confluence wiki entries that describe an unrelated, deprecated process.
4. $user gets lucky and realizes "oh not _that_ devstack, the _other_ devstack"
5. wtf

cpennington: That's an excellent point. I'll try and figure out a better name. Do you have any suggestions?

stvstnfrd: Sometimes its beneficial to have the high-level name not be directly
tied to the specifc implementation. In this case, the underlying
implemenation is kind of the point (docker vs vagrant), so I'm not
opposed to a name that references the supporting tech.

The first thing that pops to mind is `dockstack`.
`dock` implies a connection to docker.
`stack` implies that it's a collection of services.

It rolls off the tongue in a similar manner as `devstack` (two
syllables, same leading consonant on the prefix, etc.),
though maybe it's too similar.

That's just off-the-cuff; I'll put some more thought into it now too.

bradenmacdonald: +1 for a different name; that is an excellent point. "Dockstack" is alright, but if more people start using docker in production I can see potential for confusion if people use "dockstack" to refer to prod stacks that use docker. A name that implies a developer-optimized environment or a generic name like "Evergreen" would be good, I think.

oeps/oep-0005.rst
```diff
@@ -0,0 +1,187 @@
...
+                         # and run `make *` from the code directory of the service
+                         # (as the service owner)
+
+Implementation
+~~~~~~~~~~~~~~
+
+Requirements
+------------
+
+1. Services that aren't part of the Open edX ecosystem must be run using
```
e0d: I think this should be rephrases to something like, "Third-party services like MySQL and MongoDB, should be run using vendor provided official containers.  If no reasonable alternative exists to customizing a provided container, the minimal changes should be applied as a layer on the provided container."

oeps/oep-0005.rst
```diff
@@ -0,0 +1,187 @@
...
+                         # and access a shell on the primary container
+                         # (as the user that owns the service, in the directory
+                         # that has the service source code)
+
+    make devstack.clean  # Stop the devstack services and remove any intermediate
+                         # data that they stored in the .devstack/ directory
+
+    make devstack.*      # Start the devstack services (if they aren't already)
+                         # and run `make *` from the code directory of the service
+                         # (as the service owner)
```
mulby: This is a little unclear... if I have an "insights" service which depends on an "analyticsapi" service and an "lms" service and a "forums" service that I'm not working with/on at the moment, how exactly would this work?

Do these make commands only operate on *all* services? Ideally I'd like to be able to specify a group of services that I care about at the moment, and easily stand them up and tear them down without having to fiddle with services that I don't care about.

Maybe we could have a git ignored yaml file that specifies the services I care about at the moment:

```yaml
services:
  - lms
  - insights
  - analyticsapi
```

Or some kind of parameter that lets me choose:

```bash
SERVICES="lms,insights,analyticsapi" make devstack.watch
```

This may be too far into the weeds for this OEP - do we need to specify this interface here? Or could we simply define the objectives of the interface?

oeps/oep-0005.rst
```diff
@@ -0,0 +1,187 @@
...
+
+1. Services that aren't part of the Open edX ecosystem must be run using
+   their own published containers (where available).
+2. Open edX containers must run all of their processes via `supervisor`_ to
+   provide a single entrypoint even when the container has multiple processes
+   (such as `nginx`_ and a `Django`_ service running under `gunicorn`_).
+3. Open edX services running in containers must log to ``stdout`` and ``stderr``
+   so that those logs can be captured by `supervisor`_.
+4. Dependent services (such as databases) should be configured to mount their
+   storage directories in ``.devstack/`` inside the source code directory (this
```
e0d: This feels unnecessarily prescriptive.  I think that that the convention for all persistent stores should be defined under a base directory, that directory should have a default location, that location should be configurable.  Personally, I'd like that location to be outside of the git working tree and personally I don't like defaulting to a hidden directory.

oeps/oep-0005.rst
efagin: I really like @e0d's phrasing here and think it's a good practice in general (like, minimize customizations on first-party containers too)

oeps/oep-0005.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Implementation
+~~~~~~~~~~~~~~
+
+Requirements
+------------
+
+1. Services that aren't part of the Open edX ecosystem must be run using
+   their own published containers (where available).
+2. Open edX containers must run all of their processes via `supervisor`_ to
```
mulby: Do we need additional guidance here?

* Should I run all of the hadoop services in a single container, or split them into separate ones?
* Can I run my open edX service and the hadoop services in a single container?
* Do I have to start the hadoop services using supervisor?
* What about custom containers for third party applications that don't have a good container already built that we can use off the shelf?

oeps/oep-0005.rst
cpennington: My reasoning for putting it in the source directory was that I don't want the data from mysql containers being used by different instances of devstack (say, edx-platform and ecommerce) to get mixed together. I agree that configurability would be good, though.

oeps/oep-0005.rst
cpennington: > Should I run all of the hadoop services in a single container, or split them into separate ones?

What would you suggest? When you say hadoop services, do you mean the infrastructure for running hadoop? Or your particular hadoop jobs? Is there an existing standard practice for running hadoop in docker?

> Can I run my open edX service and the hadoop services in a single container?

I probably wouldn't, since it would make that container bigger, and would force it to be rebuilt if either the hadoop stuff or the edX stuff changed. Again, though, you know hadoop better than I do. What would you recommend?
    Do I have to start the hadoop services using supervisor?
    What about custom containers for third party applications that don't have a good container already built that we can use off the shelf?

oeps/oep-0005.rst
cpennington: The docker-compose v2 format allows you to specify dependencies for services, and then to bring up either all services specified in the file, or just a specific service and all of its dependencies. So, I think for your case, I'd have those three services as separate top level services in the compose file, and then if you just want `analyticsapi`, then `docker compose up analyticsapi` would start just that.

Going back to the question of "all services". In general, these commands are all targeted at _the service whose repo we're in_. So, in the `analyticsapi` repo, that's `make devstack.shell` would drop you into the `analyticsapi` service.

I haven't decided on a standard practice for multi-service repos (except maybe just "Don't do that") However, I could see standardizing on something like "make devstack.shell.analyticsapi" (just using the service name as a suffix).

oeps/oep-0005.rst
cpennington: Yes, agreed.

oeps/oep-0005.rst
mulby: I was mostly just bringing up the types of questions that I don't see guidance for in the OEP. Happy to make whatever decisions I want for these various things, just wanted to bringing up some of the decision points I've had to face in the past to see if any were worth explicitly providing guidance for.

Also is it worth calling out that we expect the containers to be built for development purposes. If they happen to work in production as well, that's a bonus, but it's a secondary objective and if any mutually exclusive decisions need to be made, we should choose the development friendly option. That is, we shouldn't prevent commits of Dockerfiles that won't work well in production. If someone wants to modify them in the future or invest extra effort to make them also work in prod, more power to them as long as they don't sacrifice our developer-facing goals.

oeps/oep-0005.rst
mulby: Ah, I misunderstood the proposal (carrying too much baggage from the current stuff in edx/configuration). It may be worth adding more detail here to clarify that these make targets would exist in each service's core repo, not in some centralized repo with a bunch of Dockerfiles in it or the edx/configuration repo.

oeps/oep-0005.rst
e0d: @antoviaque 

We've endeavored to find an abstraction that supports multiple virtualization targets in ansible using he vhost and container roles -- which, parenthetically, was useful for the recent open stack changes. 

The abstraction is by-no-means perfect and it doesn't solve for the fact that some, important, differences will be in play comparing vagrant to aws to docker.

I'm not convinced that the mechanism for starting supervisord is likely to be a big source of risk.  I take the point that not every service is managed by supervisor, but I think we can manage that by moving more things to supervisor, adopting 12 factor principles when sensible and minimizing customizations to "vendor" services.

To @cpennington's point, we can and do run multiple processes on Docker containers, say nginx, rsyslog and gunicorn, though I think that we should attempt, over time, to limit that.

I don't think we will have complete symmetry between various virtualization targets, but the value of allowing deployers to choose their target is high enough, that navigating that risk is worthwhile.

Internally, we think using Docker in dev environments will increase flexibility, increase operational awareness and decrease surprise.  Still, there remain thorny problems to solve.

antoviaque: > We've endeavored to find an abstraction that supports multiple virtualization targets in ansible using he vhost and container roles -- which, parenthetically, was useful for the recent open stack changes.
> The abstraction is by-no-means perfect and it doesn't solve for the fact that some, important, differences will be in play comparing vagrant to aws to docker.

Yup, I agree that having multiple virtualization targets is a good thing, as it helps making the infrastructure less tied to a specific service (and specifically AWS, which is proprietary). The part I'm objecting to, it's to extend this support to pseudo-virtualization, which comes with a larger set of quirks of its own, and breaks assumptions that are common to the other full virtualization approaches.

But I also recognize that I'm in the minority here - so I mainly intended to mention my reservations, not to block the initiative.

> I'm not convinced that the mechanism for starting supervisord is likely to be a big source of risk. 

Yes, starting supervisord shouldn't be an issue - it's actually a point of similarity with the full virtualization approaches, which already use supervisord to run the services. (Although I always wondered why we weren't using the standard init scripts for this?)

> I take the point that not every service is managed by supervisor, but I think we can manage that by moving more things to supervisor, adopting 12 factor principles when sensible and minimizing customizations to "vendor" services.

That's still a significant overhead, no? And it also means we can't easily reuse standard ansible galaxy roles for standard services.

> To @cpennington's point, we can and do run multiple processes on Docker containers, say nginx, rsyslog and gunicorn, though I think that we should attempt, over time, to limit that.

You do that via supervisord?

oeps/oep-0005.rst
cpennington: I've added some notes around the latter concern. I'm not sure how best to provide guidance around the types of questions you had for hadoop.

oeps/oep-0005.rst
cpennington: "Locstack"? "Local"? Some translation of "local" into a different language?

cpennington:
I changed most references to `devstack` to just be "local development environment", and changed the make targets to just reference `make dev.*`.

oeps/oep-0005.rst
```diff
@@ -0,0 +1,263 @@
...
+
+   **Rationale**: We should be aiming to minimize the amount of work needed to
+   support our development environments. Minimizing changes to third-party
+   containers means we have less code/configuration that we need to update as
+   third-party containers and applications change. Minimizing customization
+   should also keep push us in the direction of making it easy to deploy our
+   stack, because there will be fewer customizations needed by operators of
+   Open edX in production.
+
+2. Open edX containers must run all of their processes via `supervisor`_ to
```
gsong: This should be the exception, not the rule. Refer to [Docker best practice](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/run-only-one-process-per-container). For development purposes especially, I'm not sure what the benefits of using Supervisor would be, it feels like another barrier to me.

In the example cited, we would just run nginx and gunicorn in their own separate containers, right?

cpennington: No. The goal of OEP-5 is to make the development environment run faster while still mirroring our production deployment as much as is reasonable. In production, nginx is deployed to each of the webhosts. As such, we should be deploying it to the same container in development mode, so that it's easy to validate that all of the managed configuration is correct.

If, in the future, we get to a point where nginx is deployed on a separate host, then I would recommend we move it to a second container.

gsong: Do we use Supervisor to manage the nginx process in production as well?

PR #14, opened 2016-07-25
Add scope description governance guidelines.

jcdyer:
## Reviewers 

- [ ] @efagin 
- [ ] Other

FYI: @nasthagiri @e0d @clintonb

### DIFF
--- a/oeps/oep-0004.rst
+++ b/oeps/oep-0004.rst
@@ -121,6 +121,11 @@ Governance
 * Resources should not, on the other hand, be too broad to be useful, i.e.,
   instead of ``courseware``, consider ``discussions`` or ``catalog`` or
   whatever specific resource is of interest
+ 
+* Scope descriptions need to be presented to the user by the token management
+  system.  They should therefore be maintained with the central list of scopes.
+  Scope descriptions should be internationalized, but should be presented with
+  the scope name, which is consistent across translations.
 
 Open questions
 --------------
--- a/oeps/oep-0004.rst
+++ b/oeps/oep-0004.rst
@@ -127,6 +127,16 @@ Governance
   Scope descriptions should be internationalized, but should be presented with
   the scope name, which is consistent across translations.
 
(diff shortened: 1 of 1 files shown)


oeps/oep-0004.rst
```diff
@@ -121,6 +121,21 @@ Governance
...
+
+Alt:
+
+* Scope metadata should be maintained centrally with the canonical list of 
+  scopes.
+
+* Scope descriptions should be internationalized strings, managed by Transifex.
+
+* Scopes should be presented to the user during authorization as a combination 
+  of the canonical name and the internationalized description.
```
jcdyer: The first bulletpoint is more specifically focused on descriptions, rather than metadata, and provides more rationale.  Reading the rest of the OEP, I'm not sure that belongs here.  The alternate version is more prescriptive, and is generally applicable to metadata.  I'm leaning toward the alternate version, but interested in feedback.

jcdyer:
Note to self: If we decide to go with the alternate text, we can delete the third bullet from Open questions.

oeps/oep-0004.rst
efagin: I also prefer the alternate version. Keep an eye on this line in Governance, in the currently live version: `Centralize the complete list of scopes but decentralize the associated metadata and definitions`. You're proposing that we significantly alter this particular governance point with `Scope metadata should be maintained centrally...`. You should consider, in this PR, modifying that bullet - and consider whether there's a way to decentralize these descriptions or if centralizing them makes the most sense.

efagin: @jcdyer are you planning on finishing this PR or should I have somebody else pick it up?

efagin:
@jcdyer given that this was a set of suggestions and alternates, and is long dormant, I'm going to close this PR. Let's open up a new one once we've got a clear path forward. (This edit is minor enough that it should be easy to re-create when we're ready.)

PR #15, opened 2016-08-04
OEP-6: Context-scoped XBlock Fields

bradenmacdonald:
This is an OEP for context-scoped XBlock fields, also called "course-scoped XBlock fields."

@efagin can you please check this over and assign an arbiter?

### DIFF
--- a/oeps/oep-0006.rst
+++ b/oeps/oep-0006.rst
@@ -1,0 +1,429 @@
+===================================
+OEP-6: Context-scoped XBlock Fields
+===================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-6 </oeps/oep-0006>`             |
++---------------+-------------------------------------------+
+| Title         | Context-scoped XBlock Fields              |
++---------------+-------------------------------------------+
+| Last Modified | 2016-12-14                                |
++---------------+-------------------------------------------+
+| Author        | Braden MacDonald <braden@opencraft.com>   |
++---------------+-------------------------------------------+
+| Arbiter       |                                           |
++---------------+-------------------------------------------+
+| Status        | Accepted                                  |
++---------------+-------------------------------------------+
+| Type          | Architecture                              |
++---------------+-------------------------------------------+
+|  Created      | 2016-08-03                                |
++---------------+-------------------------------------------+
+| `Resolution`  |                                           |
++---------------+-------------------------------------------+
+
+
+
+Abstract
+========
+
+There is currently no standard way for XBlocks to define data fields such that
+content, settings, or user input can be shared among multiple XBlocks within a
+particular "context" such as a course. Such "course-scoped" or "context-scoped"
+fields are a common requirement; for example, an author may wish to specify the
+default "number of attempts allowed" once for all problems in a course, or a
+video player XBlock may want to "remember" the user's preferred video playback
+speed for videos within a particular course.
+
+This OEP proposes a new mechanism for defining XBlock fields that are explicitly
+scoped to a particular context. The generic term "context" refers to a course,
+content library, blog post, or any other place where XBlocks can be used (see
+"Rationale").
+
+
+
+Background
+==========
+
+The XBlock API uses the abstract notion of "`scopes`_" to associate each field
+with particular environment(s) and user(s). A field's ``Scope`` is generally a
+composite property, defined as the combination of a ``BlockScope``, a
+``UserScope``, and a ``name``.
+
+.. _scopes: http://edx.readthedocs.io/projects/xblock/en/latest/fields.html
+
+For reference, the ``UserScope`` can have three values (no changes to these are
+being proposed):
+
+* ``NONE``: data not related any particular user, that does not change while
+  users interact with the XBlock
+* ``ALL``: data not related to any particular user, but which may be modified
+  during user interactions
+* ``ONE``: data related to an individual user
+
+The ``BlockScope`` can have the following values:
+
+* ``USAGE``: "The data is related to a particular use of a block in a course."
+* ``DEFINITION``: "The data is related to the definition of the block" (one
+block definition can be used multiple times in a course or even among multiple
+courses).
+* ``TYPE``: "The data is related to all instances of this type of XBlock."
+* ``ALL``: "The data is common to all blocks."
+
+XBlock runtimes would generally not be expected to implement every possible
+combination of ``BlockScope`` and ``UserScope``. Instead, certain "named scopes"
+are defined, which are the only scopes that most runtimes support:
+
+The existing named scopes, and an example use case for each, are as follows:
+
+* ``Scope.content`` (UserScope.NONE, BlockScope.DEFINITION):
+
+  * Example: The text of a multiple choice question
+* ``Scope.settings`` (UserScope.NONE, BlockScope.USAGE)
+
+  * Example: The number of attempts allowed for a problem
+* ``Scope.user_state`` (UserScope.ONE, BlockScope.USAGE)
+
+  * Example: The student's answer to a multiple choice question
+* ``Scope.user_state_summary`` (UserScope.ALL, BlockScope.USAGE)
+
+  * Example: The number of students who have answered the question (side note:
+    there is no locking or atomicity for XBlock field writes, so in practice
+    this field is currently of little use in a MOOC)
+* ``Scope.user_info`` (UserScope.ONE, BlockScope.ALL)
+
+  * Example: user age (note: this scope seems little-used and overlaps with the
+    user service)
+* ``Scope.preferences`` (UserScope.ONE, BlockScope.TYPE)
+
+  * Example: (For a video XBlock) Boolean setting to prefer YouTube or
+    non-YouTube video playback.
+
+
+
+Motivation
+==========
+
+As one would expect, ``Scope.user_info`` and ``Scope.preferences`` fields in the
+Open edX platform are implemented globally, and not scoped to the course that
+uses them. There is currently no standard, extensible way to use XBlock fields
+to store data that is scoped to a particular "context," such as a course,
+although this is a common need.
+
+In the absence of a formal method for achieving context-scoped fields through
+the XBlock API, many other solutions have been put in place, such as: adding
+custom Django models to XBlocks, storing fields on the root "course" XBlock,
+defining inherited fields using InheritanceMixin, using edx-submissions as a
+generic data store, and more.
+
+It is hoped that the new canonical mechanism described by this OEP can provide
+consistency, improve interoperability, allow new types of behavior to be
+implemented by XBlocks, and give both authors and learners additional control
+over learner interactions with XBlocks.
+
+
+
+Specification
+=============
+
+New BlockScope value: CONTEXT
+-----------------------------
+
+To support context-scoped fields, ``BlockScope`` will allow a new value:
+``CONTEXT``. This new ``BlockScope`` will denote data that is related to all
+instances of this type of XBlock within a particular context.
+
+A "context" is defined by the runtime but is usually a "course" or the nearest
+equivalent. In the Open edX platform, the context may be a course or a content
+library. Different runs of the same course would be considered different
+contexts and would not share data, though UserScope.NONE fields would be copied
+when creating a new course run. On other platforms, a "context" could be defined
+as a blog post or a textbook.
+
+New named scopes
+----------------
+
+As is the situtation today, XBlock runtimes would generally not be expected to
+implement every possible combination of ``BlockScope`` and ``UserScope``.
+Instead, XBlock runtimes are encouraged to implement these new named scopes,
+which would be defined in the XBlock python package:
+
+* ``Scope.context_settings`` (UserScope.NONE, BlockScope.CONTEXT)
+
+  * Example: For an LTI XBlock, a list of the LTI provider URLs and secrets used
+    in the course
+  * Example: For a discussions XBlock: Enable or disable discussions in this
+    course, as well as settings like "allow anonymous discussion posts", "allow
+    anonymous discussion posts to peers", "discussion blackout dates", etc.
+
+* ``Scope.context_preferences`` (UserScope.ONE, BlockScope.CONTEXT)
+
+  * To avoid performance issues, runtimes are not required to support field
+    values larger than 4 KiB for fields that use this scope. This restriction is
+    intended to avoid the temptation to use this type of XBlock field for
+    logging or message passing. Such uses could cause data loss or performance
+    concerns, because the XBlock API does not support appends nor locking for
+    field data.
+  * Example: Enable or disable notifications about new discussion posts in this
+    course
+  * Example: Video playback speed for this course
+  * Example: An XBlock in which students "accumulate building blocks over the
+    course of the semester" (`PLAT-325`_)
+
+.. _PLAT-325: https://openedx.atlassian.net/browse/PLAT-325
+
+New XBlock view: context_settings_view
+--------------------------------------
+
+An XBlock which uses ``Scope.context_settings`` fields will need a way for
+content authors to edit the data in those fields. (e.g. when creating a course,
+the course author may wish to change course-wide settings for the discussion
+XBlock used throughout that course.)
+
+A new XBlock view, ``context_settings_view`` shall be used for this purpose.
+Just like the existing `studio_view` defined by edX Studio, this new view will
+allow an XBlock to return an arbitrary HTML Fragment which contains a form or
+other UI for editing its ``Scope.context_settings`` fields.
+
+It is expected (but not required) that runtimes like edX Studio which want to
+support editing of context-scoped fields will define a list of XBlocks that are
+enabled within a particular context/course, and will display a list of XBlocks
+that define the ``context_settings_view``. Clicking on the name of that XBlock
+would display the HTML fragment and allow the author to edit the context-scoped
+fields.
+
+.. _studio_view: http://edx.readthedocs.io/projects/xblock-tutorial/en/latest/edx_platform/edx_studio.html
+
+Context-scoped fields in OLX (Provisional)
+------------------------------------------
+
+When serializing a course with ``Scope.context_settings`` fields to Open
+Learning XML (OLX), any context-scoped fields could be described within a new
+``<xblock-settings>`` XML element that is a child of the ``<course>`` element.
+This would typically be found in ``course/course.xml``. Each XBlock class that
+has ``Scope.context_settings`` field data could be represented as a child
+element of ``<xblock-settings>``, with the name of the element matching the
+XBlock's entry point name.
+
+An example follows:
+
+.. code-block:: xml
+
+   <course display_name="Example Course" language="en" advanced_modules='["drag-and-drop-v2", "lti_consumer", "xblock-dalite"]'>
+     <chapter url_name="6915ee3dd6ab403d8c05c0ea3180a0ee"/>
+     <wiki slug="A.B.C"/>
+     <xblock-settings>
+       <discussion allow_anonymous_discussion_posts="true" discussion_blackout_dates='[["2015-09-15", "2015-09-21"], ["2015-10-01", "2015-10-08"]]' />
+       <lti_consumer lti_passports='["myapp:key1:secret1", "otherapp:key2:secret2"]' />
+       <xblock-dalite dalite_lti_passports='["dalite-ng-demo:https://dalite-ng-demo.example:key:secret"]' />
+     </xblock-settings>
+   </course>
+
+The reasoning behind the above provisional spec is as follows:
+* Conceptually, in the OLX format, all course content and all XBlock fields are
+  serialized as XML nodes, within a root ``<course>`` node `(docs)`_. The entire
+  course's XML may be in a single course.xml file or spread out into multiple
+  files and subdirectories, where the subdirectory name is the XBlock entry
+  point name (XML node name) and the file name is the block ID (``url_name``).
+* Since context-scoped fields are XBlock fields just like Scope.content fields,
+  they should be serialized to OLX the same way (XML, not JSON).
+* Many existing course-scoped settings that are analogous to context-scoped
+  fields (discussion settings, LTI passports, etc.) are currently stored as
+  attributes on the ``<course>`` XML element.
+* The option of serializing context-scoped fields as namespaced attributes on
+  the ``<course>`` element is not feasible as it would be messy and conflicts
+  with the current approach where almost every attribute is an XBlock field
+  belonging to the course XBlock.
+* ``<xblock-settings>`` is suggested as a wrapper to keep these new XML nodes
+  organized and indicate their purpose clearly.
+
+.. _(docs): http://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/directory-structure.html#top-level-directory
+
+Rationale
+=========
+
+Many of the ideas in this proposal are driven by the desire for new types of
+functionality to be implemented via XBlocks, and thus to be pluggable and easily
+installed, enabled, or disabled for each Open edX installation or each course.
+
+The idea of naming the new scope "context" comes from other technologies such as
+LTI, where the term "context" is used analogously. For example, when embedding
+an LTI tool into the Open edX LMS, the LMS will pass the current course ID as
+the LTI "context" parameter.
+
+Previous discussions of note include:
+
+* https://github.com/edx/open-edx-proposals/pull/15
+* https://openedx.atlassian.net/browse/PLAT-325
+* https://groups.google.com/d/msg/edx-code/ywjXV0wzQiw/FRzaK5nTAgAJ
+* On 2016-07-08, the author met with several stakeholders from edX (Calen
+  Pennington, David Ormsbee, and Robert Raposa) to discuss course-scoped XBlock
+  fields. The discussion tended toward using "context" as an additional
+  dimension (see "Context as a New Scope Dimension" in "Rejected Alternatives"),
+  but we did not think of any use cases for 8 of the 12 additional scopes that
+  would be created by defining an additional dimension.
+
+Other related proposals of note include:
+
+* https://github.com/edx/XBlock/pull/317 - XBlock field sharing
+* https://openedx.atlassian.net/wiki/display/AC/Feature+Plugins+for+edX+Platform#FeaturePluginsforedXPlatform-Plug-inArchitecture - Plug-in Architecture
+
+
+
+Backward Compatibility
+=======================
+
+This OEP introduces new field scopes but does not remove or modify any existing
+functionality. XBlock authors that add new fields to their XBlock may opt-in to
+using this feature.
+
+XBlock runtimes may choose to implement context-scoped fields or not. Attempting
+to access a context-scoped field in a runtime that does not support such fields
+should simply return the default value of the field. Attempting to modify such a
+context-scoped field in a runtime that doesn't support such fields should raise
+an ``InvalidScopeError``.
+
+Some existing XBlocks currently use workarounds for the lack of context-scoped
+fields, such as storing data in the fields of the root 'course' XBlock.
+Converting such existing techniques to use context-scoped fields instead would
+be best achieved with an XBlock data migration API, which currently does not
+exist and is outside the scope of this OEP (however, see `here`_ for an example
+implementation).
+
+.. _here: https://github.com/open-craft/problem-builder/blob/c5879dd/problem_builder/v1/xml_changes.py
+
+Details on the current implementation of field scopes in edx-platform can be
+found in `this gist`_. The `field test XBlock`_ can be used to test how a
+runtime behaves with different field scopes.
+
+.. _this GIST: https://gist.github.com/bradenmacdonald/b576c59e46af8df9949fa84b3bc7a4b6
+.. _field test XBlock: https://github.com/open-craft/xblock-field-test
+
+
+
+Open Questions
+==============
+
+#. Should runtimes like the LMS be expected to implement a "Preferences" tab for
+   each course that allows students to edit all ``Scope.context_preferences``
+   fields for XBlocks used in that course?
+
+#. If yes (see previous bullet), can we use a field attribute to allow XBlocks
+   to opt-out of this auto-generated settings UIs on a per-field basis? This may
+   best be addressed at the same time as a standard API for definining editor UI
+   options such as ``allow_reset``, ``multiline_editor``, ``list_style``, and
+   ``values_provider`` as implemented in `xblock-utils`_.
+
+#. Do we need to consider supporting contexts smaller than a course, such as a
+   section or exam? And if so, could that be done in a future change, such as
+   adding a non-scope field property to select a context which defaults to
+   "course"?
+
+   * The author's thought on this is that we likely cannot come up with a
+     reasonable proposal to support every possible use case, and it's better to
+     offer simple, flexible, easily understandable support for the main "course"
+     use case now.
+
+#. Should there be a way to define a field such that the runtime will search
+   through a hierarchy of ``BlockScope`` scopes and use the first value found,
+   such as "check for a ``Scope.settings`` value, then a
+   ``Scope.context_settings`` value"?
+
+.. _xblock-utils: https://github.com/edx/xblock-utils/blob/48fa1a9/xblockutils/studio_editable.py
+
+
+
+Reference Implementation
+========================
+
+TBD. Sample code to implement this feature in the XBlock package and the XBlock
+SDK/Workbench will be produced once the draft has had some initial scrutiny and
+feedback.
+
+
+
+Rejected Alternatives
+=====================
+
+Additional Scopes
+-----------------
+
+This proposal originally included two new ``BlockScope`` values: ``CONTEXT``,
+and ``CONTEXT_ALL_TYPES``; the latter was a proposed new scope that would be
+used for field data that is accessible to XBlocks of any type, within a
+particular context.
+
+It also defined two named scopes that used ``CONTEXT_ALL_TYPES``:
+
+* ``Scope.context_shared_preferences`` (UserScope.ONE, BlockScope.CONTEXT_ALL_TYPES)
+
+  * This scope has been dropped from the proposal as it seems to lack a clear
+    and compelling use case.
+
+* ``Scope.context_shared_settings`` (UserScope.NONE, BlockScope.CONTEXT_ALL_TYPES)
+
+  * This scope has been dropped from the proposal as it seems to lack a clear
+    and compelling use case; further, it is unclear how an authoring experience
+    for such fields would work, since no one XBlock could provide an authoring
+    UI for fields that could be shared by many unrelated XBlocks.
+
+
+Inheritance
+-----------
+
+The current proposal deliberately does not support inheritance (i.e. a "context"
+in the LMS would always be a course, not a "section" or any smaller unit, and
+field values don't propagate from parent blocks to children), so it is not a
+replacement for `InheritanceMixin`_.
+
+Initial discussions of this OEP agreed that context-scoped fields should not
+participate in inheritance. Improving inheritance of fields or making
+inheritable fields pluggable are worthy goals, but for a separate proposal.
+
+.. _InheritanceMixin: https://github.com/edx/edx-platform/blob/46d69eba7ac45/common/lib/xmodule/xmodule/modulestore/inheritance.py#L30
+
+Context as a New Scope Dimension
+--------------------------------
+
+Currently, a ``BlockScope``, a ``UserScope``, and a ``name`` are combined to
+make a Scope for a field. Under this alternative option, a new
+``RelevanceScope`` dimension would be added, which could have one of two values:
+``ALL`` (default) or ``CONTEXT``. The ``ALL`` value would represent the current
+XBlock API's definitions, and the ``CONTEXT`` value would mean that the field's
+relevance is limited to one particular context (e.g. that field's value relates
+to one particular course.)
+
+The downside of this is that it creates a large matrix/tensor of scopes:
+4 BlockScopes × 3 UserScopes × 2 RelevanceScopes = 24 distinct scope types.
+Not only is this more difficult to understand, but there is no value in defining
+a "context-specific" version of any ``BlockScope.USAGE`` or
+``BlockScope.DEFINITION`` fields. As a result of that and other conflicting or
+not useful Scope combinations, it is likely that fewer than half of the 24
+possible Scope values would be used.
+
+Shared data service
+-------------------
+
+Another alternate option is to create a "shared data" or "course-scoped data"
+XBlock service that XBlocks can use to read and write data at a course level,
+rather than making this a new type of field. Such an approach seems less
+consistent, and would not allow the automatic creation of a UI for editing
+context-scoped XBlock settings. (With the field approach, we can replace the
+"Advanced Settings" page in Studio with an auto-generated UI that allows authors
+to configure the course-scoped settings of all XBlocks enabled for that course.
+We can also generate a "preferences" tab in the LMS that allows students to
+change XBlock preferences for that course.)
+
+
+
+Copyright
+=========
+
+.. image:: https://i.creativecommons.org/l/by-sa/4.0/88x31.png
+    :alt: Creative Commons License CC-BY-SA
+    :target: http://creativecommons.org/licenses/by-sa/4.0/
+
+This work is licensed under a `Creative Commons Attribution-ShareAlike 4.0
+International License`_.
+
+.. _Creative Commons Attribution-ShareAlike 4.0 International License: https://creativecommons.org/licenses/by-sa/4.0/


oeps/oep-0006.rst
```diff
@@ -0,0 +1,222 @@
...
+
+* ``Scope.context_settings`` (UserScope.NONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: For an LTI XBlock, a list of the LTI provider URLs and secrets used in the course
+  * Example: For a discussions XBlock: Enable or disable discussions in this course
+
+* ``Scope.context_shared_settings`` (UserScope.NONE, BlockScope.CONTEXT_ALL)
+
+  * Example: Course-specific defaults that can apply to many different blocks, such as "max_attempts" and "show_answer", which would be consumed by the Capa Problem XBlock, Drag and Drop problem XBlock, Problem Builder XBlock family, etc. (These are currently implemented via ``InheritanceMixin``, which is built-in to edx-platform and does not allow XBlock plugins to extend the list of inherited fields. However, a downside of making them a context-scoped field is that each block would require two field definitions: a ``Scope.settings`` field for the usage-specific value, and a ``Scope.context_shared_settings`` field for the default/fallback value. )
```
cpennington: I actually think that inheritance is the right mechanism, from an authoring perspective, and that fields should just be able to set an attribute `inheritable=True` to allow fields to extend the list of inheritable fields.

cpennington: (Which is to say, I don't find this particular example compelling, although I think that something like a shared secret might be a more compelling example of the same concern)

oeps/oep-0006.rst
```diff
@@ -0,0 +1,222 @@
...
+   in any sort of auto-generated "course preferences" UI (for
+   BlockScope.CONTEXT_TYPE|UserScope.ONE fields) or "course advanced settings"
+   UI (for BlockScope.CONTEXT_TYPE|UserScope.NONE fields)?
+
+   * This may be outside the scope of this OEP, and would best be addressed at
+     the same time as a standard API for definining editor UI options such as
+     ``allow_reset``, ``multiline_editor``, ``list_style``, and
+     ``values_provider`` as implemented in `xblock-utils`_.
+
+#. The current proposal does not support inheritance (i.e. a "context" in the LMS would always be a course, not a "section" or any smaller unit, and field values don't propagate from parent blocks to children), so it is not a replacement for `InheritanceMixin`_. Should it be?
```
cpennington: I think not participating in inheritance is the right decision. In my mind, context-scoped things are always global in the context.

cpennington: (And there's only one context, not a set of nested contexts)

oeps/oep-0006.rst
bradenmacdonald: Are there any performance concerns with a proliferation of inheritable fields? I imagine that getting an inherited value that could be set on any parent block is more expensive than a flat course-scoped lookup. But perhaps that's already been solved.

Also, thinking about auto-generated settings UI (which I've discussed in this OEP): If we allow LMS/Studio to auto-generate settings UIs for XBlocks, how would/should an XBlock define "at what level should this setting be set"? (e.g. if I add a shared secret, I want it to be set at the course level, but for 'show answer" I might want the auto-generated settings UI to appear at the course level, section level, subsection level, etc. to give authors more flexibility)

oeps/oep-0006.rst
```diff
@@ -0,0 +1,222 @@
...
+* ``Scope.context_shared_settings`` (UserScope.NONE, BlockScope.CONTEXT_ALL)
+
+  * Example: Course-specific defaults that can apply to many different blocks, such as "max_attempts" and "show_answer", which would be consumed by the Capa Problem XBlock, Drag and Drop problem XBlock, Problem Builder XBlock family, etc. (These are currently implemented via ``InheritanceMixin``, which is built-in to edx-platform and does not allow XBlock plugins to extend the list of inherited fields. However, a downside of making them a context-scoped field is that each block would require two field definitions: a ``Scope.settings`` field for the usage-specific value, and a ``Scope.context_shared_settings`` field for the default/fallback value. )
+  * Example: Submission blackout dates - don't allow users to submit answers during the specified time period.
+
+* ``Scope.context_preferences`` (UserScope.ONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: Enable or disable notifications about new discussion posts in this course
+  * Example: Video playback speed for this course
+  * Example: An XBlock in which students "accumulate building blocks over the course of the semester" (PLAT-325)
```
pdpinch: nit: Can the reference to PLAT-325 be a full URL?

bradenmacdonald: Fixed.

efagin:
@cpennington: if it's all right with you, I'd like you to act as Arbiter for this OEP. Please confirm, thanks!

oeps/oep-0006.rst
cpennington: Yes, there's a use for both inheritance and context scope fields.

For inheritance, we're hoping to solve the performance problems using the block-transforms mechanism in the LMS (which basically can just pre-compute the inheritance on publish).

cpennington:
Yup, I'm fine acting as Arbiter on this one.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+Specification
+=============
+
+New BlockScope values
+---------------------
+
+To support context-scoped fields, ``BlockScope`` will allow these new additional values:
+
+* ``CONTEXT_TYPE``: The data is related to all instances of this type of XBlock within a particular context.
+* ``CONTEXT_ALL``: The data is accessible to XBlocks of any type, within a particular context. (As with ``BlockScope.ALL``, XBlock authors who use such fields should be careful to use a namespace that will not conflict with ``BlockScope.CONTEXT_ALL`` fields defined by other XBlock authors.)
```
andy-armstrong: My first read of this made me think that ``CONTEXT_ALL`` was available to all contexts. Is there another name we could use? ``CONTEXT_ALL_TYPES``?

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+* ``Scope.context_settings`` (UserScope.NONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: For an LTI XBlock, a list of the LTI provider URLs and secrets used in the course
+  * Example: For a discussions XBlock: Enable or disable discussions in this course
+
+* ``Scope.context_shared_settings`` (UserScope.NONE, BlockScope.CONTEXT_ALL)
+
+  * Example: Course-specific defaults that can apply to many different blocks, such as "max_attempts" and "show_answer", which would be consumed by the Capa Problem XBlock, Drag and Drop problem XBlock, Problem Builder XBlock family, etc. (These are currently implemented via ``InheritanceMixin``, which is built-in to edx-platform and does not allow XBlock plugins to extend the list of inherited fields. However, a downside of making them a context-scoped field is that each block would require two field definitions: a ``Scope.settings`` field for the usage-specific value, and a ``Scope.context_shared_settings`` field for the default/fallback value. )
+  * Example: Submission blackout dates - don't allow users to submit answers during the specified time period.
```
andy-armstrong: I'm a bit confused by this use case. Is the idea that an individual XBlock Type could register a field that would introduce submission blackout dates for all block types? Do we do things like that currently with XBlocks Maybe it would help if there was an example of code that would declare a field with such a scope.

There then is the UI question of who is responsible for providing an authoring experience for this field? Since a number of XBlocks provide a custom editor, they wouldn't dynamically pick up these shared fields. Maybe this is out-of-scope of this OEP, but I'm interested in understanding how you would see this working.

As I currently understand this, it seems like something that should be introduced outside of individual XBlock Types entirely. Maybe something like an entry point to register new shared XBlock fields? This would then also need the ability to provide the view for rendering the editor for the field too.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+New named scopes
+----------------
+
+As is the situtation today, XBlock runtimes would generally not be expected to implement every possible combination of ``BlockScope`` and ``UserScope``. Instead, XBlock runtimes are encouraged to implement these new named scopes, which would be defined in the XBlock python package:
+
+* ``Scope.context_settings`` (UserScope.NONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: For an LTI XBlock, a list of the LTI provider URLs and secrets used in the course
+  * Example: For a discussions XBlock: Enable or disable discussions in this course
```
andy-armstrong: If you'd like more examples, this will be a fantastic way to remove all the root course settings for discussions. This would be another great step to allowing the discussion XBlock to be moved to a separate repo and made into an optional install. 

Those settings are:
 - allow anonymous discussion posts
 - allow anonymous discussion posts to peers
 - discussion blackout dates
 - discussion sorting alphabetical
 - discussion topic mapping

Note that a challenge with moving these is how to get the authoring UI to work. We'll need to have a new ``course_settings_view`` or something that the XBlock can declare, and then we'll need to provide a new settings UI in Studio that can show all of the available settings views.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+  * Example: Allow users to self-assign themselves into groups, which affects multiple types of problem XBlock (ORA, Problem Builder)?
+  * This seems to be the least-compelling new Scope, and may be dropped from the proposal if compelling use cases cannot be found.
+
+.. _PLAT-325: https://openedx.atlassian.net/browse/PLAT-325
+
+In the Open edX platform, it is expected that Studio could implement a new UI that displays all ``Scope.context_settings`` and ``Scope.context_shared_settings`` fields from any XBlocks enabled for a given course, allowing authors to configure those settings for that course. Likewise, the LMS could implement a "Preferences" tab for each course that allows students to edit all ``Scope.context_preferences``/``Scope.context_shared_preferences`` fields for XBlocks used in that course. (Though in the future, XBlocks should be able to opt-out of these auto-generated settings UIs on a per-field basis, and instead update those settings via their own UX in views like ``studio_view``, ``student_view``, or the separately proposed ``course_view``. This is expanded upon below in "Open Questions".)
```
andy-armstrong: I'd change this to say that "Studio will implement a new UI" because without this these new fields won't be any use to anyone.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Previous discussions of note include:
+
+* https://openedx.atlassian.net/browse/PLAT-325
+* https://groups.google.com/d/msg/edx-code/ywjXV0wzQiw/FRzaK5nTAgAJ
+* On 2016-07-08, the author met with several stakeholders from edX (Calen Pennington, David Ormsbee, and Robert Raposa) to discuss course-scoped XBlock fields. The discussion tended toward using "context" as an additional dimension (see "Context as a New Scope Dimension" in "Rejected Alternatives"), but we did not think of any use cases for 8 of the 12 additional scopes that would be created by defining an additional dimension.
+
+Other related proposals of note include:
+
+* https://github.com/edx/XBlock/pull/317 - XBlock field sharing
```
andy-armstrong: https://openedx.atlassian.net/wiki/display/AC/Feature+Plugins+for+edX+Platform#FeaturePluginsforedXPlatform-Plug-inArchitecture

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Details on the current implementation of field scopes in edx-platform can be found in `this gist`_. The `field test XBlock`_ can be used to test how a runtime behaves with different field scopes.
+
+.. _this GIST: https://gist.github.com/bradenmacdonald/b576c59e46af8df9949fa84b3bc7a4b6
+.. _field test XBlock: https://github.com/open-craft/xblock-field-test
+
+Open Questions
+==============
+
+#. How can an XBlock specify whether a context-scoped field should participate
```
andy-armstrong: IMO, the XBlock should be able to declare a view for its context-scoped fields. The UI I'm imagining would not show all the fields in one giant list, but would instead show the list of XBlocks that support context-scoped field editing, and then clicking on one would show its editing view. This could be something like the "System Preferences" tool in macOS.

image

oeps/oep-0006.rst
andy-armstrong: I agree. I do think we need a better mechanism for introducing inheritable fields, i.e. to make this a pluggable entry point too, but that can be a different proposal. For example, Student Notes should be a pure plugin, but it has two fields in InheritanceMixin (edxnotes and edxnotes_visibility). Now that I write this, I wonder why they need to be inheritable anyway...

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+   UI (for BlockScope.CONTEXT_TYPE|UserScope.NONE fields)?
+
+   * This may be outside the scope of this OEP, and would best be addressed at
+     the same time as a standard API for definining editor UI options such as
+     ``allow_reset``, ``multiline_editor``, ``list_style``, and
+     ``values_provider`` as implemented in `xblock-utils`_.
+
+#. The current proposal does not support inheritance (i.e. a "context" in the LMS would always be a course, not a "section" or any smaller unit, and field values don't propagate from parent blocks to children), so it is not a replacement for `InheritanceMixin`_. Should it be?
+
+#. Related: Do we need to consider supporting contexts smaller than a course, such as a section or exam? And if so, could that be done in a future chanage, such as adding a non-scope field property to select a context which defaults to "course"?
```
andy-armstrong: Typo: chanage

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+   * The author's thought on this is that we likely cannot come up with a reasonable proposal to support every possible use case, and it's better to offer simple, flexible, easily understandable support for the main "course" use case now.
+
+#. Should there be a way to define a field such that the runtime will search through a hierarchy of ``BlockScope`` scp[es and use the first value found, such as "check for a ``Scope.settings`` value, then a ``Scope.context_settings`` value"?
```
andy-armstrong: Typo: scp[es

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
++---------------+-------------------------------------------+
+|  Created      | 2016-08-03                                |
++---------------+-------------------------------------------+
+| `Resolution`  |                                           |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+There is currently no standard way for XBlocks to define data fields such that content, settings, or user input can be shared among multiple XBlocks within a particular "context" such as a course. Such "course-scoped" or "context-scoped" fields are a common requirement; for example, an author may wish to specify the default “number of attempts allowed” once for all problems in a course, or a video player XBlock may want to “remember” the user’s preferred video playback speed for videos within a particular course.
```
bradenmacdonald: Including hard line wrapping in markdown/rst files can make it messier to see changes during code review, when adding a single word can require re-wrapping all the subsequent lines in a paragraph - which is also annoying. If we want to include line length limits I would prefer to see a system where line wrapping is added just before merging. However, that's just my opinion - I'll use whatever system is agreed on.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Other related proposals of note include:
+
+* https://github.com/edx/XBlock/pull/317 - XBlock field sharing
+
+
+Backward Compatibility
+=======================
+
+This OEP introduces a new field scope but does not remove or modify any existing functionality. XBlock authors that add new fields to their XBlock may opt-in to using this feature.
```
ormsbee: Nitpick: As it stands, it introduces multiple scopes.

oeps/oep-0006.rst
ormsbee: But by default, we just have a generic view that shows all of them?

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+#. Related: Do we need to consider supporting contexts smaller than a course, such as a section or exam? And if so, could that be done in a future chanage, such as adding a non-scope field property to select a context which defaults to "course"?
+
+   * The author's thought on this is that we likely cannot come up with a reasonable proposal to support every possible use case, and it's better to offer simple, flexible, easily understandable support for the main "course" use case now.
+
+#. Should there be a way to define a field such that the runtime will search through a hierarchy of ``BlockScope`` scp[es and use the first value found, such as "check for a ``Scope.settings`` value, then a ``Scope.context_settings`` value"?
+
+.. _xblock-utils: https://github.com/edx/xblock-utils/blob/48fa1a9/xblockutils/studio_editable.py
+.. _InheritanceMixin: https://github.com/edx/edx-platform/blob/46d69eba7ac45/common/lib/xmodule/xmodule/modulestore/inheritance.py#L30
+
+
```
ormsbee: What will the OLX representation for these fields look like? A separate doc that has an entry for each block type?

oeps/oep-0006.rst
bradenmacdonald: Typo fixed.

oeps/oep-0006.rst
bradenmacdonald: Typo fixed.

oeps/oep-0006.rst
bradenmacdonald: Fixed.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
++---------------+-------------------------------------------+
+| `Resolution`  |                                           |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+There is currently no standard way for XBlocks to define data fields such that content, settings, or user input can be shared among multiple XBlocks within a particular "context" such as a course. Such "course-scoped" or "context-scoped" fields are a common requirement; for example, an author may wish to specify the default “number of attempts allowed” once for all problems in a course, or a video player XBlock may want to “remember” the user’s preferred video playback speed for videos within a particular course.
+
+In the absence of a formal method for achieving this through the XBlock API, many other solutions have been put in place, such as: adding custom django models to XBlocks, storing fields on the root “course” XBlock, defining inherited fields using InheritanceMixin, using edx-submissions as a generic data store, and more.
```
bradenmacdonald: Fixed.

oeps/oep-0006.rst
bradenmacdonald: Argh, sorry, not sure how those smart quotes made their way into the abstract.

bradenmacdonald: @cpennington Any comment on line wrapping / length?

oeps/oep-0006.rst
bradenmacdonald: @andy-armstrong That sounds good to me. If we go that route (each XBlock can declare a view for its context-scoped fields), I'd like to make sure we standardize on something like [https://github.com/edx/xblock-utils/#studioeditablexblockmixin](https://github.com/edx/xblock-utils/#studioeditablexblockmixin) so that basic boilerplate UI can easily be generated and we don't get XBlock authors coding idiosyncratic settings UIs where they aren't needed.

There's also the question of how to edit fields with scope `Scope.context_shared_settings` (course-wide settings shared among multiple XBlock types), but I guess the answer there is that developers would have to choose one XBlock to implement the UI, or allow setting the UI via the course-wide settings view of multiple XBlock types.

Oh, that also gets close to the "course_view" concept that we'd also [discussed on the mailing list](https://groups.google.com/d/msg/edx-code/ywjXV0wzQiw/uBeIkoyjBAAJ), which is something that this work is intended to enable.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+* ``Scope.context_preferences`` (UserScope.ONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: Enable or disable notifications about new discussion posts in this course
+  * Example: Video playback speed for this course
+  * Example: An XBlock in which students "accumulate building blocks over the course of the semester" (`PLAT-325`_)
+
+* ``Scope.context_shared_preferences`` (UserScope.ONE, BlockScope.CONTEXT_ALL)
+
+  * Example: Allow users to self-assign themselves into groups, which affects multiple types of problem XBlock (ORA, Problem Builder)?
+  * This seems to be the least-compelling new Scope, and may be dropped from the proposal if compelling use cases cannot be found.
```
bradenmacdonald: Sure, I will drop the `context_shared_preferences` scope. We could always add it later if anyone had a use case.

oeps/oep-0006.rst
bradenmacdonald: Good point. `CONTEXT_ALL_TYPES` sounds good to me and more obviously complements `CONTEXT_TYPE`

oeps/oep-0006.rst
bradenmacdonald: One more thought: will there be any problems with XBlock runtimes instantiating an XBlock and using it to render this sort of "course_settings_view" when there is no `usage_id` for that block? I *think* that's fine as far as the XBlock API  is concerned (?), but it could break some assumptions in the existing runtime code.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+In the absence of a formal method for achieving this through the XBlock API, many other solutions have been put in place, such as: adding custom django models to XBlocks, storing fields on the root “course” XBlock, defining inherited fields using InheritanceMixin, using edx-submissions as a generic data store, and more.
+
+This OEP proposes a new mechanism for defining XBlock fields that are explicitly scoped to a particular context. It is hoped that such a canonical mechanism can provide consistency, improve interoperability, allow new types of behavior to be implemented by XBlocks, and give both authors and learners additional control over learner interactions with XBlocks.
+
+The generic term "context" refers to a course, content library, blog post, or any other place where XBlocks can be used (see "Rationale").
+
+Background
```
bradenmacdonald: I wanted to be clear that this was just a summary of some existing things and not to be confused with a formal/canonical specification of those things. Hence calling this section "Background". For that reason, I would prefer not to put this text in "specification" but could move it elsewhere.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+
+In the absence of a formal method for achieving this through the XBlock API, many other solutions have been put in place, such as: adding custom django models to XBlocks, storing fields on the root “course” XBlock, defining inherited fields using InheritanceMixin, using edx-submissions as a generic data store, and more.
+
+This OEP proposes a new mechanism for defining XBlock fields that are explicitly scoped to a particular context. It is hoped that such a canonical mechanism can provide consistency, improve interoperability, allow new types of behavior to be implemented by XBlocks, and give both authors and learners additional control over learner interactions with XBlocks.
```
bradenmacdonald: Sure, I moved some of the details to the motivation section via 24d982646b217eb9435cb7232ac423bc0874cc35.

oeps/oep-0006.rst
bradenmacdonald: Thanks for these examples! I'll add a few of them to the doc.

bradenmacdonald: With the (now deprecated) mentoring XBlock, we [used](https://github.com/edx-solutions/xblock-mentoring/blob/f461107/mentoring/models.py) a custom django model to store answers so that authors could use that XBlock to prompt students for a free text response answer early in the course, then use a separate instance of that XBlock later in the course to show the student's earlier answer and give them the opportunity to revise it. This approach required that `'mentoring'` be added to the platform's `OPTIONAL_APPS` setting, which is not a pluggable/scalable solution.

For its successor, the Problem Builder suite of XBlocks, we use `edx-submissions` to store answers, and that approach also allows this sort of data sharing among XBlocks (and is a django model). The advantage of that is that it doesn't require any change to `OPTIONAL_APPS`.

oeps/oep-0006.rst
bradenmacdonald: > I'm a bit confused by this use case. Is the idea that an individual XBlock Type could register a field that would introduce submission blackout dates for all block types?

That was my thinking, yes.

Based on this comment and @ormsbee's above, it sounds like we don't yet have a safe and compelling use case for this scope.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,224 @@
...
+* ``Scope.context_shared_settings`` (UserScope.NONE, BlockScope.CONTEXT_ALL)
+
+  * Example: Course-specific defaults that can apply to many different blocks, such as "max_attempts" and "show_answer", which would be consumed by the Capa Problem XBlock, Drag and Drop problem XBlock, Problem Builder XBlock family, etc. (These are currently implemented via ``InheritanceMixin``, which is built-in to edx-platform and does not allow XBlock plugins to extend the list of inherited fields. However, a downside of making them a context-scoped field is that each block would require two field definitions: a ``Scope.settings`` field for the usage-specific value, and a ``Scope.context_shared_settings`` field for the default/fallback value. )
+  * Example: Submission blackout dates - don't allow users to submit answers during the specified time period.
+
+* ``Scope.context_preferences`` (UserScope.ONE, BlockScope.CONTEXT_TYPE)
+
+  * Example: Enable or disable notifications about new discussion posts in this course
+  * Example: Video playback speed for this course
+  * Example: An XBlock in which students "accumulate building blocks over the course of the semester" (`PLAT-325`_)
```
bradenmacdonald: Sure, the first example isn't great if there is a better mechanism for that.

> As for the video example -- I suppose we could make the case that there are some subjects or languages where a student might want that flexibility, but I think that's an edge case. 

I mentioned that from personal experience - I normally watch videos at 1x, but I took one course on edX where the prof (who was talking throughout most of the videos) spoke slowly and took long pauses between every clause that he spoke - so for that particular course I sped the videos up to make the frequent silences less annoying.

> I'm concerned that the most compelling use cases I can think of for this scope make it very easy for developers to shoot themselves in the foot.

Good points. Limiting the allowed data size (and/or type?) seems like a way of reducing the potential impact of these problems, though I do worry that developers might not be aware of the limit at first, and build and deploy an XBlock that keeps a growing log, only to find out about the length limit when it's in production.

oeps/oep-0006.rst
bradenmacdonald: I'm not familiar with the OLX architecture goals, but I think something those lines would work - e.g. a `policy/context_fields/[block_type].json` (or .xml) for storing `context_settings` field values. (Or they could be in `[block_type]/fields.json` though not `[block_type]/fields.xml` since 'fields' could theoretically be used as a block ID, and per-usage fields are stored in `[block_type]/[block_id].xml`)

oeps/oep-0006.rst
cpennington: I tend to prefer wrapped lines in the raw text, simply because it's easier to read in an editor (and in the email summaries that github sends). Github wraps them nicely in the diff, but in this case I lean towards optimizing towards the lowest common denominator.

oeps/oep-0006.rst
cpennington: I'd prefer we not use a .json file, and, ideally, I'd prefer we not force a separate XML file, either. On the other hand, adding a top-level node to the root course.xml document is also a pretty big addition.

Can you flesh out a proposal for the changes to OLX?

bradenmacdonald:
Hi everyone, sorry for the long period of silence on this PR; I forgot to post an update on here before my recent extended vacation. I am back now and working on this again. Currently putting together a proposal for the changes to OLX, and will soon be addressing the other outstanding comments (including some which GitHub has hidden but which aren't resolved).

oeps/oep-0006.rst
robrap: @bradenmacdonald: I think we need to do some more thinking about `CONTEXT_ALL` (a.k.a `CONTEXT_ALL_TYPES`).  

I was discussing with @andy-armstrong the [needs of UIBlocks and UIBlockTabs](https://openedx.atlassian.net/wiki/display/TNL/UIBlocks+for+Course+Outline) in the context of Discussions.  It seems clear that we need a way to share settings across UIBlocks and XBlocks (like the proposed `CONTEXT_ALL`), but it is unclear what the best mechanism would be that allows for clear ownership over the declaration and administration of these settings, as well as some mechanism for sharing these settings, even with UIBlocks.

While trying to define the differences between UIBlocks and XBlocks, we said UIBlocks would not be part of the content hierarchy.  The owner of these settings also doesn't need to be part of the content hierarchy, but shares functionality that XBlocks provide.  Is this a third type of block?  Is it just another XBlock with a specialized purpose?   Other?  This is just to kickstart additional thoughts on this topic.

oeps/oep-0006.rst
andy-armstrong: @bradenmacdonald I think there's a key question here about how folks can access course-scoped fields. As @robrap mentioned, a lot of the use cases for course-scoped fields are to provide advanced settings for a feature, not necessarily for a particular block. Currently advanced settings are added to the course module itself, and then any block can access them. Having a course-scoped field attached to a particular block seems to limit the ability for other blocks (both XBlocks and maybe UI blocks) from being able to access them. It also seems odd that you would have to instantiate an XBlock to be able to access these fields, as that requires a usage id etc that may not exist yet. Would the XBlock provide class methods so that you don't need to instantiate it?

To use discussions as an example, there is a discussion feature that is currently exposed as an inline discussion XBlock and as a tab. In this proof-of-concept (https://github.com/edx/edx-platform/pull/13937) I'm experimenting with providing the tab as a UI block. We are also discussing ways in which discussions could be used in a wider scope than just within a course, e.g. to have a program-wide discussion, or an enterprise-wide discussion. It isn't clear how we will implement these new capabilities, but we won't be able to scope settings related to them to a course. This means that we need to define how we manage discussion settings at the feature level. Maybe this is outside the scope of this OEP, and maybe the solution is to have a Django model for the feature.

andy-armstrong: I should add that the admin UI I described above really should be listing features, not XBlocks, IMO. That means we need to provide an admin view per feature, maybe implemented as UI blocks. As we discussed with @mattdrayer, there's interest in an enterprise admin dashboard that might well be modeled in this way. I now think it would be a mistake to build this UI based upon course-scoped views of installed XBlocks. It should instead be built from a list of installed features.

cpennington:
I think that if there's administration that an XBlock needs to expose (to
configure block-type-wide settings), then it should be exposing that view.
Separately, we might have UI blocks that expose other types of
administrative tasks, but I don't think that forcing the block to set up a
separate UI block that has its own strategy for changing block-type
configuration seems like overhead without an upside.

oeps/oep-0006.rst
bradenmacdonald: @robrap Sure. I have already removed `Scope.context_shared_preferences` from the proposal. I think for now I will remove `Scope.context_shared_settings` as well. Per your suggestion, I'll put them into "Rejected Alternatives" or similar.

bradenmacdonald:
I just pushed a bunch of changes which should address almost all review comments to date:
* Removed  `context_shared_settings`, and put it in "Rejected Alternatives" (wherein I also renamed the now-unneeded `CONTEXT_ALL` to `CONTEXT_ALL_TYPES`, per another review comment)
* Moved mention of inheritance from "Open Questions" to "Rejected Alternatives"
* Proposed a view that XBlocks can declare to implement editing of context-scoped fields : https://github.com/edx/open-edx-proposals/pull/15#discussion_r78983373
* Imposed a strict 4 KiB size limit on `context_preferences` fields : https://github.com/edx/open-edx-proposals/pull/15#discussion_r79205976
* Added word wrapping

TODO:
* Proposal for the changes to OLX

bradenmacdonald:
I have added a simple OLX proposal to the spec. Feedback welcome!

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+XBlock's entry point name.
+
+An example follows:
+
+.. code-block:: xml
+
+   <course display_name="Example Course" language="en" advanced_modules='["drag-and-drop-v2", "lti_consumer", "xblock-dalite"]'>
+     <chapter url_name="6915ee3dd6ab403d8c05c0ea3180a0ee"/>
+     <wiki slug="A.B.C"/>
+     <xblock-settings>
```
robrap: @cpennington: Should we use this (i.e. a new top-level tag) as an opportunity to introduce a wrapping tag around the chapters?

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+not useful Scope combinations, it is likely that fewer than half of the 24
+possible Scope values would be used.
+
+Shared data service
+-------------------
+
+Another alternate option is to create a "shared data" or "course-scoped data"
+XBlock service that XBlocks can use to read and write data at a course level,
+rather than making this a new type of field. Such an approach seems less
+consistent, and would not allow the automatic creationn of a UI for editing
```
robrap: typo: creationn => creation

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+context-scoped XBlock settings. (With the field approach, we can replace the
+"Advanced Settings" page in Studio with an auto-generated UI that allows authors
+to configure the course-scoped settings of all XBlocks enabled for that course.
+We can also generate a "preferences" tab in the LMS that allows students to
+change XBlock preferences for that course.)
```
robrap: This is good food for thought as we think through all the scenarios required for Discussions, including its possible future evolution.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+As is the situtation today, XBlock runtimes would generally not be expected to
+implement every possible combination of ``BlockScope`` and ``UserScope``.
+Instead, XBlock runtimes are encouraged to implement these new named scopes,
+which would be defined in the XBlock python package:
+
+* ``Scope.context_settings`` (UserScope.NONE, BlockScope.CONTEXT)
+
+  * Example: For an LTI XBlock, a list of the LTI provider URLs and secrets used
+    in the course
+  * Example: For a discussions XBlock: Enable or disable discussions in this
```
bradenmacdonald: Yeah, this OEP would not provide as much value for the discussions use case if we are not planning to use a `discussion` XBlock that supports both inline and course-tab views, and it would not support the other contexts that @nasthagiri mentions there

bradenmacdonald:
@cpennington @andy-armstrong @robrap What are the next steps for this OEP?

cpennington:
@bradenmacdonald: I'll take a look through today and just double-check that all of the concerns seem resolved, and then we can look at getting it merged.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+Context-scoped fields in OLX
+----------------------------
+
+When serializing a course with ``Scope.context_settings`` fields to Open
+Learning XML (OLX), any context-scoped fields should be described within a new
+``<xblock-settings>`` XML element that is a child of the ``<course>`` element.
+This would typically be found in ``course/course.xml``. Each XBlock class that
+has ``Scope.context_settings`` field data should be represented as a child
+element of ``<xblock-settings>``, with the name of the element matching the
+XBlock's entry point name.
```
ormsbee: Could you please add why it's a child of `<course>` vs. a separate doc or other options? I don't feel particularly strongly on this issue, but it'd be useful to walk through the thinking (backwards compatibility, ease of implementation, etc.)

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+   <course display_name="Example Course" language="en" advanced_modules='["drag-and-drop-v2", "lti_consumer", "xblock-dalite"]'>
+     <chapter url_name="6915ee3dd6ab403d8c05c0ea3180a0ee"/>
+     <wiki slug="A.B.C"/>
+     <xblock-settings>
+       <discussion allow_anonymous_discussion_posts="true" discussion_blackout_dates='[["2015-09-15", "2015-09-21"], ["2015-10-01", "2015-10-08"]]' />
+       <lti_consumer lti_passports='["myapp:key1:secret1", "otherapp:key2:secret2"]' />
+       <xblock-dalite dalite_lti_passports='["dalite-ng-demo:https://dalite-ng-demo.example:key:secret"]' />
+     </xblock-settings>
+   </course>
+
```
ormsbee: How does this interact with `parse_xml()`? Is there a separate method to serialize and deserialize this scope?

cpennington: Yeah, this does seem like it would require a new method to be added to the XBlock APIs.

I'm also not sure that `xblock-settings` quite captures the intent, although I'm struggling to come up with a more accurate name.

Lastly: The `lti_passports` example exposes a downside to this whole mechanism (at least without security markings), which is that it's an attractive nuisance for adding private values and perhaps leaking them out during OLX export.

oeps/oep-0006.rst
bradenmacdonald: @ormsbee Is the following helpful? This is about the extent of my thoughts on the matter.

* Conceptually, in the OLX format, all course content and all XBlock fields are serialized as XML nodes, within a root `<course>` node [(docs)](http://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/directory-structure.html#top-level-directory). The entire course's XML may be in a single `course.xml` file or spread out into multiple files and subdirectories, where the subdirectory name is the XBlock entry point name (XML node name) and the file name is the block ID (`url_name`).
* Since context-scoped fields are XBlock fields just like Scope.content fields, they should be serialized to OLX the same way (XML, not JSON).
* Many existing course-scoped settings that are analogous to context-scoped fields (discussion settings, LTI passports, etc.) are currently stored as attributes on the `<course>` XML element.
* The option of serializing context-scoped fields as namespaced attributes on the `<course>` element is not feasible as it would be messy and conflicts with the current approach where almost every attribute is an XBlock field belonging to the `course` XBlock.
* Based on these reasons, the author recommends putting the new fields as attributes on xblock-type-specific elements that are a child of the `<course>` element. `<xblock-settings>` is suggested as a wrapper to keep these new XML nodes organized and indicate their purpose clearly.

If that is helpful I will add it to the doc.

oeps/oep-0006.rst
ormsbee: Also, I don't think that we should have tags that are named identically to their XBlock content scoped counterparts, if they're going to have different validation rules. But maybe some convention?

A strawman XML:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<context id="course-v1:OECx+TESS101x+3T2016">
  <discussion-context allow_anonymous_discussion_posts="true">
    <blackout start="2015-09-15" end="2015-09-21"/>
    <blackout start="2015-10-01" end="2015-10-08"/>
  </discussion-context>
  <lti_consumer-context>
    <passport>myapp:key1:secret1</passport>
    <passport>otherapp:key2:secret2</passport>
  </lti_consumer-context>
</context>
```

I wonder if that could be applied to overrides as well...?

Anyhow, if there's a desire to have it all map into one document somewhere, maybe we can keep all the course.xml stuff where it is and use XInclude to bring this new doc and course.xml into a more unified doc structure (the specifics of which would be out of scope for this OEP).

ormsbee: Sorry, to clarify the half-thought about overrides -- I meant the notion of describing overrides with tags like `<problem-override>`, but that also fits in the general frame of this structure. That would give it a presence in OLX, a separate tag name for different validation rules, and a reasonable way to import/export.

It would also imply that there might be a family of XML parsing (even one for user state, or a separate one for sensitive data). In that case, `parse_xml` might stay the same, but there might be a higher level method that inflates an XBlock's state in several steps (first the context stuff, then the definition, then the overrides, then the user state...)

I realize that's all way out of scope of what should be covered in this OEP. I mention it because if we had a document with these sorts of things (or multiple documents for secrets, overrides, and context), then that's a concrete data representation of what @nasthagiri and you were talking about earlier in regards to simplifying CCX as part of a more unified course run management system.

bradenmacdonald: I'm not opposed to putting these into a separate file; it's just not clear to me what such a separate file should be called nor how it should be referenced.

> Also, I don't think that we should have tags that are named identically to their XBlock content scoped counterparts

Good point, though I think we'd need to use a prefix like `<xblock-settings:blockname>` because if we went for a `-context` suffix or similar, there's nothing to stop someone from creating an XBlock with a name like `discussion-context`.

> having written that makes me wonder if that scope belongs in Studio at all. The use cases do seem to fall more on the execution side of the authoring/execution divide.

Yes, I can see both arguments. Currently a lot of execution settings are set via Studio, and we have no proper "course run" tool. Also, I think it doesn't really matter as far as this OEP goes whether those settings are administered in Studio or in the LMS - most of this OEP would be unchanged, other than the removal of an OLX spec since it wouldn't be serialized into the OLX.

cpennington: Around the names: I think as far as validation goes, you can have different rules in different contexts, right? So it might be ok to have the course-level settings use the same tag names as the block contents.

ormsbee: @cpennington: You could, but I think it would be more confusing, particularly if we ever mixed them up in the same document.

@bradenmacdonald: While it's true that nothing stops people from using an arbitrary tag name, I do think we can reasonably enforce a convention, even if the actual mapping happens in the form of explicit registry in entry points. I think that in general, the namespaces should be reserved on an XBlock basis -- so that it's main job is to disambiguate two "video" tags from two XBlocks, rather than two scoped tags from the same XBlock.

In any case, I apologize for stirring up this particular pot. I thought that OLX representation could be straightforward and simple, but I don't want to drag us into a full definition of OLX representations across different scopes as part of this OEP. Do you folks think it's reasonable to accept this OEP on the new scope, and punt on the serialization aspect as a separate follow-on step?

Thank you.

cpennington:
@andy-armstrong: Do you have any additional concerns?

oeps/oep-0006.rst
cpennington: That seems like a separate concern from the format for this particular OEP. I think I'd rather see that tackled as a concentrated OEP on OLX itself.

oeps/oep-0006.rst
bradenmacdonald: Thanks @cpennington. Sounds good.

ormsbee:
👍  as long as the OLX spec is marked as provisional. Thank you!

cpennington:
@bradenmacdonald looks like you've got a few tidying up things to do, and then we should be able to mark this accepted and merge.

oeps/oep-0006.rst
```diff
@@ -0,0 +1,410 @@
...
+content authors to edit the data in those fields. (e.g. when creating a course,
+the course author may wish to change course-wide settings for the discussion
+XBlock used throughout that course.)
+
+A new XBlock view, ``context_settings_view`` shall be used for this purpose.
+Just like the existing `studio_view` defined by edX Studio, this new view will
+allow an XBlock to return an arbitrary HTML Fragment which contains a form or
+other UI for editing its ``Scope.context_settings`` fields.
+
+It is expected (but not required) that runtimes like edX Studio which want to
```
robrap: @bradenmacdonald: Does the runtime know the difference between an XBlock instance and the XBlock definition (or class or whatever the appropriate name is)?  Are all existing views (like `studio_view`) XBlock instance views?  

Are we certain that an XBlock view like `context_settings_view` is the right solution for this given these differences?  Should this also be marked as provisional?

bradenmacdonald: @robrap Good questions.

> Does the runtime know the difference between an XBlock instance and the XBlock definition (or class or whatever the appropriate name is)?

Yes, that's pretty clear in the code I think.

> Are all existing views (like studio_view) XBlock instance views?

Yes. So I think that `context_settings_view` should perhaps be an instance view as well - it's just that in such a case, the XBlock instance would be initialized with a [`ScopeIds`](https://github.com/edx/XBlock/blob/ab636a82bc23c3f65a618e438637d518b209790f/xblock/fields.py#L237) value where `def_id` and `usage_id` are `None`.

This is something that I will have a more clear idea of once I have a working prototype.

bradenmacdonald:
@cpennington I've made the cleanups related to OLX - anything else?

I will also start working on prototyping this, which should help clarify a few outstanding questions.

cpennington:
@bradenmacdonald: This looks good to me. Can you mark it as `Accepted`, and add a link to this discussion, and then I'll merge it?

bradenmacdonald:
@cpennington Done, thanks. I didn't squash the commits but am happy to do that if you'd like.

cpennington:
Yeah, I guess lets squash them.

bradenmacdonald:
Squashed from 7d72dbf62c37848ad6bc8d9f6b6c57a2e9f8aa3f to 8dbf0dd52a351cf01a08e6ada9d81f07b35e85d6

PR #16, opened 2016-08-08
Create OEP-7 draft to replace OEP-1, to support Product proposals

jbarciauskas:
The purpose of this OEP is to add support for proposing changes to the
functionality of the Open edX platform to the OEP process. Rather than
using multiple processes, I believe the OEP process can be easily
adapted to this case, particularly with some light support for less technical
forms of discussion.

Changelog:

* New "Product" proposal type
* Remove references to arch@ email address
* Created "Inital Submission" section
* Increased scope of Arbiter role to include helping with GitHub and
  other technical mechanics as needed
* Added support for Google Docs and other external forums for discussion
  of the proposal
* Added "References" field to the table as a place to put links to
  external discussions

### DIFF
--- a/oep-template.rst
+++ b/oep-template.rst
@@ -32,6 +32,9 @@ OEP-XXX: OEP Template
 +---------------+-------------------------------------------+
 | `Replaced-By` | <OEP number>                              |
 +---------------+-------------------------------------------+
+| `References`  | <links to any other relevant discussions  |
+|               | or relevant related materials>            |
++---------------+-------------------------------------------+
 
 Abstract
 ========
--- a/oeps/oep-0007.rst
+++ b/oeps/oep-0007.rst
@@ -1,0 +1,395 @@
+====================================
+OEP-7: OEP Purpose and Guidelines v2
+=====================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-7 </oeps/oep-0007>`             |
++---------------+-------------------------------------------+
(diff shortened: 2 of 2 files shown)


jbarciauskas:
@efagin @nedbat @cpennington Pinging you before sending this more widely

bradenmacdonald:
@jbarciauskas Can you call this OEP-7? I've already started using OEP-6 in #15.

nedbat:
Two points of process: Why do we need a new OEP, why not just update OEP-1?  And we should have an index page to avoid the number collisions that Braden points out.

jbarciauskas:
I'm fine updated OEP-1, I created a similar PR here: https://github.com/edx/open-edx-proposals/pull/17

We can debate which to use, I'm agnostic.

@bradenmacdonald sure

jbarciauskas:
One argument in favor of #17 is it's easier to read the diff on GitHub :)

efagin:
My opinion: keep incremental changes to the original OEPs, but move to a new OEP if the change is breaking somehow. I mean, really, this would be OEP-1.1.0 right? :)

jbarciauskas:
Closing in favor of #17 for now

cpennington:
Per the instructions on OEP-1 (I thought, anyway), you should be checking the existing open proposal PRs before giving yourself a number.

cpennington:
Also, yes, updating OEP-1 was the right call, IMO.

jbarciauskas:
Yes, I was lazy and didn't check.

PR #17, opened 2016-08-08
Add support for product changes to OEP-1

jbarciauskas:
The purpose of this pull request is to add support for proposing changes to the
functionality of the Open edX platform to the OEP process. Rather than
using multiple processes, I believe the OEP process can be easily
adapted to this case, particularly with some light support for less technical
forms of discussion.

Changelog:

* New "Product" proposal type
* Remove references to arch@ email address
* Created "Inital Submission" section
* Increased scope of Arbiter role to include helping with GitHub and other
  technical mechanics as needed
* Added support for Google Docs and other external forums for discussion of
  the proposal
* Added "References" field to the table as a place to put links to external
  discussions

### DIFF
--- a/oeps/oep-0001.rst
+++ b/oeps/oep-0001.rst
@@ -89,11 +89,11 @@ Initial Submission
 ------------------
 
 Once the Author has asked the Open edX community whether an idea has any chance
-of acceptance, a draft OEP should be written submitted as a pull request against the
-`central OEP repository`_. To identify the draft proposal, the Author should check the
-numbered list of previous OEP pull requests and select the next available
-number. The pull request title must start with "OEP-XXX", which claims that OEP
-number for the included proposal.
+of acceptance, a draft OEP should be submitted as a pull request against the
+`central OEP repository`_. To identify the draft proposal, the Author should
+check the numbered list of previous OEP pull requests and select the next 
+available number. The pull request title must start with "OEP-XXX", which 
+claims that OEP number for the included proposal.
 
 .. _central OEP repository: http://github.com/edx/open-edx-proposals
 
--- a/oeps/oep-0001.rst
+++ b/oeps/oep-0001.rst
@@ -51,6 +51,9 @@ OEP Types
 * An **Architecture** proposal describes relationships between Open edX
   services and libraries, including their APIs and data formats.
 
+* A **Product** proposal describes a significant enhancement to the
+  functionality of the Open edX platform.
+
 Workflow
 ========
 
@@ -58,8 +61,8 @@ Workflow
   :local:
   :depth: 1
 
-Submitting an OEP
------------------
+Before Submitting an OEP
+------------------------
 
 The OEP process begins with a new idea for Open edX. It is highly recommended
 that a single OEP contain a single key proposal or new idea. Small enhancements
@@ -70,8 +73,7 @@ Each OEP must have a Author: someone who writes the OEP using the style and
 format described below, shepherds the discussions in the appropriate forums,
 and attempts to build community consensus around the idea. The OEP Author
 should first attempt to ascertain whether the idea is appropriate for an OEP.
-Posting to either the `edx-code`_ mailing list, or to ``arch@edx.org`` and
-``archexternal@edx.org`` is the best way to go about this.
+Posting to the `edx-code`_ mailing list is the best way to go about this.
 
 Vetting an idea publicly before going as far as writing an OEP is meant to save
 the potential author time. Many ideas have been brought forward for changing
@@ -83,28 +85,23 @@ sure the idea is applicable to the entire community and not just the author.
 Just because an idea sounds good to the author does not mean it will work for
 most Open edX installations.
 
+Initial Submission
+------------------
+
 Once the Author has asked the Open edX community whether an idea has any chance
-of acceptance, a draft OEP should be submitted as a pull request against the
-`central OEP repository`_. To identify the draft proposal, the Author should
-check the numbered list of previous OEP pull requests and select the next
-available number. The pull request title must start with "OEP-XXX", which
-claims that OEP number for the included proposal.
+of acceptance, a draft OEP should be written submitted as a pull request against the
+`central OEP repository`_. To identify the draft proposal, the Author should check the
+numbered list of previous OEP pull requests and select the next available
+number. The pull request title must start with "OEP-XXX", which claims that OEP
+number for the included proposal.
 
 .. _central OEP repository: http://github.com/edx/open-edx-proposals
 
-As updates are necessary, the OEP Author can update the pull request.
-
-OEPs can be reference pull requests to other Open edX repositories which
-will act as reference implementations.
-
-OEP Review & Resolution
------------------------
-
-After the Author drafts an OEP, and submits it as a pull request, they request
-an Arbiter from the edX Chief Architect. This Arbiter will be recorded in the
-"Arbiter" header on the OEP. The rest of Open edX community will be given the
-opportunity to comment on the OEP, with the Arbiter serving to keep the
-discussion on track and to evaluate when it has reached a final conclusion.
+After the Author drafts an OEP in a format in which they are comfortable, they
+request an Arbiter from the edX Chief Architect. This Arbiter will be recorded
+in the "Arbiter" header on the OEP. The rest of Open edX community will be
+given the opportunity to comment on the OEP, with the Arbiter serving to keep
+the discussion on track and to evaluate when it has reached a final conclusion.
 
 For an OEP to be accepted by the Arbiter, it must meet certain minimum
 criteria. It must be a clear and complete description of the proposed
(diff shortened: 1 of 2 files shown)


oeps/oep-0001.rst
```diff
@@ -83,28 +85,23 @@ sure the idea is applicable to the entire community and not just the author.
...
+Initial Submission
+------------------
+
 Once the Author has asked the Open edX community whether an idea has any chance
-of acceptance, a draft OEP should be submitted as a pull request against the
-`central OEP repository`_. To identify the draft proposal, the Author should
-check the numbered list of previous OEP pull requests and select the next
-available number. The pull request title must start with "OEP-XXX", which
-claims that OEP number for the included proposal.
+of acceptance, a draft OEP should be written submitted as a pull request against the
```
efagin: `written submitted` - is this a typo? I'm not quite following the intent.

jbarciauskas:
fyi @nedbat @cpennington - going ahead with this version

jbarciauskas:
@efagin removed - was likely thinking I was going to change wording there, moved it below instead.

oeps/oep-0001.rst
```diff
@@ -7,7 +7,7 @@ OEP-1: OEP Purpose and Guidelines
 +---------------+-------------------------------------------+
 | Title         | OEP Purpose and Guidelines                |
 +---------------+-------------------------------------------+
-| Last-Modified | 2016-06-24                                |
+| Last-Modified | 2016-08-08                                |
 +---------------+-------------------------------------------+
 | Author        | Calen Pennington <cale@edx.org>           |
```
feanil: Should the Author list contain both you and Cale now?

efagin: I think pluralizing Author to Authors and adding Joel's name makes sense. We should leave Cale's name first as the original author.

oeps/oep-0001.rst
```diff
@@ -284,6 +291,12 @@ ReStructuredText [8] allows for rich markup that is relatively easy to read,
 and can also be rendered into good-looking and functional HTML. OEPs are
 rendered to HTML using Sphinx. An `OEP template`_ can be found in the repo.
 
+OEPs may be discussed in a more convenient format, such as a Google Doc, if
+it is deemed most appropriate for the audience. The final OEP should be
+committed as described above with a link to the location of any discussion,
```
feanil: So even if the discussion happens elsewhere it is expected to be public.  Can we make that more clear here.

oeps/oep-0001.rst
efagin: (or not pluralize to keep the header consistent, doesn't matter much to me either way)

antoviaque:
@jbarciauskas Great idea imho - thank you for submitting this.

oeps/oep-0001.rst
```diff
@@ -83,28 +85,23 @@ sure the idea is applicable to the entire community and not just the author.
...
-OEP Review & Resolution
------------------------
-
-After the Author drafts an OEP, and submits it as a pull request, they request
-an Arbiter from the edX Chief Architect. This Arbiter will be recorded in the
-"Arbiter" header on the OEP. The rest of Open edX community will be given the
-opportunity to comment on the OEP, with the Arbiter serving to keep the
-discussion on track and to evaluate when it has reached a final conclusion.
+After the Author drafts an OEP in a format in which they are comfortable, they
+request an Arbiter from the edX Chief Architect. This Arbiter will be recorded
```
antoviaque: Does this also applies to product-related OEPs? Or would it be someone from product selecting the Arbiter in these cases?

antoviaque:
> In general, OEPs are not modified after they have reached the Final state. They can be replaced by subsequent OEPs, however (OEPs that are replaced are given the status "Replaced").

This should probably be updated to reflect the decision from https://github.com/edx/open-edx-proposals/pull/16#issuecomment-238285077

antoviaque:
What about governance decisions? Could propositions be submitted as OEPs for them, too?

marcotuts:
#Subscribing. :) @Marcotuts

oeps/oep-0001.rst
```diff
@@ -114,6 +111,15 @@ implementation, if applicable, must pass the existing
 
 .. _Open edX Contribution Guidelines: http://edx.readthedocs.org/projects/edx-developer-guide/en/latest/process/index.html
 
+As updates are necessary, the OEP Author or Arbiter can update the pull
+request.
+
+OEPs can be reference pull requests to other Open edX repositories which
```
jzoldak: typo: can reference?

msegado:
Lines [24](https://github.com/edx/open-edx-proposals/pull/17/files#diff-73453e51458e7b9deae9ba5b1e3aced8R24) and [60](https://github.com/edx/open-edx-proposals/pull/17/files#diff-73453e51458e7b9deae9ba5b1e3aced8R60) in the OEP template should have "Product" added too, right?

oeps/oep-0001.rst
jbarciauskas: I think for the sake of simplicity we will leave it as the Chief Architect. @efagin loves our product team, right?

jbarciauskas:
@msegado yep, thanks!

jbarciauskas:
@antoviaque with regard to your two comments:

"This should probably be updated to reflect the decision from #16 (comment)"
In what way? OEP-1 is Accepted but not Final. It may not be worth drawing that sort of fine distinction, though, and just remove that clause, and leave the decision to edit versus replace to be a case by case decision.

On the governance question, I see this process as largely representing our governance process, so proposals to change our governance process might be Best Practice or Process OEPs or updates to OEP-1. Open to your feedback.

jbarciauskas:
@marcotuts can you provide any further feedback by Thursday?

Ditto for anyone else interested

marcotuts:
@jbarciauskas  - If possible I'd love to discuss offline in more detail. I'm not sure I understand OEP's and their intended use beyond tracking architectural decisions in the community for the platform. 

My concern is on driving product changes and proposals through github and the pull request process in general, as this isn't necessarily a place or format that will feel familiar to large fractions of the stakeholders that generally propose these kinds of improvements / changes.

jbarciauskas:
I guess one question I have is whether we should use Product Enhancement or Product Direction... the latter feels more accurate for this type of process

jbarciauskas:
@marcotuts happy to chat in person, but the purpose of this specific change to OEP-1 was to support proposals in formats other than GitHub PRs / ReStructured Text

marcotuts:
:+1:

cpennington:
The OEP repo will still be the system of record, though, right?

antoviaque:
@jbarciauskas 

>> "This should probably be updated to reflect the decision from #16 (comment)"
> In what way? OEP-1 is Accepted but not Final. It may not be worth drawing that sort of fine distinction, though, and just remove that clause, and leave the decision to edit versus replace to be a case by case decision.

That seem fine to me, yep.

> On the governance question, I see this process as largely representing our governance process, so proposals to change our governance process might be Best Practice or Process OEPs or updates to OEP-1. Open to your feedback.

Cool, that would definitely address my comment. To give a specific example - how would you approach suggesting a process/responsibilities for the "external" core commiters? Would that fit what you are mentioning? Would it be its own OEP? An edit somewhere else?

oeps/oep-0001.rst
```diff
@@ -289,6 +298,12 @@ ReStructuredText [8] allows for rich markup that is relatively easy to read,
 and can also be rendered into good-looking and functional HTML. OEPs are
 rendered to HTML using Sphinx. An `OEP template`_ can be found in the repo.
 
+OEPs may be discussed in a more convenient format, such as a Google Doc, if
+it is deemed appropriate for the audience and sufficiently open to comment and
+review. The final OEP should be committed as described above with a link to the
```
jbarciauskas: @cpennington I guess we could make it more clear here that we intend the final form of the document under discussion to be committed to this repo?

jbarciauskas:
@antoviaque @efagin I would say the process for becoming a core committer would be a reasonable process OEP. The process for contributions being merged could also be an OEP, though we try to keep that relatively flexible and allow different policies on different repos, depending on their activity and criticality.

oeps/oep-0001.rst
cpennington: Yeah, I'd like to make sure that's clear, since one of the original goals of the OEP process was to make sure we had a single source of truth for accepted proposals.

nedbat: Are we going to commit PDF files? I don't see another way that will accurately and simply capture the Google Doc.

cpennington: Personally, I'd want the doc translated to the OEP reST format, and committed that way (that is, the Google Doc would be the bulk of the discussion, but the resulting proposal would match up with the existing process).

jbarciauskas: That was my expectation, yes @cpennington 

Not sure what the concern with that might be @nedbat ?

nedbat: People designing features often include screenshots. And I'm not sure I see the benefit of transcribing to .rst for feature proposals.  Do we all have the same understanding about how frequently OEPs might be written for product-level decisions? This seems like it will be really foreign to the people who usually make product decisions.

cpennington: My concern is that by not transcribing, we lose the central source of truth that actually makes OEPs useful. With an reST doc, we can see if changes were made after agreement. With a google doc, that's not true.

nedbat: We really need the likely proposers of "significant product enhancements" to weigh in on whether this process will work for them.  I understand the value of a single auditable trail of information. But will the people designing the enhancements use this process?

antoviaque:
@jbarciauskas 

> I would say the process for becoming a core committer would be a reasonable process OEP. The process for contributions being merged could also be an OEP, though we try to keep that relatively flexible and allow different policies on different repos, depending on their activity and criticality.

That sounds like a great approach - perfect on my side :+1:

PR #18, opened 2016-08-24
Add definitions for a few more standard OEP-2 keys

cpennington:
@nedbat, @efagin: This adds definitions for a few of the keys that we are using in the `openedx.yaml` files.

### DIFF
--- a/oeps/oep-0002.rst
+++ b/oeps/oep-0002.rst
@@ -7,7 +7,7 @@ OEP-2: Repository Metadata
 +---------------+-------------------------------------------+
 | Title         | Repository Metadata                       |
 +---------------+-------------------------------------------+
-| Last-Modified | 2016-06-29                                |
+| Last-Modified | 2016-08-24                                |
 +---------------+-------------------------------------------+
 | Author        | Calen Pennington <cale@edx.org>           |
 +---------------+-------------------------------------------+
@@ -50,14 +50,16 @@ following fields:
 ``owner``: string
     The github user/team responsible for the repository.
 
-``nick``: string
-    A short-name for the repository used by reporting tools.
+``nick``: string (optional)
+    A short-name for the repository used by reporting tools created by the
+    edX Open Source team. Should be 3-4 characters, because this value
+    is used in charts with very limited space on their axes.
 
-``tags``: list of strings
+``tags``: list of strings (optional)
     Used to group repositories by in reporting tools. Sample tags: core,
     mobile analytics.
 
-``oeps``: dictionary
+``oeps``: dictionary (optional)
     A list of keys corresponding to Best Practice OEPs (in the format
     ``oep-n``). Each value would be one of ``True``, ``False``, or a
     reason dictionary. ``True`` and ``False`` would indicate accordance (or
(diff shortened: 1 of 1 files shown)


oeps/oep-0002.rst
```diff
@@ -69,6 +70,32 @@ following fields:
...
+
+    .. This should be moved to a separate OEP
+
+    Whether or not this repository should be tracked for creation of OSPR tickets.
+
+``openedx-release``: dictionary (optional)
+
+    .. This should be moved to a separate OEP
+
+    This controls how this repo is managed during the edX Named Release process.
```
nedbat: "during the Open edX release process".  We are moving away from "named release" as a term.

oeps/oep-0002.rst
```diff
@@ -69,6 +70,32 @@ following fields:
     If an OEP isn't listed in the ``oeps`` dictionary, then it is assumed to be
     ``False``, unless the reporting tool can auto-detect accordance.
 
+``track-pulls``: boolean (optional)
+
+    .. This should be moved to a separate OEP
+
+    Whether or not this repository should be tracked for creation of OSPR tickets.
```
nedbat: There's an irony to having this key any more.  When read from repos.yaml, it meant that we didn't have to look at every repo in the organization. Now that it's distributed, we have to look at every repo to see if this is set. Probably we should just check every repo no matter what.  This was just an optimization.

oeps/oep-0002.rst
```diff
@@ -69,6 +70,32 @@ following fields:
...
+
+``openedx-release``: dictionary (optional)
+
+    .. This should be moved to a separate OEP
+
+    This controls how this repo is managed during the edX Named Release process.
+
+    Possible keys:
```
nedbat: Suggested wording (plus another key):
```
    Possible keys:
    ``ref``
        The name of a branch in this repo. This is the branch that will be tagged when an Open edX release is made.
    ``requirements``
        The path to a requirements.txt-style file in this repo. This file will be consulted to find the commit to tag in dependent repos indicated by ``parent-repo`` keys.
    ``parent-repo``
        The name of another repo, one that installs this repo via a requirements file.  The parent repo's requirements file will indicate the commit to tag in this repo.
```

oeps/oep-0002.rst
```diff
@@ -7,7 +7,7 @@ OEP-2: Repository Metadata
 +---------------+-------------------------------------------+
 | Title         | Repository Metadata                       |
 +---------------+-------------------------------------------+
-| Last-Modified | 2016-06-29                                |
+| Last-Modified | 2016-08-24                                |
```
nedbat: There should be an explicit Change History section.

cpennington: Good call, I'll add that to OEP-1.

oeps/oep-0002.rst
cpennington: Ah, I thought that was there because we didn't want the bot commenting on every repo.

oeps/oep-0002.rst
cpennington: Added in #19.

nedbat:
:+1:

oeps/oep-0002.rst
```diff
@@ -50,14 +50,15 @@ following fields:
 ``owner``: string
     The github user/team responsible for the repository.
 
-``nick``: string
-    A short-name for the repository used by reporting tools.
+``nick``: string (optional)
+    A short-name for the repository used by reporting tools created by the
+    edX Open Source team. Should be 3-4 characters.
```
benpatterson: `Should be 3-4 characters` that is so helpful! thank you.

benpatterson:
👍 from me. Thank you for putting in this update.

oeps/oep-0002.rst
```diff
@@ -69,6 +70,39 @@ following fields:
...
+        The path to a requirements.txt-style file in this repo. This file will
+        be consulted to find the commit to tag in dependent repos indicated by
+        ``parent-repo`` keys.
+
+    ``parent-repo``
+        The name of another repo, one that installs this repo via a requirements
+        file.  The parent repo's requirements file will indicate the commit to
+        tag in this repo.
+
+``dead``: boolean (optional)
```
efagin: Why can't we use a more standard (and less morbid) term like `obsolete` here? (You already use that in the description.)

oeps/oep-0002.rst
efagin: Agree. Sounds like we should rip this out and revisit if that use case ever emerges. Seems easy enough to bolt on later if we need it.

efagin:
👍 this is technically merge-ready even though I had a few minor suggestions.

oeps/oep-0002.rst
cpennington: Yeah, I can do that. I'll update the existing repos that are marked w/ `dead`.

PR #19, opened 2016-08-24
Add a Change History section to OEP-1

cpennington:
@nedbat, @efagin: Review?

### DIFF
--- a/oep-template.rst
+++ b/oep-template.rst
@@ -109,6 +109,12 @@ considered and rejected, and why they were not implemented. It should also link
 to the original source of that discussion.
 
 
+Change History
+==============
+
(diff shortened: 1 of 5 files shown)


efagin:
👍 as a follow up item, we should consider opening PRs against all outstanding OEPs, after this merges, that adds a change history section to each of them.

nedbat:
:+1:

cpennington:
@efagin: I think I'll actually just add a section to each of them as part of this PR. No need for the extra step.

efagin:
@cpennington that's fine with me, I was just noting that you didn't _need_ to do all of that yourself! But if you want to, be my guest - it's appreciated. :)

jzoldak:
👍

PR #20, opened 2016-09-02
Spelling and other fixes

nedbat:


nedbat:
@cpennington Nothing substantive (I hope).  Alison has taught me not to over-link words.

oeps/oep-0005.rst
```diff
@@ -206,32 +206,29 @@ The prototype implementation was done using `Docker`_ for the containers, and
 Rationale
 =========
 
-Choice of `Docker`_
-~~~~~~~~~~~~~~~~~~~
```
nedbat: There was only one subsection, so no need for the header.

oeps/oep-0005.rst
```diff
@@ -49,11 +49,11 @@ because it was able to use the infrastructure that was built out for installing
...
+1. As new services are added to the Open edX ecosystem, the Vagrant image
    grows in size, requiring more system memory from the host machine.
 
-2. `Vagrant`_ imposes noticeable overhead during filesystem access, which
-   especially hurts any development that it doing file-heavy operations (such
+2. Vagrant imposes noticeable overhead during file system access, which
+   especially hurts any development that is doing file-heavy operations (such
    as manipulating static assets).
 
 3. The monolithic `Vagrant`_ image isn't particularly representative of the
```
cpennington: Should we remove these other links? Just link it from the top of the section?

nedbat: I wasn't that disciplined about where to remove links, just when I noticed there were a lot of them, I started removing them.  I think @lamagnifica said something like, link it once per page (or per top-level heading on the page?)

lamagnifica: Guidance from Mark S to the doc team was that multiple links to the same destination can be a nuisance to those using screen readers.

cpennington: Ah, good to know. I'll trim the duplicates.

jzoldak:
👍

PR #22, opened 2016-09-12
OEP-8 Unit Testing Tools and Frameworks

jzoldak:
Status: Under Review. Feedback requested. Please see https://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0001.html for context on the process.

### DIFF
--- a/oeps/oep-0008.rst
+++ b/oeps/oep-0008.rst
@@ -1,17 +1,17 @@
-==================================
-OEP-8: Unit Testing Best Practices
-==================================
+========================================
+OEP-8: Unit Testing Tools and Frameworks
+========================================
 
 +---------------+-------------------------------------------+
 | OEP           | :doc:`OEP-8 </oeps/oep-0008>`             |
 +===============+===========================================+
-| Title         | Unit Testing Best Practices               |
+| Title         | Unit Testing Tools and Frameworks         |
 +---------------+-------------------------------------------+
 | Last Modified | 2016-09-09                                |
 +---------------+-------------------------------------------+
 | Author        | Jesse Zoldak                              |
 +---------------+-------------------------------------------+
-| Arbiter       |                                           |
+| Arbiter       | Ben Patterson                             |
 +---------------+-------------------------------------------+
 | Status        | Draft                                     |
 +---------------+-------------------------------------------+
@@ -19,31 +19,40 @@ OEP-8: Unit Testing Best Practices
 +---------------+-------------------------------------------+
 | Created       | 2016-09-09                                |
 +---------------+-------------------------------------------+
-| `Resolution`  |                                           |
-+---------------+-------------------------------------------+
-| `Replaces`    |                                           |
-+---------------+-------------------------------------------+
-| `Replaced-By` |                                           |
-+---------------+-------------------------------------------+
+
 
 Abstract
 ========
 
-The edX ecosystem has grown from a handful of developers working on a few
-repos to many developers working on many repos. As such, questions arise as to
-tools, frameworks, and patterns to use in unit testing. This proposal is
-intended to codify best practices and to provide a reference for contributors
-when making choices.
+The edX ecosystem has grown from a handful of developers working on a few repos
+to many developers working on many repos. As such, questions arise as to tools
+and frameworks to use in unit testing. This proposal is intended to codify best
+practices and to provide a reference for contributors when making choices.
 
 
 Motivation
 ==========
 
-Having common unit testing best practices across the organization's repos will
-increase the productivity of developers by reducing the ramp-up cost when
-moving among repos. It will also increase the quality of the software by
-decreasing the cognitive load on developers who must call to mind the various
-tips and techniques while using the disparate tools and frameworks.
+Having common best practices for unit testing tools and frameworks across the
+organization's repos will increase the productivity of developers by reducing
+the ramp-up cost when moving among repos. It will also increase the quality of
+the software by decreasing the cognitive load on developers who must call to
+mind the various tips and techniques while using the disparate tools and
+frameworks.
+
+Note that edX already has a few dozen or more existing repos, and they are most
+likely currently not in full compliance with this OEP. This includes the edx-
+platform repo.
+
+Acceptance of this OEP does not imply that we will immediately work to bring all
+repos in full compliance. Rather we will track compliance using the Repository
+Metadata file specified in `OEP-2`_.
+
+Another reason for an exception might be to prototype a proposed new best
+practice. In that case, the openedx.yaml file could be marked as such (using the
+reason field).
+
+.. _OEP-2: https://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html
 
 
 Specification
@@ -90,29 +99,38 @@ If the Makefile gets too complex:
 * For Python projects, use it as a wrapper to `Paver`_ scripts
 * For JavaScript projectes, use it as a wrapper to `Gulp`_
 
-Standard Makefile Targets:
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+Standard Makefile Target Naming Conventions:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-+--------------+--------------------------------------------------------------+
-| target       | description                                                  |
-+==============+==============================================================+
-| help         | display a help message                                       |
-+--------------+--------------------------------------------------------------+
-| pip-compile  | update the ``requirements/*.txt`` files with the             |
-|              | latest packages satisfying ``requirements/*.in``             |
-+--------------+--------------------------------------------------------------+
-| requirements | install development environment requirements                 |
-+--------------+--------------------------------------------------------------+
-| quality      | check coding style                                           |
-+--------------+--------------------------------------------------------------+
-| test         | run all unit, integration, and acceptance tests for the repo |
-+--------------+--------------------------------------------------------------+
-| validate     | run tests and quality checks                                 |
-+--------------+--------------------------------------------------------------+
-| e2e          | run End to End tests against a deployed environment to       |
-|              | verify the system as a user, initiating their actions        |
-|              | through the UI                                               |
-+--------------+--------------------------------------------------------------+
++-------------------+--------------------------------------------------------------+
+| target            | description                                                  |
++===================+==============================================================+
+| help              | display a help message                                       |
++-------------------+--------------------------------------------------------------+
+| clean             | delete the artifacts created when running the validations    |
+|                   | and/or the application itself                                |
++-------------------+--------------------------------------------------------------+
+| pip-compile       | update the ``requirements/*.txt`` files with the latest      |
+|                   | packages satisfying ``requirements/*.in``                    |
++-------------------+--------------------------------------------------------------+
+| requirements      | install development environment requirements                 |
++-------------------+--------------------------------------------------------------+
+| requirements.prod | install production requirements                              |
+| (optional)        |                                                              |
++-------------------+--------------------------------------------------------------+
+| requirements.test | install requirements for testing                             |
+| (optional)        |                                                              |
++-------------------+--------------------------------------------------------------+
+| quality           | check coding style                                           |
++-------------------+--------------------------------------------------------------+
+| test              | run all unit, integration, and acceptance tests for the repo |
++-------------------+--------------------------------------------------------------+
+| validate          | run tests and quality checks                                 |
++-------------------+--------------------------------------------------------------+
+| e2e (optional)    | run End to End tests against a deployed environment to       |
+|                   | verify the system as a user, initiating their actions        |
+|                   | through the UI                                               |
++-------------------+--------------------------------------------------------------+
 
 .. _Travis CI: https://travis-ci.org/
 .. _PyPI: https://pypi.python.org/pypi
@@ -253,8 +271,8 @@ Code Coverage
 .. _JSCover: http://tntim96.github.io/JSCover/
 
 
-Rejected Alternatives:
-----------------------
+Rationale:
+----------
 
 `nose`_ / `nose2`_
   The original choice in the edx-platform was to use nose
--- a/oeps/oep-0008.rst
+++ b/oeps/oep-0008.rst
@@ -39,7 +39,7 @@ when making choices.
 Motivation
 ==========
 
-Having common unit testing best practices across the organization's repo will
+Having common unit testing best practices across the organization's repos will
 increase the productivity of developers by reducing the ramp-up cost when
 moving among repos. It will also increase the quality of the software by
 decreasing the cognitive load on developers who must call to mind the various
@@ -51,9 +51,18 @@ Specification
 
 CI (Continuous Integration) Framework
 -------------------------------------
-We use `Travis CI`_ for both public and private repos, with the exception of
-the edx-platform repo for which we use an internally-managed Jenkins
-environment.
+Our standard is to use `Travis CI`_ for both public and private repos.
+
+We make exceptions for repos which have more complicated CI requirements For
+these we use an internally-managed Jenkins environment.
+
+Examples of special cases for which we use Jenkins rather than Travis:
+
+* The setup and/or execution time for the validation is prohibitively long (e.g.
+  the edx-platform repo)
+* A repo contains secrets that we do not want to share with a third party (e.g.
+  secure repos)
+
 
 Guidelines:
 ^^^^^^^^^^^
@@ -62,8 +71,10 @@ Guidelines:
 
 * Use the container-base (sudo-less) CI environment, if possible
 
-* Execute tests for each supported Python version, and if appropriate also in
-  combination with different Django versions
+* Execute tests in each supported environment
+
+  * For Python projects, execute tests for each supported Python version(s)
+  * For Django, also test in combination with the supported Django version(s)
 
 * Configure with automated deployment of tagged releases to `PyPI`_
 
@@ -74,8 +85,10 @@ We use Makefiles for task management. The reason for this choice is that
 almost any system no matter the OS should be able to use them without any
 special software installation or configuration.
 
-If the Makefile gets too complex, use it as a wrapper to `Paver`_ scripts,
-which are written in Python.
+If the Makefile gets too complex:
+
+* For Python projects, use it as a wrapper to `Paver`_ scripts
+* For JavaScript projectes, use it as a wrapper to `Gulp`_
 
 Standard Makefile Targets:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -104,6 +117,7 @@ Standard Makefile Targets:
 .. _Travis CI: https://travis-ci.org/
 .. _PyPI: https://pypi.python.org/pypi
 .. _Paver: https://pythonhosted.org/Paver/
+.. _Gulp: http://gulpjs.com/
 
 
 Python Language Specific
@@ -132,12 +146,12 @@ Requirements
 * Requirements filename examples:
 
   * ``requirements.*``: contains the base requirements
+
   * ``test-requirements.*``: contains the requirements needed for writing and
     running tests
 
-* Editable requirements format example: ``-e
-  git+https://github.com/edx/django- oauth2-provider.git@0.2.7-fork-edx-6a#egg
-  =django-oauth2-provider==0.2.7 -fork-edx-6``
+* Editable requirements format example:
+  ``-e git+https://github.com/edx/django-oauth2-provider.git@0.2.7-fork-edx-6a#egg =django-oauth2-provider==0.2.7-fork-edx-6``
 
 Code Quality
 ^^^^^^^^^^^^
@@ -200,12 +214,30 @@ Code Quality
 * `ESLint`_ for static analysis in order to find problematic patterns or code
   that doesn’t adhere to style guidelines
 
+  * `edX ESLint config`_  for generating ESLint configuration files that verify
+    edX best practices
+
+  * Configure in either ``package.json`` or ``.eslintrc.json``
+
+  * Specify exclusions in ``.eslintignore``
+
+
 Test Framework
 ^^^^^^^^^^^^^^
 
 * `Jasmine`_ as the framework for testing JavaScript code
 
+  * `jasmine-jquery`_ matchers and fixture loader for Jasmine
+
+  * Jasmine-based helper classes from the testing folder of the `edX UI
+    Toolkit`_
+
 * `Karma`_ as the test runner for the JS tests
+
+  * We also use plugins to extend Karma. As this list could change, it is not
+    codified here. Please look at the ``devDependencies`` section of the
+    ``package.json`` file in existing edX repos to survey those that we are
+    currently using.
 
 Code Coverage
 ^^^^^^^^^^^^^
@@ -213,8 +245,11 @@ Code Coverage
 * `JSCover`_ for calculating code coverage
 
 .. _ESLint: http://eslint.org/docs/about/
+.. _edX ESLint config: https://github.com/edx/eslint-config-edx
 .. _Jasmine: http://jasmine.github.io/
+.. _jasmine-jquery: https://github.com/velesin/jasmine-jquery
 .. _Karma: https://karma-runner.github.io/
+.. _edX UI Toolkit: http://ui-toolkit.edx.org/
 .. _JSCover: http://tntim96.github.io/JSCover/
 
 
@@ -275,6 +310,19 @@ Rejected Alternatives:
     platform codebase, and we have good examples there of its usage,
     including tests.
 
+`JSHint`_ / `JSCS`_
+  The original choice in the edx-platform was to use JSHint for JavaScript code
+  linting.
+
+  However JSHint didn't have a good story for linting ES2015+ as we aimed to
+  upgrade to that version of JavaScript.
+
+  And JSCS, which was another leader in the JS code style linter and formatter
+  has merged with ESLint.
+
+  Thus we determined that ESLint is preferred to cover our immediate, short
+  term, and long term needs.
+
 `Coveralls`_
   The original choice for recording code coverage for trending
   reports was to write the coverage statistics from edx-platform test
@@ -298,6 +346,9 @@ Rejected Alternatives:
 .. _nose2: https://github.com/nose-devs/nose2/
 .. _rake: https://ruby.github.io/rake/
 .. _Invoke: http://docs.pyinvoke.org/
+.. _JSHint: http://jshint.com/
+.. _JSLint: http://www.jslint.com/
+.. _JSCS: http://jscs.info/
 .. _Coveralls: https://coveralls.io/
 
 
--- a/oeps/oep-0008.rst
+++ b/oeps/oep-0008.rst
@@ -1,0 +1,323 @@
+==================================
+OEP-8: Unit Testing Best Practices
+==================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-8 </oeps/oep-0008>`             |
++===============+===========================================+
+| Title         | Unit Testing Best Practices               |
++---------------+-------------------------------------------+
+| Last Modified | 2016-09-09                                |
++---------------+-------------------------------------------+
+| Author        | Jesse Zoldak                              |
++---------------+-------------------------------------------+
+| Arbiter       |                                           |
++---------------+-------------------------------------------+
+| Status        | Draft                                     |
++---------------+-------------------------------------------+
+| Type          | Best Practice                             |
++---------------+-------------------------------------------+
+| Created       | 2016-09-09                                |
++---------------+-------------------------------------------+
+| `Resolution`  |                                           |
++---------------+-------------------------------------------+
+| `Replaces`    |                                           |
++---------------+-------------------------------------------+
+| `Replaced-By` |                                           |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+The edX ecosystem has grown from a handful of developers working on a few
+repos to many developers working on many repos. As such, questions arise as to
+tools, frameworks, and patterns to use in unit testing. This proposal is
+intended to codify best practices and to provide a reference for contributors
+when making choices.
+
+
+Motivation
+==========
+
+Having common unit testing best practices across the organization's repo will
+increase the productivity of developers by reducing the ramp-up cost when
+moving among repos. It will also increase the quality of the software by
+decreasing the cognitive load on developers who must call to mind the various
+tips and techniques while using the disparate tools and frameworks.
+
+
+Specification
+=============
+
+CI (Continuous Integration) Framework
+-------------------------------------
+We use `Travis CI`_ for both public and private repos, with the exception of
+the edx-platform repo for which we use an internally-managed Jenkins
+environment.
+
+Guidelines:
+^^^^^^^^^^^
+
+* Configuration file: ``.travis.yml``
+
+* Use the container-base (sudo-less) CI environment, if possible
+
+* Execute tests for each supported Python version, and if appropriate also in
+  combination with different Django versions
+
+* Configure with automated deployment of tagged releases to `PyPI`_
+
+
+Task Management
+---------------
+We use Makefiles for task management. The reason for this choice is that
+almost any system no matter the OS should be able to use them without any
+special software installation or configuration.
+
+If the Makefile gets too complex, use it as a wrapper to `Paver`_ scripts,
+which are written in Python.
+
+Standard Makefile Targets:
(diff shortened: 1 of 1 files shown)


jzoldak:
@edx/testeng ready for review.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+environment.
+
+Guidelines:
+^^^^^^^^^^^
+
+* Configuration file: ``.travis.yml``
+
+* Use the container-base (sudo-less) CI environment, if possible
+
+* Execute tests for each supported Python version, and if appropriate also in
```
andy-armstrong: I'd like to see this generalized so that this OEP doesn't only apply to Python and Django repos. For example, I'd like to see this OEP cover JavaScript unit test guidelines. Can you rephrase this to start with something like "For Python projects, ...".

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+* Configure with automated deployment of tagged releases to `PyPI`_
+
+
+Task Management
+---------------
+We use Makefiles for task management. The reason for this choice is that
+almost any system no matter the OS should be able to use them without any
+special software installation or configuration.
+
+If the Makefile gets too complex, use it as a wrapper to `Paver`_ scripts,
```
andy-armstrong: For our front-end repos, we use a mixture of make and gulp.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+Standard Makefile Targets:
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
++--------------+--------------------------------------------------------------+
+| target       | description                                                  |
++==============+==============================================================+
+| help         | display a help message                                       |
++--------------+--------------------------------------------------------------+
+| pip-compile  | update the ``requirements/*.txt`` files with the             |
```
andy-armstrong: Can we generalize this to work for other languages? update-requirements?

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+|              | latest packages satisfying ``requirements/*.in``             |
++--------------+--------------------------------------------------------------+
+| requirements | install development environment requirements                 |
++--------------+--------------------------------------------------------------+
+| quality      | check coding style                                           |
++--------------+--------------------------------------------------------------+
+| test         | run all unit, integration, and acceptance tests for the repo |
```
andy-armstrong: Some of our repos also run ``quality`` when you run ``test``.

oeps/oep-0008.rst
maxrothman: +1, also want to point out the configuration and secure repos.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+
+JavaScript Language Specific
+----------------------------
+
+Code Quality
+^^^^^^^^^^^^
+
+* `ESLint`_ for static analysis in order to find problematic patterns or code
+  that doesn’t adhere to style guidelines
```
andy-armstrong: Use the edX eslint plugin to adopt the edX rules:

https://github.com/edx/eslint-config-edx

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+
+Specification
+=============
+
+CI (Continuous Integration) Framework
+-------------------------------------
+We use `Travis CI`_ for both public and private repos, with the exception of
+the edx-platform repo for which we use an internally-managed Jenkins
+environment.
```
maxrothman: We also use jenkins with any repo that contains secrets to avoid shipping those secrets to a third party.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+* `ESLint`_ for static analysis in order to find problematic patterns or code
+  that doesn’t adhere to style guidelines
+
+Test Framework
+^^^^^^^^^^^^^^
+
+* `Jasmine`_ as the framework for testing JavaScript code
+
+* `Karma`_ as the test runner for the JS tests
```
andy-armstrong: Use the edX UI Toolkit for Jasmine-based helper classes:

http://ui-toolkit.edx.org/testing/ajax-helpers/
http://ui-toolkit.edx.org/testing/spec-helpers/

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+  However the SLA for Coveralls posting the data back to GitHub does not meet
+  our needs, and developers were waiting many hours after the test execution
+  on the CI system had completed to get the feedback on their PRs. This led to
+  frustration and productivity hits.
+
+  In addition, developers starting using CodeCov on some repos and found that
+  they prefered the integration on PRs, including seeing the non-covered lines
+  in details. And the codecov Chrome Extension that shows coverage while in
+  the GitHub web app has also been found useful.
+
```
andy-armstrong: Maybe add a section on why we chose ESLint over JSHint and JSLint.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
++--------------+--------------------------------------------------------------+
+| target       | description                                                  |
++==============+==============================================================+
+| help         | display a help message                                       |
++--------------+--------------------------------------------------------------+
+| pip-compile  | update the ``requirements/*.txt`` files with the             |
+|              | latest packages satisfying ``requirements/*.in``             |
++--------------+--------------------------------------------------------------+
+| requirements | install development environment requirements                 |
```
maxrothman: perhaps "dev-requirements"? If these are the dev requirements, more might be added in the future.

oeps/oep-0008.rst
andy-armstrong: For Karma, it might be worth enumerating some of the plugins we use. Here are the ones from edx-platform:

    "karma-chrome-launcher": "^0.2.3",
    "karma-coverage": "^0.5.5",
    "karma-firefox-launcher": "^0.1.7",
    "karma-jasmine": "^0.3.8",
    "karma-jasmine-html-reporter": "^0.2.0",
    "karma-junit-reporter": "^0.4.1",
    "karma-requirejs": "^0.2.6",
    "karma-spec-reporter": "^0.0.20",

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+Code Quality
+^^^^^^^^^^^^
+
+* `ESLint`_ for static analysis in order to find problematic patterns or code
+  that doesn’t adhere to style guidelines
+
+Test Framework
+^^^^^^^^^^^^^^
+
+* `Jasmine`_ as the framework for testing JavaScript code
```
andy-armstrong: We also use ``jasmine-jquery``.

andy-armstrong:
Excellent stuff @jzoldak. Thanks for writing this all up.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+* Requirements files are kept either at root level or within a directory named
+  ``requirements``.
+
+* Direct dependencies are listed in ``requirements/*.in`` for use in
+  development and `Tox`_ virtual environments with as few version constraints
+  as possible.
+
+* Full pinned dependencies in ``requirements/*.txt`` are auto-generated via
+  ``make pip-compile``.
```
nedbat: We don't have much experience yet with pip-tools. I don't understand how privately installed packages (like pudb) will be managed while pip-tools tries to uninstall packages that shouldn't be there.  I also don't understand how we will manage pinning if pip-tools wants to do an all-or-nothing upgrade to the latest versions.  That will be changing many versions all at once.  Is that what we want?

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+tips and techniques while using the disparate tools and frameworks.
+
+
+Specification
+=============
+
+CI (Continuous Integration) Framework
+-------------------------------------
+We use `Travis CI`_ for both public and private repos, with the exception of
+the edx-platform repo for which we use an internally-managed Jenkins
```
benpatterson: I would reword this as something like "except for repos with more complicated CI requirements due to system time or system requirements for executing the validation."

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+The edX ecosystem has grown from a handful of developers working on a few
+repos to many developers working on many repos. As such, questions arise as to
+tools, frameworks, and patterns to use in unit testing. This proposal is
+intended to codify best practices and to provide a reference for contributors
+when making choices.
```
nedbat: Do we have some way to allow people to experiment with new tools? I like codifying what we *currently* think is the best way to do things, but tools will change over time.  Do we need to mention how to deal with that here?

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+
+Motivation
+==========
+
+Having common unit testing best practices across the organization's repo will
```
nedbat: "repos"

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+increase the productivity of developers by reducing the ramp-up cost when
+moving among repos. It will also increase the quality of the software by
+decreasing the cognitive load on developers who must call to mind the various
+tips and techniques while using the disparate tools and frameworks.
```
nedbat: edx-platform is often an exception to these best practices.  Should we mention it up front? Also, many repos have smaller discrepancies from these best practices.  I don't imagine we're expecting that all repos will be brought into compliance with these practices.  A statement one way or the other would be good.

oeps/oep-0008.rst
jzoldak: Yeah good point. I think this one is up for debate, I can see going either way. I'm very interested in peoples's opinions.

In general for this OEP the intended process is to decide what we want the standards to be, then circle back and enter tickets to get the non-conforming repos in line.

oeps/oep-0008.rst
jzoldak: Does JS have something equivalent? I'm not familiar.
If this only applies to python, I can state that in the description to make it more clear.

benpatterson:
This is moreso "recommended tools" than best practices. Should we just call it that?

oeps/oep-0008.rst
andy-armstrong: You can use ``npm update --save``:

https://docs.npmjs.com/cli/update

There's a similar command for Ruby too.

oeps/oep-0008.rst
jzoldak: I'm a bit wary to name plugins as they seem too fine-grained for this OEP. I'll put in a section that mentions Karma plugins and where to find the latest list of the ones that we are using.

oeps/oep-0008.rst
jzoldak: Great suggestion. Added.

oeps/oep-0008.rst
jzoldak: Good points. Refactored.

oeps/oep-0008.rst
jzoldak: Ah! Great. Added.

oeps/oep-0008.rst
jzoldak: Thanks. Added.

oeps/oep-0008.rst
jzoldak: Done

oeps/oep-0008.rst
jzoldak: added

oeps/oep-0008.rst
jzoldak: reworded this section

oeps/oep-0008.rst
jzoldak: That's a good point. I had imagined that future changes would be made in PRs to this OEP, and can include some wording to that effect.
Would that suffice, or were you getting at something different?

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+* A repo contains secrets that we do not want to share with a third party (e.g.
+  secure repos)
+
+
+Guidelines:
+^^^^^^^^^^^
+
+* Configuration file: ``.travis.yml``
+
+* Use the container-base (sudo-less) CI environment, if possible
```
benpatterson: Should we reference Travis docs here?

benpatterson:
To illustrate my previous point (or argue for it, perhaps), best practices that are not contained in this OEP include things around how and when to use coverage, how and when to mock, etc. 

I'm not sure where this is going to go, but while I'm thinking on it, I'm listing below some of the questions I have around "best practices" as opposed to recommended tools (apologies if this long list is coming off too strongly, but in case this is germane, I'm listing them here)

Coverage:
* Calculating and reporting code coverage (i.e., that you should do it, and that you should send it somewhere like `codecov`)
* Providing output files for test systems that can be consumed by downstream analytics or systems, like xunit-style testcases, or coverage output.
* What areas of code we do or do not consider when calculating coverage statistics (e.g., will we report coverage on */test/* files?)
* Any expected coverage number that the repo should be targeting
* Whether or not we combine JS and python unit tests when reporting coverage, or if OTOH we use separate numbers (I suspect this is a loaded topic; I vote for separate)

Other things related to unit tests:
* What happens for the systems that are not called out in this repo (e.g., Ruby, which has its own set of tools, bash testing or avoidance)?
* I think Ned asked about this, but we should have some reference for "if we find a better way, this is how we'll change this document"
* When to use a mock or when not to use it
* Definition of unit test vs other tests (e.g., what is out of scope for this OEP)
* If we call out travis and tox config, should we also call out IDE config?
* When to use fixtures and whatnot, how much production code one is can cannibalize

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+
+  * We also use plugins to extend Karma. As this list could change, it is not
+    codified here. Please look at the ``devDependencies`` section of the
+    ``package.json`` file in existing edX repos to survey those that we are
+    currently using.
+
+Code Coverage
+^^^^^^^^^^^^^
+
+* `JSCover`_ for calculating code coverage
```
benpatterson: Is that true? I thought some of the `karma` tools use other ways of calculating coverage out of the box.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+  unittest/unittest2). As we still want to use a framework that extends
+  unittest, the question then becomes whether edX standardize on nose2 or
+  pytest.
+
+  Note that edX developers have already begun using pytest in new repos,
+  having chosen it over nose2 for some or all of the reasons listed below.
+  Given this and lacking other barriers that would make it a bad choice, we
+  have decided to standardize on pytest.
+
+  Here are some of the reasons that we have chosen pytest:
```
benpatterson: I'm not sure I buy this, TBH. We never even evaluated nose2. I think in this case we are setting as a best practice what everyone is already doing. Each of the "reasons why pytest" is true for `nose2` as far as I understand it based on their docs. The only reason listed here which makes sense to me here is "it's what the industry is doing." That can be a valid reason because of reduced cognitive load and whatnot. But we should just redact the reasoning section, IMO, and simply point out that `pytest` is our tool of choice, even those many (really, almost all) of our repos are using nose.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+
+  Thus we determined that ESLint is preferred to cover our immediate, short
+  term, and long term needs.
+
+`Coveralls`_
+  The original choice for recording code coverage for trending
+  reports was to write the coverage statistics from edx-platform test
+  execution to DataDog.
+
+  This homegrown method was neither scalable no maintainable. As such, we
```
benpatterson: typo: "no maintainable"

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+  started integrating with Coveralls.
+
+  However the SLA for Coveralls posting the data back to GitHub does not meet
+  our needs, and developers were waiting many hours after the test execution
+  on the CI system had completed to get the feedback on their PRs. This led to
```
benpatterson: Feels like we're being overly critical of a free tool... Similar to `nose` I think it's ok to point out that we use coveralls in some repos, but our new norm is codecov. To me, this is because we can aggregate the data or the reports are preferred by developers. (Though, again I'd stress that this kind of information can be had locally, and codecov's main function IMO is to help us with reporting.)

So, really, I wonder if we scale back our "reasons why we did not choose tool `x`"

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
+
+  * For Python projects, execute tests for each supported Python version(s)
+  * For Django, also test in combination with the supported Django version(s)
+
+* Configure with automated deployment of tagged releases to `PyPI`_
+
+
+Task Management
+---------------
+We use Makefiles for task management. The reason for this choice is that
```
benpatterson: I know we made this decision a long time ago, but there is no testing infrastructure for `Make` that I know of, which makes for an unfortunate irony that our recommendation is not something that's easily testable.

oeps/oep-0008.rst
nedbat: For example.  Say we have a repo, edx/ned-tools. It uses pytest for testing, so it is in conformance with OEP-8.  Then a new test runner (holotest) appears on the scene, and we want to try it out.  We switch over ned-tools to use holotest.  From an OEP-8 point of view, that repo has regressed: it was in compliance, and now it is not.  Is this OK?  It's too soon to say that holotest is the best practice.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,374 @@
...
++---------------+-------------------------------------------+
+| Type          | Best Practice                             |
++---------------+-------------------------------------------+
+| Created       | 2016-09-09                                |
++---------------+-------------------------------------------+
+| `Resolution`  |                                           |
++---------------+-------------------------------------------+
+| `Replaces`    |                                           |
++---------------+-------------------------------------------+
+| `Replaced-By` |                                           |
```
cpennington: You can leave these lines out of the header if they aren't used.

oeps/oep-0008.rst
cpennington: My intention was to allow exceptions to OEPs with good reasons. One such reason might be that the exception is to prototype a proposed new best practice. In that case, the `openedx.yaml` file could be marked as such (using the `reason` field).

oeps/oep-0008.rst
cpennington: I used `make upgrade` for this in https://github.com/edx/repo-tools/blob/master/Makefile#L17

oeps/oep-0008.rst
cpennington: Since the Makefile is mostly for managing development tasks, I think using `requirements` to mean development requirements, and `production-requirements` (or something similar) to mean the subset that's for production might make sense. (See https://github.com/edx/course-discovery/blob/master/Makefile#L45-L53 for prior art).

oeps/oep-0008.rst
cpennington: I think having a way to run tests w/o quality, and then having a separate task that checks both, is a good way to go, since you often (as a developer) want to do one or the other, but not both.

oeps/oep-0008.rst
cpennington: For private packages, we can use a `private.txt` file, and have pip-sync install that, but have it `.gitignore`d so that it doesn't get checked in.

For pinning: we can explicitly pin things in the `.in` files, and `pip-compile` will respect that. `--upgrade` will call for upgrading everything, but we don't have to actually check in the changes to the `.txt` files in one chunk. It will also be easier to manage if we can do better about staying on top of dependency changes.

@jzoldak: There are differences in our strategies for libraries vs IDAs (in particular, libraries should use requirements files for their development requirements, but IDAs use them for production requirements as well). Do we need to call those differences out here?

oeps/oep-0008.rst
cpennington: Yeah, I agree, the reasons listed below don't really distinguish between `py.test` and `nose2`.

oeps/oep-0008.rst
andy-armstrong: That works for me. I see now that there is a validate target defined below that does just this.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
+==================================
+OEP-8: Unit Testing Best Practices
+==================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-8 </oeps/oep-0008>`             |
++===============+===========================================+
+| Title         | Unit Testing Best Practices               |
```
andy-armstrong: I wonder if this should be generalized to include more than just unit testing. There is a lot of useful material in here about linting, requirement management, end-to-end testing etc. If not, maybe some of that material doesn't belong in here.

oeps/oep-0008.rst
nedbat: It sounds like these are good details to include in the pip-tools best practices section.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,323 @@
...
+
+* Requirements filename examples:
+
+  * ``requirements.*``: contains the base requirements
+  * ``test-requirements.*``: contains the requirements needed for writing and
+    running tests
+
+* Editable requirements format example: ``-e
+  git+https://github.com/edx/django- oauth2-provider.git@0.2.7-fork-edx-6a#egg
+  =django-oauth2-provider==0.2.7 -fork-edx-6``
```
nedbat: Hyphens here seem to have some extra spaces around them.

oeps/oep-0008.rst
maxrothman: I'd think "requirements" would mean base requirements, "dev-requirements" would add dev-specific ones, "test-requirements" would add testing specific ones, etc.

oeps/oep-0008.rst
maxrothman: I think it's important to include at least some reasoning so future us knows why we made the decision. However, that reasoning might be "it seemed like a good idea at the time"

oeps/oep-0008.rst
nedbat: And actually, maybe a simpler example (with a simpler version number) would be clearer.

benpatterson:
Also, under the notion that this OEP is for recommended tools, do we recommend `diff-quality` or `diff-cover`?

estute:
Would this document be the appropriate place to mention mobile dev unit testing?

oeps/oep-0008.rst
nedbat: In addition, to @benpatterson's point, there are many practices around unit testing which this OEP won't cover: when do you mock, and what granularity to test, etc.  There are lots of unit testing techniques and approaches that this document isn't meant to discuss.

Perhaps this is less about unit testing, and more about repo structure and tooling?

oeps/oep-0008.rst
jmbowman: I see "requirements" already in use in at least a couple dozen of our repos, and "production-requirements" used in two.  "dev-requirements.txt" files exist in a few repos, but we don't yet have any make targets of that name.  I'm inclined to keep this documenting what we already do unless we're ready to go rename a bunch of make targets and update all scripts accordingly.

oeps/oep-0008.rst
jzoldak: Cool. Leaving this as-is, with quality, test, and validate as three separate targets.

oeps/oep-0008.rst
jzoldak: @maxrothman I think your suggestion of "requirements" for base requirements is a good one. I also agree with @jmbowman tho that the cost of changing the meaning from what we already have in most of our repos is too great for the benefit.
Let's leave 'requirements' for dev requirements.
I've added in requirements.prod and requirements.test (as optional) so that there will be those examples for people to follow.

cpennington: I think I'd rather state what we want, and then we can worry about migrating things whenever people get to it (no need to hurry). But that way we end up in a better state for new repos as we create them.

oeps/oep-0008.rst
jzoldak: Added

doctoryes:
Just as a general reminder:
OpenEdX now has cookiecutter repos for both a Django application & a top-level, independent Django application here:
https://github.com/edx/cookiecutter-django-app
https://github.com/edx/cookiecutter-django-ida
Once this OEP lands, an audit of those repos should be performed to guarantee they conform to the OEP. It might also be informative to browse the repos now to see if anything is missing from the OEP.

nedbat:
Just one data point: the IDA cookiecutter uses nose instead of pytest.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,372 @@
...
+^^^^^^^^^^^
+
+* Configuration file: ``.travis.yml``
+
+* Use the container-base (sudo-less) CI environment, if possible
+
+* Execute tests in each supported environment
+
+  * For Python projects, execute tests for each supported Python version(s)
+  * For Django, also test in combination with the supported Django version(s)
```
bjacobel: Should we add something about suggested node versions to test with here? In the UXPL and UITK we're testing with both node 4 (current LTS) and node 6 (becoming LTS in a month or two)

oeps/oep-0008.rst
```diff
@@ -0,0 +1,372 @@
...
+Task Management
+---------------
+We use Makefiles for task management. The reason for this choice is that
+almost any system no matter the OS should be able to use them without any
+special software installation or configuration.
+
+If the Makefile gets too complex:
+
+* For Python projects, use it as a wrapper to `Paver`_ scripts
+* For JavaScript projectes, use it as a wrapper to `Gulp`_
```
bjacobel: I'd like to see a discussion around Gulp as a build tool for our JS, as I don't feel like it's served us particularly well in places we've implemented it heavily like the UXPL.

cc @andy-armstrong  @dsjen

oeps/oep-0008.rst
```diff
@@ -0,0 +1,372 @@
...
+JavaScript Language Specific
+----------------------------
+
+Code Quality
+^^^^^^^^^^^^
+
+* `ESLint`_ for static analysis in order to find problematic patterns or code
+  that doesn’t adhere to style guidelines
+
+  * `edX ESLint config`_  for generating ESLint configuration files that verify
```
bjacobel: Nit - `eslint-config-edx` doesn't really _generate_ configuration files, it's a common config file that projects can implement or extend

oeps/oep-0008.rst
bjacobel: We also may want to mention Sinon in this section, which we're using to help test and mock async behavior / network calls in the UITK and I think in edx-platform?

oeps/oep-0008.rst
```diff
@@ -0,0 +1,372 @@
...
+^^^^^^^^^^^^^^
+
+* `Jasmine`_ as the framework for testing JavaScript code
+
+  * `jasmine-jquery`_ matchers and fixture loader for Jasmine
+
+  * Jasmine-based helper classes from the testing folder of the `edX UI
+    Toolkit`_
+
+* `Karma`_ as the test runner for the JS tests
```
bjacobel: Karma's a pretty configurable tool, it may be worth going into some of what we expect to get out of it here, eg: should tests run in the terminal (node) or in the browser? If so, what browsers? Multiple in parallel? If not, how do we approach debugging? Should we mock browser APIs? We have a few different approaches across different projects.

bjacobel:
to @benpatterson's comment:

> Whether or not we combine JS and python unit tests when reporting coverage, or if OTOH we use separate numbers (I suspect this is a loaded topic; I vote for separate)

I vote for separate as well, but I don't know how to implement it in Codecov; I think this document would be a good place to link to a resource on doing that, if one exists.

oeps/oep-0008.rst
dsjen: Thanks!  Gulp seems ok to me.  We use it in Insights to kick off karma, which then runs our jasmine unit tests backed by phantomjs.  Of course, I'm totally game for another system exists that works smoothly too!

bjacobel: In the UXPL we had [some issues](https://openedx.atlassian.net/browse/FEDX-216) stemming from (IMO) the Gulp scripts turning into a dependency rat's nest - synchronous and asynchronous tasks calling each other in a dependency graph that was hard to introspect, through which tasks that supposedly piped output to each other would fail to pass data. (We're now doing this all inside Webpack and have deleted ~60% of the Gulp scripts).

In my projects (personal and at previous jobs) I've used a combination of [npm run-script](https://docs.npmjs.com/cli/run-script) and Webpack to do all the asset management/test running/linting tasks, and I feel like that's covered the bases pretty well.

dsjen: Aside: there is a small spelling mistake with `projectes` in the doc.

Back to the topic at hand...  Maybe the statements on using paver and gulp scripts are a bit restrictive, as the UXPL is moving away from gulp, Insights uses gulp, yet doesn't use paver (and I don't think should).  Should these statements in themselves spawn another OEP?  Will any app be compliant?

oeps/oep-0008.rst
jzoldak: Updated with this example.

oeps/oep-0008.rst
jzoldak: done

oeps/oep-0008.rst
jzoldak: oh good catch! I hadn't realized that.

oeps/oep-0008.rst
jzoldak: redid this section

oeps/oep-0008.rst
jzoldak: fixed

oeps/oep-0008.rst
jzoldak: Fair points. Updated.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,399 @@
...
+* Requirements are managed via `pip-tools`_ and organized by which ones need
+  to be installed for different Python virtual environments.
+
+* Requirements files are kept either at root level or within a directory named
+  ``requirements``.
+
+* Direct dependencies are listed in ``requirements/*.in``.
+
+* Full pinned dependencies are auto-generated via ``make pip-compile`` and
+  saved in ``requirements/*.txt``. This file is tracked under version control.
```
cpennington: You should explicitly talk about the distinction between applications and libraries. In particular:

- libraries shouldn't pin their dependencies in setup.py
- applications should pin all of their dependencies in a requirements.txt file/folder
- libraries should provide a dev-requirements file with the requirements needed to build/develop on the library. That file may or may not be pinned.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,399 @@
...
+
+* For a developer to update package versions, she would run ``pip-compile
+  --upgrade`` and check in the changes to the .txt file(s).
```
cpennington: We should standardize on some make targets that run the pip-tools operations.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,399 @@
...
+* Requirements filename examples:
+
+  * ``requirements.*``: contains the base requirements
+
+  * ``test-requirements.*``: contains the requirements needed for writing and
+    running tests
+
+* Editable requirements format example:
+  ``-e git+https://github.com/edx/django-oauth2-provider.git@0.2.7-fork-edx-6a#egg =django-oauth2-provider==0.2.7-fork-edx-6``
+
```
cpennington: We may want to call out an explicit recommendation for the version numbers in the `egg=` stanza for repositories urls.

If you use `0.0`, pip will always try and reinstall the package, so you'll be guaranteed to get updates when they happen. Otherwise, you have to rely the version number of the git tag and the version number of the egg updating in tandem.

clintonb:
It seems that work has been done to implement this OEP, despite the OEP not being approved/merged. What's going on here? It doesn't seem right to implement an OEP before it's been approved.

* https://openedx.atlassian.net/browse/TE-1636
* https://github.com/edx/ecommerce/pull/1041

@estute @jzoldak @benpatterson

cpennington:
In general, I actually would expect some amount of prototyping of the implementation of any given OEP to happen before the OEP is approved.

oeps/oep-0008.rst
```diff
@@ -0,0 +1,419 @@
...
+  * Build status
+  * Coverage
+
+  .. _shields.io: http://shields.io/
+
+
+Python Language Specific
+------------------------
+
+Requirements
```
cpennington: It seems like maybe this should be a separate OEP.

jzoldak:
Closing this as it is too broad. Will open a more focused proposal in the future.

PR #23, opened 2016-09-15
OEP-9: User Authorization (Permissions)

jmbowman:
Presenting an outline of proposed best practices for implementing and using permission checks in Open edX Python code.  Mostly reuses existing APIs and packages, although there is one noted gap for implementing queryset filtering with rule-based permissions.  A rough sketch of a proposed package to fill that gap is provided, although custom model manager methods and Django REST Framework filter classes can be used in the meantime.

### DIFF
--- a/oeps/oep-0009.rst
+++ b/oeps/oep-0009.rst
@@ -3,17 +3,17 @@ OEP-9: User Authorization (Permissions)
 =======================================
 
 +---------------+-------------------------------------------+
-| OEP           | :doc:`OEP-9 </oeps/oep-0009>`             |
+| OEP           | :doc:`OEP-9 <oep-0009>`                   |
 +---------------+-------------------------------------------+
 | Title         | User Authorization (Permissions)          |
 +---------------+-------------------------------------------+
-| Last Modified | 2016-09-14                                |
+| Last Modified | 2016-10-11                                |
 +---------------+-------------------------------------------+
 | Author        | Jeremy Bowman <jbowman@edx.org>           |
 +---------------+-------------------------------------------+
 | Arbiter       | Eddie Fagin <eddie@edx.org>               |
 +---------------+-------------------------------------------+
-| Status        | Draft                                     |
+| Status        | Approved                                  |
 +---------------+-------------------------------------------+
 | Type          | Best Practice                             |
 +---------------+-------------------------------------------+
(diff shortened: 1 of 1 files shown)


maxrothman:
@bderusha this sounds similar to work you were doing at the hackathon at the last Open edX conference.

bderusha:
@maxrothman is correct. That hackathon project work lives in its entirety in [ECOM-4774](https://openedx.atlassian.net/browse/ECOM-4774) and includes a piece for redesigning how staff permissions work ([ECOM-4841](https://openedx.atlassian.net/browse/ECOM-4841)).  The gist being that staff should not need to be enrolled in a course to access the course content and that the enrollment table is for students.

It sounds like this work would provide a nice framework for where the new rules should live and how the rules around being able to access course content could work.  👍

oeps/oep-0009.rst
```diff
@@ -0,0 +1,315 @@
...
+The ``rules`` package can be added to existing packages with minimal impact,
+as it doesn't inherently change the outcome of any authorization checks.  In
+the handful of places where Open edX code actually uses the Django
+authorization API already, the permission implementation can be switched to
+use ``rules`` when convenient (at which point whichever implementation is
+already in use should be deactivated so they don't come to disagree over
+time.)  New code can be written to follow the guidelines in this OEP, and
+existing code can be gradually updated as the need arises.  Effort should
+be made to update all relevant code related to a new explicit permission,
+however, to avoid tempting developers into thinking they've fully updated the
```
doctoryes: Nit: This `however` might be better placed at the beginning of the sentence. It took me a few reads to parse the whole sentence due to the `however` in the middle.

doctoryes:
This standard would be a big improvement for the platform. The code samples greatly help - if adopted, maybe even a sample conversion could be done and documented.
👍

efagin:
@jmbowman I think this OEP is ready to be merged. Do you want to ask for another round of feedback or are you satisfied? (If you are, please update the state to Approved and I'll merge it.)

Thanks for the contribution! This is a really helpful OEP.

jmbowman:
@efagin I think we've gotten about all the feedback we can expect at this point.  Status is updated and it's ready to merge; I removed the copyright footer because the theme provides it as of #28 , so we may want to merge that first.

efagin:
Got it. Thanks - you have my go-ahead to merge after the theme's been
installed.

PR #24, opened 2016-10-03
Improve guidance on when to start an OEP

shaunagm:
I've separated out the steps for brainstorming an OEP & getting feedback from the actual process of submission.  The main reason for this PR is that I'd like clearer guidance as to what kind of suggestions make for good OEPs.  I've added a section for examples but cannot populate it myself, as I do not know what makes for a good or bad OEPs.

### DIFF
--- a/oeps/oep-0001.rst
+++ b/oeps/oep-0001.rst
@@ -57,9 +57,9 @@ Workflow
 .. contents::
   :local:
   :depth: 1
-
-Submitting an OEP
------------------
+  
+Initiating an OEP
+-----------------  
 
 The OEP process begins with a new idea for Open edX. It is highly recommended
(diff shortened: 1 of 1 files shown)


efagin:
@cpennington as the original OEP-1 author, would you mind working with @shaunagm to resolve this PR that updates OEP-1? If not, please just let me know and I'll find somebody else.

cpennington:
@efagin: Well, I failed to get this done before @shaunagm left, but I'll try and work up some examples.

efagin:
@cpennington bump

efagin:
I'm going to close this PR due to lack of activity. The changes are minimal and we should consider revisiting this if we experience commons barriers to entry on OEPs.

PR #26, opened 2016-10-04
OEP-10: Open edX Releases

nedbat:


nedbat:
@gsong @cpennington @edx/devops @jbarciauskas @clintonb  Opinions welcome before opening up for wider feedback.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+        ``parent-repo`` keys.
+
+    ``parent-repo``
+        The name of another repo, one that installs this repo via a
+        requirements file.  The parent repo's requirements file will indicate
+        the commit to tag in this repo.
+
+        .. note::
+            We believe we no longer need to tag dependent repos, so this may go
+            away.
```
cpennington: Does it make sense to tackle this question as part of this OEP? Or are you just looking to make an incremental step, for now?

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+containers.
+
+.. note::
+
+   We haven't determined how best to allow developers to configure which
+   services to run and which should be editable.
+
+The native and fullstack installation will use Ansible playbooks to install
+Open edX components onto the machine, and install Docker and some number of
+Docker containers.
```
cpennington: It seems like it would be good to be specific about the Ansible playbook. Right now, as an Open edX developer, I don't have enough information to determine for myself whether a given thing is installed in an Open edX release or not.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+========
+
+Open edX is packaged occasionally into releases. This document details the
+process.
+
+
+Motivation
+==========
+
+Open edX releases happen infrequently, every four to six months. Edx engineers
```
antoviaque: Would it be a good place to discuss switching to a different release schedule? As mentioned in other places, it would be useful for this to be predictible, so that upgrades can be scheduled once for all, and make it easier for course teams of Open edX instances to plan their courses schedules around this?

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+Installation methods
+--------------------
+
+There are three supported installation methods:
+
+- Devstack: for developers planning to modify Open edX code.
+
+- Native: for adopters looking to experiment with Open edX in a non-production
+  environment.  Also useful as a starting point for people designing their
+  production deployment.
```
antoviaque: Why not explicitly target a production environment, instead of "native"? It seem strange that no production environment is supported.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+
+- Fullstack: same as Native, but in a pre-packaged Vagrant image.
```
antoviaque: What is the goal of fullstack? Is this actively used?

oeps/oep-0010.rst
antoviaque: +1. And it would be good to make sure at least one of them allows for proper explicit support of development environments. One possibility is to use the mono-instance stateless production playbook we have worked on recently: https://github.com/edx/configuration/pull/3370

oeps/oep-0010.rst
jbarciauskas: We do not have the resources to provide a level of testing that would make us feel confident in recommending a particular production configuration. We do not want others relying on our recommendation rather than doing their own testing. If members of the community wanted to volunteer to support and test a particular set of configuration and deployment choices, that would be welcome.

oeps/oep-0010.rst
nedbat: We are definitely discussing working to a more specific schedule. Frankly, I'm not sure we have the organizational maturity to do it, though.

oeps/oep-0010.rst
nedbat: Yes, i will settle it one way or the other in this OEP. I wanted to capture the uncertainty.

oeps/oep-0010.rst
nedbat: Yes, I meant to leave a placeholder here for the specifics. 

@antoviaque We may have to define our terms better.  What do you mean by "development environments", and how is it different than devstack?

antoviaque: @nedbat Sorry, I meant production environment, this was a remark related to https://github.com/edx/open-edx-proposals/pull/26/files/d7dcce76695531ca440049ea3198d1feb4bfe1df#diff-8d9ecff0109cafc3244e70b99e8fc687

oeps/oep-0010.rst
antoviaque: We do this type of testing for our own clients, whenever there is a new release. So that could be something we do more formally as part of the release process, that could be a quite natural contribution for us. How would you see this working? What would be our responsibilities, precisely?

CC @bradenmacdonald for when he'll be back, so he knows about this : )

oeps/oep-0010.rst
nedbat: @antoviaque ah, ok: we are hesitant to be too opinionated about production environments.

oeps/oep-0010.rst
nedbat: @antoviaque this is a very generous offer, thanks! We should talk about what you already do, and how we could make use of it.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+- Feature: large applications have major features which are configurable, and
+  should be called out explicitly if their support status is different than the
+  application they are a part of.
+
+
+
+Levels of support
+-----------------
+
+Units in Open edX are either either *included* or *supported* in a release.
```
gsong: Double "either" typo here.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+
+To be **included** in an Open edX release, a component, facility or feature
```
gsong: What's a facility? It's the first time the concept is introduced.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+
+    Possible keys:
+
+    ``ref``
+        The name of the release-from branch in this repo. This is the branch
+        that will be tagged when an Open edX release is made.
+
+    ``requirements``
+        The path to a requirements.txt-style file in this repo. This file will
+        be consulted to find the commit to tag in dependent repos indicated by
```
gsong: I don't really understand what this means. And looking at https://github.com/edx/edx-platform/blob/master/requirements/edx/github.txt didn't make it any more clear to me.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,202 @@
...
+----------------
+
+Our installations are based on Anisble playbooks.  Up until the Eucalyptus
+release, all three supported installation methods were single-machine: all of
+the Open edX software was installed and ran on a single machine, either a
+Virtualbox image, or a native machine.
+
+That model does not scale up as the number of services and applications grows.
+Newer services are supporting Docker for installation.  Eventually, we would
+like to be in an all-Docker world where an installation is just a constellation
```
gsong: Is this true even for production deployment? There are plenty of horror stories for using Docker in production. Maybe it's more accurate to describe as multi-machine/container installations?

oeps/oep-0010.rst
nedbat: fixed.

oeps/oep-0010.rst
nedbat: This should just be "component," since that term is defined above to include a variety of things.

oeps/oep-0010.rst
nedbat: The lines in github.txt mention specific repos, and specific commits in those repos. The idea was that if one of those repos (django-wiki, let's say) had a parent-repo key pointing to edx-platform, then edx-platform would have a requirements key indicating the requirements file, and we could find django-wiki's entry in edx-platform's requirements file, and then know what commit to tag in django-wiki.

But: I now think the tagging of dependent repos isn't needed, which I mentioned in the ``parent-repo`` section, but forgot to mention here.

oeps/oep-0010.rst
nedbat: I meant for this to only apply to the supported installation methods, which doesn't include production.

antoviaque: Even if it isn't a supported method, others might have the same question reading this. If we state the long-term plan, it seem a sane assumption that this would then also apply to production setups? If not, it would be clearer (and more maintainable) to make the distinction explicitly, and specify what the difference would be.

oeps/oep-0010.rst
bradenmacdonald: @nedbat @antoviaque I'm not sure if there's been any follow-up on this, but I think this idea is great, and would be happy to develop a plan so that our team can help with more formal and systematic testing of Open edX releases in the future. As @antoviaque mentioned, it's something we do already for our clients.

antoviaque: @bradenmacdonald No follow-up yet on this no - from last week's community hangout ficus seem to be in the early planning stage for December, so I imagine that would be part of the discussions/planning as we get closer to this. It will be good to define a specific scope btw, to be sure we appreciate the size of the work before committing to it : )

oeps/oep-0010.rst
nedbat: I've come around to the idea of dropping fullstack.

oeps/oep-0010.rst
```diff
@@ -0,0 +1,209 @@
...
+- Feature: large applications have major features which are configurable, and
+  should be called out explicitly if their support status is different than the
+  application they are a part of.
+
+
+
+Levels of support
+-----------------
+
+Units in Open edX are either *included* or *supported* in a release.
```
efagin: Nit: "Units" should be "Components" to be consistent with your terminology.

PR #28, opened 2016-10-11
PLAT-1084 Use edx-sphinx-theme for output

jmbowman:
Use the new edx-sphinx-theme package for the output, and remove the explicit copyright footer from each page which is now put there by the theme.  (This will need to be done for the new OEPs already created in pending PRs too.)  Also flesh out the .gitignore file with a few IDE and editor artifacts, etc.

Before merging, we should make sure that the Read the Docs project is configured to build in a virtualenv and install the dependencies from requirements.txt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,12 @@
 _build
+
+# IDEs and text editors
+*~
+*.swp
+.idea/
+.project
+.pycharm_helpers/
+.pydevproject
+
+# OS X artifacts
+*.DS_Store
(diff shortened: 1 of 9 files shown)


jmbowman:
@cpennington @efagin Up for reviewing this?

efagin:
@cpennington I can't answer @jmbowman's question:

> Before merging, we should make sure that the Read the Docs project is configured to build in a virtualenv and install the dependencies from requirements.txt.

Can you answer that and review as well? Feel free to merge when done, if you're satisfied.

cpennington:
I just updated the project to install requirements.txt into a virtualenv.

PR #30, opened 2016-10-14
Two minor fixes for OEP-5

nedbat:


nedbat:
@cpennington easy-peasy.

cpennington:
:+1:

PR #31, opened 2016-10-17
Pin requirements so the theme tweaks will be used

nedbat:


nedbat:
@cpennington @jmbowman Surely no one could object to this?

PR #32, opened 2016-10-19
OEP-11: front end best practices

andy-armstrong:
https://openedx.atlassian.net/browse/FEDX-273

This is an early draft of an OEP for front end best practices.

You can view the proposal in rendered form here:
 
https://github.com/edx/open-edx-proposals/blob/andya/front-end-best-practices/oeps/oep-0011.rst

FYI @AlasdairSwan @alisan617 @bjacobel @cahrens @dsjen

### DIFF
--- a/oeps/oep-0011.rst
+++ b/oeps/oep-0011.rst
@@ -1,0 +1,240 @@
+======================================
+OEP-11: Front End Technology Standards
+======================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-11 </oeps/oep-0011>`            |
++---------------+-------------------------------------------+
+| Title         | Front End Technology Standards            |
++---------------+-------------------------------------------+
+| Last Modified | 2017-03-06                                |
++---------------+-------------------------------------------+
+| Author        | Andy Armstrong <andya@edx.org>            |
++---------------+-------------------------------------------+
+| Arbiter       | Matt Drayer <mattdrayer@edx.org>          |
++---------------+-------------------------------------------+
+| Status        | Draft                                     |
++---------------+-------------------------------------------+
+| Type          | Best Practice                             |
++---------------+-------------------------------------------+
+| Created       | 2016-10-19                                |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+This document describes the technology standards for developing
+front end applications and components for Open edX. It covers the techniques
+and approaches that are currently recommended, as well as the technologies
+and libraries that should be used.
+
+Motivation
+==========
+
+The rapid pace of development in front end tooling has created opportunities
+to greatly improve the quality of the Open edX user and developer experience.
+After an assesment of industry best practices, the edX Front End Working Group
+(FedX) has made a number of technology recommendations for libraries, frameworks
+and tooling to modernize the edX front end.
+
+Within this document, the phrase "front end" is used to mean any part
+of the platform that is shown to users. This encompasses views rendered in
+Python on the server, interactive interfaces written using JavaScript, and
+CSS styling.
+
+Specification
+=============
+
+Technology Selection
+~~~~~~~~~~~~~~~~~~~~
+
+1. **Server-side content should be rendered with Django Templates**
+
+   **Rationale**: There are many template languages available for Django,
+   but the simplest option is to use the built-in Django template engine.
+   The Open edX codebase has a mixture of Django and Mako templates, but the
+   former are easier to reason about because they don't support arbitrary
+   code evaluation. See `Reasons for rejecting Mako`_ for a more detailed
+   explanation.
+
+   **Exception**: Mako templates can continue to be used within edx-platform
+   for consistency with the existing code base. This is because the base
+   templates and theming support are all provided via Mako, so it is too
+   much to expect a new feature to be implemented with a different framework.
+   There is much desire to replace Mako within edx-platform so this
+   exception may eventually be removed.
+
+2. **Target the latest standardised JavaScript version (ECMA-262)**
+
+   **Rationale**: edX JavaScript should be written consistent with the latest
+   ECMA-262 specification in order to ensure future support, the largest
+   community and the availability of modern features. To support this syntax
+   in older browsers, use `Babel`_. Babel may also be configured to add syntax
+   extensions widely adopted by the community of our recommended framework
+   (e.g., `JSX`_).
+
+   **Exception**: Much of edX's existing front end code is written conformant
+   to the version of ECMA-262 released in 2009 (ES5). Files written in ES5
+   should be gradually converted to the newer standard as new development in
+   those feature areas requires.
+
+   **Note**: edX previously used CoffeeScript, but its use has now been
+   deprecated. Community interest in TypeScript has also grown, but it and
+   other languages that do not follow the ECMA-262 spec should not be used. See
+   `Reasons for rejecting CoffeeScript`_ and
+   `Reasons for rejecting TypeScript`_ for more detail.
+
+3. **JavaScript libraries should be installed via npm**
+
+   **Rationale**: It is important that JavaScript libraries are kept
+   up-to-date, and one key component is to make it as simple as possible
+   to perform upgrades. Projects that use npm maintain all of their
+   library requirements in a single `package.json` file, which can be
+   easily changed as the versions change.
+
+4. **JavaScript code should follow the edX ESLint configuration**
+
+   **Rationale**: In order to standardize and enforce Open edX's JavaScript
+   coding style across multiple codebases, edX has published an ESLint
+   configuration that provides an enforceable specification. The
+   `edX ESLint Config`_ is made available as an npm package that can be
+   installed into any Open edX package.
+
+   **Exception**: The `edX ESLint Config for ES5`_ may be used where ES5
+   is in use. Both configs may be used in the same codebase through the
+   use of `ESLint configuration cascading`_.
+
+5. **Use React and Redux when building new client-side applications**
+
+   **Rationale**: React should be used for building new UIs, as it is
+   widely adopted by the community and strikes a balance between
+   flexibility and feature richness. For state management of complex
+   client-side interactions, Redux should be used. This library was chosen
+   because it sees strong use in the React community, but is also flexible
+   enough to be used in situations where a hybrid React/Backbone architecture
+   exists.
+
+   See `React at edX`_ for more details on the evaluation process that led to
+   this technology selection. Other modern frameworks considered for use
+   included Polymer and Angular. See `Reasons for rejecting Angular`_  and
+   `Reasons for rejecting Polymer`_ for an exploration of why these
+   alternatives do not work as well as React for edX.
+
+   **Exception**: When building new applications within edx-platform, it
+   is currently acceptable to use Backbone.  Backbone is a somewhat old
+   technology in the JavaScript world and has seen a rapid drop off in usage
+   within the community. The plan is to incrementally adopt React within
+   edx-platform as it can interoperate cleanly with Backbone. More
+   documentation on this plan can be found in
+   `Modernizing the edX front end stack`_.
+
+6. **JavaScript should be bundled using Webpack**
+
+   **Rationale**: `Webpack`_ is the tool of choice for code optimization and
+   bundling, as it is widely used throughout the industry, and because
+   it seamlessly handles modern code bases as well as all of the older
+   technologies used by edX, such as `AMD Modules`_. Webpack should be
+   implemented to handle as much of the "asset pipeline" as possible,
+   rather than passing this responsibility on to Django.
+
+7. **JavaScript dependencies should be managed with ES2015 Modules**
+
+   **Rationale**: JavaScript module systems allow front end code to specify
+   its dependencies and be grouped into bundles that minimize the assets
+   needed to provide page functionality. The most prevalent module syntax
+   is currently `ES2015 Modules`_, which should be adopted everywhere
+   edX code is written to the ES2015 spec or later.
+
+   **Exception**: Much of edX's existing (ES5) JavaScript uses the older
+   `AMD Modules`_ syntax for modules. AMD Modules are interoperable
+   with ES2015 Modules if Webpack is used for bundling, so AMD is an
+   acceptable module definition if the code must remain ES5.
+
+8. **CSS should be generated using Sass**
+
+   **Rationale**: Sass's SCSS syntax is an extension of CSS that adds power
+   and elegance to the basic language. It makes the maintenance of large
+   CSS files more manageable though the ability to use variables, mixins,
+   imports and more. In particular, it makes theming possible with the
+   ability to override variables that define colors, fonts etc.
+
+   You can find out more about Sass in the official `Sass documentation`_.
+
+Rejected Alternatives
+=====================
+
+.. _Reasons for rejecting Angular:
+1. **Angular**
+
+   `Angular`_ was a popular web framework in its 1.x iteration, and its
+   maintainer (Google) recently released Angular 2. Angular 1 was not considered
+   due to its support status. Angular 2 was rejected due to its opinionated
+   characteristics including its use of TypeScript (a compile-to-JavaScript
+   language) and the difficulty of incrementally integrating Angular into an
+   existing application (e.g., alongside legacy Backbone). Consensus in the
+   front end community is that Angular has a higher learning curve than
+   alternatives like React, which matches poorly with edX's limited front end
+   developers and larger community of infrequent JavaScript contributors.
+
+.. _Reasons for rejecting Polymer:
+2. **Polymer**
+
+   `Polymer`_ is a framework built around `Web Components`_, a set of draft
+   extensions to the HTML and DOM specs for UI encapsulation. Web Components
+   are a very interesting area of emergent Web technology, but are not yet
+   finalized as a spec or supported in many browsers. Because of this, the
+   Polymer community is small and performance issues exist. As Web Components
+   become more mature, edX may reevaluate their use, either with Polymer or
+   without.
+
+.. _Reasons for rejecting CoffeeScript:
+3. **CoffeeScript**
+
+   The Open edX codebase made heavy use of CoffeeScript for several years,
+   but its use at edX has now been officially deprecated. Most of the language
+   benefits of CoffeeScript were made available in JavaScript as of the ES2015 spec,
+   and CoffeeScript's community has largely moved on.
+
+.. _Reasons for rejecting TypeScript:
+4. **TypeScript**
+
+   Several variants of typed JavaScript have grown in community popularity
+   recently, with `TypeScript`_ as the most popular. edX believes adding
+   typing to our JavaScript would raise the barrier to entry for less
+   experienced front end developers, and that using TypeScript in particular
+   would lock us into a tool with an uncertain future (as we experienced with
+   CoffeeScript). If typing is an absolute necessity for a feature
+   (e.g., an API client or sensitive business logic), `Flow`_ is recommended,
+   as it can be incrementally added and evaluated alongside non-typed code.
+
+.. _Reasons for rejecting Mako:
+4. **Mako**
+
+   There are a number of reasons why Django templates is preferred to Mako:
+
+   * Mako templates are not safe by default, meaning that it is very easy
+     to introduce cross-site scripting vulnerabilities.
+   * Mako templates are hard to unit test because they require that the
+     whole Django context be available in case the template tries to
+     access it. Django templates only have access to the context object
+     that is passed to it, so it is very easy to simulate all possible
+     scenarios.
+
+.. Cross-references
+.. _AMD Modules: https://github.com/amdjs/amdjs-api/wiki/AMD
+.. _Angular: https://angular.io/
+.. _Babel: https://babeljs.io/
+.. _edX ESLint Config: https://github.com/edx/eslint-config-edx/tree/master/packages/eslint-config-edx
+.. _edX ESLint Config for ES5: https://github.com/edx/eslint-config-edx/tree/master/packages/eslint-config-edx-es5
+.. _ESLint configuration cascading: http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy
+.. _ES2015 Modules: http://www.ecma-international.org/ecma-262/6.0/#sec-imports
+.. _ES2017: https://tc39.github.io/ecma262/
+.. _JSX: https://facebook.github.io/react/docs/introducing-jsx.html
+.. _Modernizing the edX front end stack: https://openedx.atlassian.net/wiki/display/FEDX/Modernizing+the+edX+front+end+stack
+.. _Polymer: https://www.polymer-project.org/
+.. _React: https://github.com/facebook/react
+.. _React at edX: https://openedx.atlassian.net/wiki/display/FEDX/React+at+edX
+.. _Sass documentation: http://sass-lang.com/
+.. _Typescript: https://www.typescriptlang.org/
+.. _Webpack: https://webpack.github.io/
+.. _Web Components: https://www.webcomponents.org/


oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   consistency with the existing code base. Unfortunately there has been too
+   much investment in Mako to consider replacing it at this point.
+
+2. **ECMAScript 2015 should be the supported version of JavaScript**
+
+   Rationale: ECMAScript is the standardized version of JavaScript,
+   and the most recently ratified version is ECMAScript 2015. It is
+   commonly referred to by ES2015, or by ES6 (its original name). Most
+   browsers now provide native support for ES2015, but it is generally
+   recommended to use `Babel`_ to transpile ES2015 code into ES5 which is
```
jmbowman: This `Babel` link doesn't have a target URLs defined.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   plans are being made to evaluate its usage. See `React at edX`_ for
+   more details.
+
+2. **How should offline support be provided?**
+
+   For many mobile users, it would be very useful if they could continue
+   their Open edX courses while offline. This is true both for true mobile
+   apps as well as for responsive web applications. This is an area that
+   has not been explored much up to now, but which will only become more
+   important going forward.
```
jmbowman: The last time I did a serious investigation into this (2-3 years ago), it was a minefield of missing functionality and inconsistency across browsers...and getting worse over time rather than better.  It's probably worth looking at again with the advent of service workers and such, but I recommend going into it with healthy skepticism about the odds of being able to create a good offline experience via an actual web browser (vs. a native app which includes HTML rendering capability).

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+Rejected Alternatives
+=====================
+
+Note: this section needs fleshing out...
+
+.. _Reasons for rejecting Bootstrap:
+1. **Bootstrap**
+
+ - size
+ - not accessible
```
jmbowman: There is an [accessibility plugin](https://github.com/paypal/bootstrap-accessibility-plugin), but I'm not sure if that improves things enough to make a difference.  Might be worth reviewing if we do allow some Bootstrap usage as described below.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   Rationale: There are many template languages available for Django,
+   but the simplest option is to use the built-in Django template engine.
+   The Open edX codebase has a mixture of Django and Mako templates, but the
+   former are easier to reason about because they don't support arbitrary
+   code evaluation. See `Reasons for rejecting Mako`_ for a more detailed
+   explanation.
+
+   **Exception**: Mako templates should be used within edx-platform for
+   consistency with the existing code base. Unfortunately there has been too
+   much investment in Mako to consider replacing it at this point.
```
jmbowman: Django 1.8+ supports concurrent use of multiple template engines.  Is it worth investigation of a gradual transition to Django templates over time?

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+
+.. _Reasons for rejecting CoffeeScript:
+2. **CoffeeScript**
+
+   The Open edX codebase made heavy use of CoffeeScript for several years,
+   but it has now been officially deprecated. Most of the language
+   benefits of CoffeeScript are now made available via ES2015.
+
+.. _Reasons for rejecting Mako:
+3. **Mako**
```
jmbowman: Was Jinja2 also rejected, or just not considered yet?  I'm not necessarily advocating it, but it's popular enough that it probably deserves a mention here.

oeps/oep-0011.rst
bjacobel: Might be worth speaking a little bit about the mechanisms by which we'll accomplish this, e.g. Service Workers and IndexedDB

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+.. _JavaScript frameworks:
+
+1. **Which more modern JavaScript framework should be supported?**
+
+   The JavaScript space is very fast moving as can be seen in
+   `The State of JavaScript 2016 Survey`_. There is a lot of momentum
+   around `React`_, and it appears to have many qualities that would
+   make it a good fit for edX. It is too early to make this call, but
+   plans are being made to evaluate its usage. See `React at edX`_ for
+   more details.
```
bjacobel: This section is light on details about the reasons we're thinking about a change:
- Backbone is no longer under active development (https://i.bjacobel.com/20161019-frxhr.png)
- The community of Backbone support and actively supported tools/libraries/components built for Backbone is shrinking, while other communities are growing (https://i.bjacobel.com/20161019-5r91u.png)
- It's hard to hire Backbone developers
- Backbone doesn't provide the kind of debugging tools or overall development experience of more modern frameworks
- Backbone doesn't have good stories for many of the recommendations raised in other parts of this OEP, like server-side rendering or ES2015+

Just saying "the space is fast moving" kind of implies that we're only looking at new technologies because they're shiny and new, but the reality is that Backbone has reached EOL and we're very much in need of a new recommendation.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   The most common approach is to run the same JavaScript code both in the
+   browser and using Node on the server. This requires the use of a JavaScript
+   framework that does not perform direct manipulation of the DOM, but instead
+   works with an abstraction of the desired HTML. For example, Netflix is
+   seeing great performance gains by using `React`_ in this way, as can be
+   seen in the blog post `Making Netflix Faster`_.
+
+4. **Where do web components fit with Open edX applications?**
+
+   Web Components is a W3C recommendation that allows developers to create
```
bjacobel: Web components are kind of a distraction, IMO - they're [currently not supported in the huge majority of browsers](http://caniuse.com/#feat=custom-elements), and the most commonly suggested framework for polyfilling and using them (Polymer) is not widely used outside of Google or performant outside of Chrome. In a couple years we might revisit this but right now the only really feasible way to get our own custom components + a large library of 3rd-party components is with React.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   `Polymer Project` provides a library that allows web components to be
+   built in most modern browsers. It seems that there is a useful overlap
+   between web components and Open edX's `XBlocks`_
+
+5. **Which build tools and asset pipeline should be used?**
+
+   There are already a variety of technologies used within Open edX,
+   none of which is a clear winner. Django Pipeline is used by
+   edx-platform, Django Compressor is used by some of the Ecommerce
+   applications, and Webpack is prevalent in pure front-end implementations.
```
bjacobel: We'll need to reconcile the endorsement of RequireJS Optimizer above with the endorsement of Webpack here.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+ - not accessible
+ - too inflexible
+ - might want to consider compatibility to allow mixing Bootstrap components
+   into pattern library pages
+
+.. _Reasons for rejecting CoffeeScript:
+2. **CoffeeScript**
+
+   The Open edX codebase made heavy use of CoffeeScript for several years,
+   but it has now been officially deprecated. Most of the language
```
bjacobel: > it has now been officially deprecated

We say in one or two places "don't write new coffeescript" but to me deprecation means something else - removing support for it entirely, ie, removing the coffeescript compiler.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+
+4. **JavaScript code should follow the edX ESLint configuration**
+
+   Rationale: In order to standardize and enforce Open edX's JavaScript
+   coding style across multiple codebases, edX has published an ESLint
+   configuration that provides an enforceable specification. The
+   `edX ESLint Config`_ is made available as an npm package that can be
+   installed into any Open edX package.
+
+5. **Front-end interactivity should be added using Backbone**
```
alisan617: I am getting the impression that this is comparing apple to orange. React alone can't replace a MVC framework, w/o others such as Flux or Redux. Is the direction that we are moving towards Backbone + React or Redux + React and why the choice.

It's worth mentioning that React community is growing rapidly, that Open Source community that knows React is also the same. Maybe that is one of the arguments.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+Best Practices
+~~~~~~~~~~~~~~
+
+1. **Front ends should render themselves server-side as much as possible**
+
+   Rationale: Server-side rendering gives the content to the user
+   immediately without requiring additional round-trips. This can make
+   for large performance improvements on slow networks or mobile devices.
+
+   If a dynamic client-side experience is needed, then the server-side
```
thallada: I think we should describe the different types of "front-ends". We have many server-side rendering Django apps and then we have single-page apps. I think this document would still be useful for server-side apps, because we also talk about things like the pattern library, accessibility, internationalization, and etc. which are all still relevant to fully server-side apps.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+   pages, no matter the framework that is used. It also provides a clean
+   programming model that enforces a clean separation between models
+   and views.
+
+   Note: Backbone is a somewhat old technology in the JavaScript world
+   and has seen a rapid drop off in usage within the community. See
+   :ref:`JavaScript Frameworks` below for the open question of which
+   of the many alternate frameworks should become the new Open edX standard.
+   For now, Backbone is the library of choice because so much of the
+   Open edX codebase uses it.
```
clintonb: Agreed. We should use the best tool for the job. That may be Backbone. It may also be something else.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+
+6. **JavaScript dependencies should be declared via RequireJS**
+
+   Rationale: Open edX is a large, complicated code base, and it is
+   getting more tangled over time. RequireJS allows code to declare
+   the dependencies it requires, rather than forcing every page to load
+   all possible libraries. It also supports intelligent bundling using
+   RequireJS Optimizer, which can package together all required libraries
+   into a single file for faster performance.
```
clintonb: I want to see the ECOM Team move toward Webpack. Our IDAs with static assets waste a lot of time moving unnecessary files around due to the nature of Django-compressor and RequireJS.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Rejected Alternatives
+=====================
+
+Note: this section needs fleshing out...
+
+.. _Reasons for rejecting Bootstrap:
+1. **Bootstrap**
+
+ - size
```
clintonb: Relative to our own pattern library and toolkit Bootstrap is pretty small and feature-ful. We should revisit the decision to reject Bootstrap. It gives us a lot out-of-the-box that we have yet to replicate.

oeps/oep-0011.rst
bjacobel: +1. When I talk about a React app I mean React + Redux.

oeps/oep-0011.rst
andy-armstrong: Fixed, thanks.

oeps/oep-0011.rst
andy-armstrong: I'm not aware of any evaluation of Jinja2. I've added it as an open question.

oeps/oep-0011.rst
andy-armstrong: IMO this would be too difficult given the number of Mako templates, and how they interact with comprehensive theming. My preference would be to have new front ends be implemented in separate repos to mitigate this problem.

oeps/oep-0011.rst
andy-armstrong: That's interesting background, thanks. I haven't done any investigation myself, but it does seem that service workers et al are getting more mature. Certainly there are a lot of companies that are delivering offline experiences now.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,299 @@
...
+
+   Rationale: Sass is an extension of CSS that adds power and elegance to
+   the basic language. It makes the maintenance of large CSS files more
+   manageable though the ability to use variables, mixins, imports and more.
+   In particular, it makes theming possible with the ability to override
+   variables that define colors, fonts etc.
+
+   You can find out more about Sass with the official guide
+   `Sass: CSS with Superpowers`_.
+
```
bjacobel: One more thing to add to this section (following discussion in today's FedX WG):

**Target Node 6.7 LTS and NPM 3**
For JavaScript that runs on the server, or for tooling/build pipeline dependencies, use the most recent LTS NodeJS version, and use its bundled version of NPM to install dependencies.

oeps/oep-0011.rst
andy-armstrong: To me, deprecated means that it is still available but that it is not recommended for use going forward. Typically you mark code as deprecated in one major release, and then you remove it in the next one.

andy-armstrong:
@bjacobel @clintonb @jmbowman @bradenmacdonald @dsjen @thallada thanks for all the incredible feedback. Here's my second draft that is so much better with all of your ideas in it, and I hope that it captures everybody's intent. I still need to work some more on the rejected alternatives section, but for now I feel that this is ready for wider review. I'd love to hear what you think.

oeps/oep-0011.rst
bjacobel: Mmm, yeah, that's fair.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   Note: the other modern library that many organizations are using
+   is Angular 2.  See `Reasons for rejecting Angular`_ for an exploration
+   as to why this does not work as well as React for edX.
+
+   Exception: when building new applications within edx-platform, it
+   is currently acceptable to use Backbone.  Backbone is a somewhat old
+   technology in the JavaScript world and has seen a rapid drop off in usage
+   within the community. The plan is to incrementally adopt React within
+   edx-platform as it can interoperate cleanly with Backbone.
```
bjacobel: Changes to this section look really good to me.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+6. **JavaScript dependencies should be managed with AMD modules**
+
+   Rationale: Open edX is a large, complicated code base, and it is
+   getting more tangled over time. It is important that applications be
+   composed as modules, and AMD is the standard module system for JavaScript.
```
bjacobel: > AMD is the standard module system for JavaScript

Eeeeeehhhhhhh..... ES2015 modules are closer to being "the standard" given that they're officially endorsed by ECMA and browsers plan to implement them natively. AMD was, at the time that we picked it, a widely adopted standard, but to say that it is "the standard system" isn't 100% accurate IMO.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   It allows code to declare the dependencies it requires, rather than forcing
+   every page to load all possible libraries.
+
+   A majority of edX code uses `RequireJS`_ (and its associated libraries
+   RequireJS Text and RequireJS Optimizer) to handle AMD modueles. In
```
bjacobel: > modueles

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   particular, RequireJS Optimizer supports intelligent bundling to package
+   together all required libraries into a single file for faster performance.
+   Note that other edX applications use `Webpack`_ which can also process
+   RequireJS-style dependencies as well as the newer Common JS standard.
```
bjacobel: > the newer Common JS standard

CommonJS is an older spec than AMD - CommonJS came out in 2008-2009ish, AMD came out as a result of a group of CommonJS users that wanted to add async loading to the CommonJS spec. The newest standard is ES2015 modules, which syntactically look like CommonJS but use `import` rather than `require`. Using Webpack + Babel you can read modules written using any of the three specs (and generate code that interops with any/all of them), but most new code is being written with ES2015 modules.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+Rejected Alternatives
+=====================
+
+Note: this section needs fleshing out...
+
+.. _Reasons for rejecting Angular:
+1. **Angular**
+
+   Needs fleshing out...
```
bjacobel: Some bullet points that might provide guidance for fleshing this out:

- it's a very heavy, opinionated framework that doesn't interop with other systems nicely. In comparison, React is known for being able to integrate with other frameworks (especially Backbone) making a migration less painful. [Here's a good talk from folks at Hubspot about their transition from Backbone to React](http://product.hubspot.com/blog/moving-backbone-to-flux-react) (and I think Airbnb went through this as well).
- Angular's community management hasn't been great. Their abandonment of Angular 1 left a lot of Angular shops in the lurch and there's a general consensus that the plans for 2 were not well thought-out (originally not planned to have any kind of upgrade path from v1 to v2, originally planned to require a totally new language called AtScript, later changed after community outrage to be a soft requirement on Typescript). There's still quite a bit of ill will and I don't trust Google to not pull the same thing again. In comparison, Facebook uses React virtually everywhere in their own apps and they have a very strong commitment to supporting it, breaking the APIs as little as possible and releasing codemods for it when they do because it hurts them internally as much as it hurts the community.
- none of our Fed developers know Angular 2 or Typescript AFAIK, by contrast we have a decent amount of React experience on staff already
- this maybe leans too heavily on my personal Opinions but I was an Angular 1 developer for a year or so and I hated it - too much reliance on framework magic, very difficult to test and debug things, terrible performance characteristics on complex pages, etc. I haven't used Angular 2 but it just seems to me that it would be hard for it to not bring that legacy along in the new version unless they fired all the people that designed the first one.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+   An evaluation was performed,
+ - size
+ - not accessible (although there is this an accessibility plugin)
+ - too inflexible
+ - might want to consider compatibility to allow mixing Bootstrap components
+   into pattern library pages. See the edX Bootstrap theme provided by
+   the community as one approach.
+
+   Needs fleshing out...
```
bjacobel: I can't speak to the reasons we chose not to use Bootstrap because I wasn't around when we made that decision initially, but to me one of the biggest things against sites built with Bootstrap is they feel cheap and routine. Bootstrap feels to me like a house in a development where everything looks nice and functionally *works*, but you look next door and across the street and you realize all of the houses are the exact same. Maybe their driveway is on the other side of the house from the one next door and they have different shingles and some nice flowers in the front yard, but you can tell walking past that they all have basically the same floorplan, and they all feel cheap and interchangeable.

Accordingly, most Bootstrap sites I see are either hackathon projects or ProductHunted landing pages for MVP Uber-for-dogs startups. I'm not aware of any really serious sites with good branding/UX that outsource that work to Bootstrap. Developing our own style and components and building blocks is a lot more work but it's really important in showing who we are.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   Rationale: Open edX is used by learners all over the world, and so
+   it is critical that all new functionality support localization. This means
+   that every string in an application should be marked for translation,
+   and it also includes ensuring that all new user interfaces will switch to
+   right-to-left mode when the language requires it.
+
+   You can learn more about localization with the Developer's Guide
+   chapter `edX Internationalization`_.
+
+10. **Front end implementations should support theming**
```
cahrens: Theming is not a requirement that we have had before-- we never discussed this in the capa feedback work, for example. In my opinion, this is much lower in priority compared to the other items in this list, and it does not have company buy-in at this time.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+6. **JavaScript dependencies should be managed with AMD modules**
+
+   Rationale: Open edX is a large, complicated code base, and it is
+   getting more tangled over time. It is important that applications be
+   composed as modules, and AMD is the standard module system for JavaScript.
+   It allows code to declare the dependencies it requires, rather than forcing
+   every page to load all possible libraries.
+
+   A majority of edX code uses `RequireJS`_ (and its associated libraries
```
cahrens: What is the underscore for?

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+   You can find out more about Sass with the official documentation here:
+   `Sass`_.
+
+8. **Applications should be built for production using Webpack**
+
+   Rationale: Webpack is the tool of choice for code optimization and
+   bundling, as it is widely used throughout the industry, and because
+   it seamlessly handles modern code bases as well as all of the older
+   technologies used by edX, such as `RequireJS`_.
```
cahrens: Underscore again.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+
+2. **How should offline support be provided?**
+
+   For many mobile users, it would be very useful if they could continue
+   their Open edX courses while offline. This is true both for true mobile
+   apps as well as for responsive web applications. This is an area that
+   has not been explored much up to now, but which will only become more
+   important going forward. In particular, we should look at the great
+   capabilities provided by service workers. See the article
+   `Service Workers: An Introduction`_ for a good starting point.
```
cahrens: Underscore issue.

oeps/oep-0011.rst
clintonb: Theming is quite important for white-label and multi-tenancy.

cahrens:
@andy-armstrong This is looking great!

Did you consider splitting "Best Practices" into 2 groups-- "Requirements" and "Best Practices", for example? Some of these things are requirements for merging into edx-platform, and some are not.

For example, Requirements:
1. Designs should be fully accessible
2. Front end applications should support localization
3. Front end code should be safe from cross site scripting

Best practices: pretty much everything else.

Or maybe adding AAA, AA, A sort of specifications for each to indicate how critical it is?

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   templates and theming support are all provided via Mako, so it is too
+   much to expect a new feature to be implemented with a different framework.
+   There is much desire to replace Mako within edx-platform so this
+   exception may eventually be removed.
+
+2. **ECMAScript 2015 should be the supported version of JavaScript**
+
+   Rationale: ECMAScript is the standardized version of JavaScript,
+   and modern front end libraries typically use ES2015 (aka ES6).
+   Most browsers now provide native support for ES2015, but to support
```
ormsbee: We have a practice for amending OEPs. We could call this out as something that will be shifting, and do amendments as we decide to upgrade. I'd rather keep it updated through explicit changes though, and not say that we'll follow the spec in lock step (even if that's our general intent).

oeps/oep-0011.rst
ormsbee: Are the performance issues mostly related to the shadow DOM polyfill? Are they better now that they've switched over to Shady DOM?

oeps/oep-0011.rst
```diff
@@ -0,0 +1,412 @@
...
+   between web components and Open edX's `XBlocks`_
+
+5. **Should HTTP/2 be used in preference to asset bundling?**
+
+   The Hypertext Transfer Protocol Version 2 (HTTP/2) enables more efficient
+   network communication between clients and servers. In particular, it
+   allows for multiple requests to be handled with the same TCP connection,
+   which removes most of the benefits of asset bundling. It is not yet
+   clear whether this is ready for full scale adoption, but it would reduce
+   a lot of complexity from application's asset pipelines.
```
ormsbee: FWIW, we already serve all static assets through our CDNs over HTTP/2, and client support is widespread. I would highly encourage us to start unraveling the asset bundling we're doing. In fact, one of the best steps we can do to help with courseware loading performance on mobile is to further break apart the XModule JS bundle.

oeps/oep-0011.rst
andy-armstrong: Thanks for suggesting this, @clintonb. This prompted a very useful discussion in the FedX working group, and we agree that it is time to re-evaluate Bootstrap. This is particularly in light of Bootstrap 4 which addresses a number of the issues that stopped us using it originally (it is switching to Sass, now uses ems/rems instead of pixels, and has a more flexible grid system).

We now have a FedX story to spike on using Boostrap 4 to implement the Pattern Library which we plan on picking up as soon as we've reached agreement on the acceptance criteria.

https://openedx.atlassian.net/browse/FEDX-284

When I do my next revision of this OEP I'll update it to describe this newer thinking.

bjacobel:
Hi folks,

I'm going to be taking this OEP over from Andy. It stalled out in review last year due to ambitious scope, and we'd like to make another push on it so it can start informing the projects we'll tackle this year.

I've reduced the scope of this OEP to just technology choices related to building font-end applications. We will be following up this OEP with another that addresses UI/UX-related issues (like our recommendations for building responsive components using Bootstrap) later, and have also moved some content from earlier drafts of this OEP directly into the Developer's Guide (e.g., meet accessibility requirements, manage history correctly, etc.)

This draft is an almost complete rewrite, so if you've reviewed before please let us know your thoughts again and hopefully we can get this moving forward once more. @mattdrayer will be serving as the arbiter for this OEP.

mattdrayer:
In the spirit of a complete re-write I shall also re-read https://open-edx-proposals.readthedocs.io/en/latest/oep-0001.html !!!

bjacobel:
@jmbowman @clintonb @dsjen @bradenmacdonald @cahrens @ormsbee Hi all! Would you folks who reviewed an earlier draft of this OEP mind giving it another look over sometime this week? As I said earlier, it's a pretty much complete rewrite, so I'm sure there will be comments/errors/omissions.

If there are no comments for, say, two weeks from when I added the third draft, I'll take that as a sign we're good to go and ask @mattdrayer to merge.

ormsbee:
@bjacobel: What's here looks fine to me. There's some devil-in-the-details stuff like how we reconcile self-contained and distributed Django apps with webpack handling of assets, but that can definitely be worked out elsewhere. I'm assuming that this will be one of the more amended OEPs over time, given the pace of things in the JS community, and so we'll have opportunity to reassess things like yarn vs. npm if we see a value in switching later.

There are other things that I think should be in this OEP (e.g. Bootstrap, WGAC 2.0 compliance, RTL, responsive, the goal of having a shared pattern library). Even if it doesn't go into the details on these, I think that they're important to mention in a doc that outlines our high level direction for front end development. I've heard the argument that some of these are common sense things that are just baseline and do not need to be put into an OEP, but I don't think we should make those assumptions.

Anyhow, that's just my $0.02. Ultimately, I have less day to day involvement with front end development than just about anyone on that list of reviewers, so please don't consider my remarks to be blocking.

bjacobel:
@ormsbee Those are great comments! We are planning on writing another OEP (I will probably start it this week) that's more UI focused and will encompass things like using Bootstrap as the basis for reusable components that establish a themeable, accessible, internationalized, consistent look-and-feel across edX frontend. We're still working out if this OEP need a title change to differentiate it more as "the JavaScript stack OEP" vs. the "how do we create and share UI OEP" but both areas will definitely be covered (and there may be overlap, etc.)

As an aside, I'm definitely pushing for Yarn :) We haven't been particularly burned by the problems it solves (vs. NPM) so I haven't been pushing for it as hard as React, Webpack, etc., but it's definitely on the radar and I was on the fence about including it here.

clintonb:
Nit: it would be nice to squash these commits to a reasonable number, rather than merge as seven.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,227 @@
...
+   browsers support.
+
+   **Exception**: Much of edX's existing front end code is written conformant
+   to the version of ECMA-262 released in 2009 (ES5). Files written in ES5
+   should be gradually converted to the newer standard as new development in
+   those feature areas requires.
+
+   **Note**: Use of standards-compliant ECMAScript features should be
+   preferred to the use of external libraries such as Underscore or
+   compile-to-JavaScript languages such as CoffeeScript. See
```
bradenmacdonald: Nit: I think TypeScript might be a more useful example to mention here - I imagine it's much more tempting for modern devs than CoffeSript.

Also, this is a little inconsistent with the above paragraph, since Babel is essentially a compile-to-Javascript language framework, and its use is endorsed. (And e.g. TypeScript can also serve the same function as Babel since it also transpiles ES6 to ES5)

bjacobel: Thanks for your comments @bradenmacdonald, I think it is a good idea to add a section about our consideration of Typescript (we did consider it internally, briefly) and other typing solutions, as they are something that's getting a lot of attention. (The gist of the section, I think, will be that we don't recommend adding typing to JavaScript at the moment, but are keeping an eye on Flow.)

Fair point, Babel (depending on how you configure it) lets you turn some things that aren't in the ES spec into JavaScript (e.g., JSX). I guess saying "you must use ES262-compliant JS" is a little too strict, as it prevents us from writing React how we want to. But we want to make sure that the language features that we're using are standards-track at the same time. I don't know - I'll take a stab at rewriting this section and ping you to take another look when I have something that fits all these parameters a little better?

ormsbee: @bjacobel: In terms of rationale, is it fair to say that we're being more conservative with our front end technology stack because we have to accommodate a large pool of back end and full stack developers at edX, and we want to minimize the ramp up necessary for them to contribute code?

bjacobel: @ormsbee Yep! We want to use the language "variant" that will most allow people at all levels  of JS familiarity to get things done, and going into that is a determination of what has the largest community, best documentation, most polished toolchain and strongest chance of still being supported in 5+ years.

Currently rewording this section to work those ideas in.

bjacobel: @ormsbee @bradenmacdonald See what you think of ba63ed5.

(I'm going to move the "use the language's stdlib instead of importing Underscore where possible" into the Developer Guide, as it's common sense.)

cahrens:
👍 Though it would be nice to provide reference implementations where possible (repos where the particular technologies are being used).

bjacobel:
Just so everyone knows, I plan on merging this by the end of the week (the 31st) if there is no further discussion - I've communicated this to the #open-edx-proposals Slack channel as well.

oeps/oep-0011.rst
```diff
@@ -0,0 +1,240 @@
...
+   Polymer community is small and performance issues exist. As Web Components
+   become more mature, edX may reevaluate their use, either with Polymer or
+   without.
+
+.. _Reasons for rejecting CoffeeScript:
+3. **CoffeeScript**
+
+   The Open edX codebase made heavy use of CoffeeScript for several years,
+   but its use at edX has now been officially deprecated. Most of the language
+   benefits of CoffeeScript were made available in Javascript as of the ES2015 spec,
```
vkaracic: Nit: ``Javascript`` > ``JavaScript`` as it seem to be the format of the word that you chose.

mattdrayer:
Hi everyone -- we've had some fantastic discussion around the topic of Front-End Best Practices for Open edX, and the resulting OEP-11 writeup does a good job of capturing our decisions.  As Arbiter 😈, I'm approving this document and will merge it at 4pm Eastern (~1hr), unless @bjacobel or I hear otherwise.

PR #34, opened 2016-12-06
OEP-12: Pluggable User Interfaces

andy-armstrong:
There is a long standing desire to allow developers to extend edx-platform by
providing plugins, rather than by modifying the platform's code directly.
XBlocks provide this ability for new types of courseware components, but
courseware is only one of many ways in which developers need to extend the
platform (see https://openedx.atlassian.net/wiki/display/AC/Feature+Plugins+for+edX+Platform for a detailed enumeration).

In this OEP, I propose adding Django Component Views as the means for developers to create pluggable user interfaces in Django and install them into edx-platform or other such code
bases.

You can see the rendered version of the OEP here: https://github.com/edx/open-edx-proposals/blob/160fc3cb92b2103c7aa39841f7eea8a9c8fce47d/oeps/oep-0012.rst

FYI: @nasthagiri @robrap @ormsbee @cpennington @bradenmacdonald @e-kolpakov


oeps/oep-0012.rst
```diff
@@ -0,0 +1,305 @@
...
+
+The Instructor Dashboard provides a set of user interfaces to allow an
+instructor to manage the features of their course. Currently these user
+interfaces must be added directly to edx-platform, even if the feature
+itself lives in a Python library that has been separately installed.
+
+It would be preferable if any feature implemented as a Python library could
+contribute new sections to the Instructor Dashboard through plugins. The
+dashboard would show a list of the feature that can be managed, and then
+cliking on a feature would present the user with the pluggable interface.
```
bjacobel: "clicking"

oeps/oep-0012.rst
```diff
@@ -0,0 +1,305 @@
...
+1. **It should render its content into a section of a page**
+
+   A pluggable user interface should be able to return the HTML, CSS and custom
+   JavaScript necessary to render its content. In addition, it should be able to
+   declare the JavaScript libraries that it depends upon.
+
+   Python and JavaScript APIs will be provided so that the consumer of a UI
+   block can render it. These APIs will ensure that the necessary libraries have
+   been loaded, and that the necessary CSS and JavaScript files have been added
+   to the page (without duplication).
```
bjacobel: The "without duplication" part here will be interesting to implement. UIBlocks will have to both be self-contained and namespace all of their own dependencies so they don't conflict with each other, but also communicate somehow to resolve that only one instance of any one library version needs to be loaded.

Maybe this is an implementation detail, but I'm already thinking about how to handle this - do we make the page serve a common bundle, and then each UI block requests only their unique dependencies as a separate bundle? Or are all UIBlock dependencies on a page hoisted into one megabundle?

bjacobel:
Oops, sorry, that review was meant to go in as just a comment, not an approval

oeps/oep-0012.rst
andy-armstrong: I should have made this explicit, but XBlocks already has support for this through its Fragment class. That's one of the major benefits to building this into the XBlock infrastructure.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,305 @@
...
+========
+
+There is a long standing desire to allow developers to extend edx-platform by
+providing plugins, rather than by modifying the platform's code directly.
+XBlocks provide this ability for new types of courseware components, but
+courseware is only one of many ways in which developers need to extend the
+platform (see `Feature Plugins for edX Platform`_ for a detailed enumeration).
+UI Blocks are a proposed mechanism to allow developers to create pluggable user
+interfaces in Python and install them into edx-platform or other such code
+bases.
```
bradenmacdonald: What "other such code bases" do you have in mind? Open edX IDAs? Any python learning-related web apps? Any web/mobile app of any kind? I think that we should have a very clear answer to that question.

My understanding is that XBlocks were designed so that in theory they [could be used on many different platforms to create many types of online educational experience](http://edx.readthedocs.io/projects/xblock-tutorial/en/latest/overview/introduction.html#overview) but in practice because they depend on Python and implementing an XBlock runtime is a bit obscure, the edX LMS/CMS is the only place where they are used. (Excluding xblock-sdk which is not an end-user platform.)

Since the XBlock API is not specific to the edX LMS/CMS and the set of available "services" is also limited, there are many things that one wants to do when creating courseware components that XBlocks cannot do. So today we have many XBlocks that don't conform to the XBlock spec because they are just calling into the LMS/CMS internal API directly (and [other issues](https://openedx.atlassian.net/wiki/display/~dormsbee/XBlock+Runtime+Isolation)). I could imagine the same thing happening with UI blocks.

andy-armstrong:
@nasthagiri @robrap @cpennington @ormsbee @cahrens @efagin @bjacobel @bradenmacdonald @e-kolpakov I've published a second draft of the OEP which now recommends Django Component Views as implemented in my recent hackathon project with @cahrens. I'd love your feedback when you have time to look at this. My hope is to start using Django Component Views for our upcoming work on the new course outline view sometime in mid-January.

FYI @clintonb has kindly agreed to be the arbiter for this OEP, so I've added him to the metadata for the page. Thanks, @clintonb!

mulby:
Scope question - is this intended to describe a UI extension mechanism for all IDAs or just edx-platform?Specifically in Insights we've been searching for an extension mechanism. I could see Otto wanting something similar. I would be happy to contribute an example use case for Insights.

Also - I'm assuming we wouldn't use DjangoComponentViews for pure REST APIs (like Analytics API). Those could theoretically be extended using vanilla djangoapps? Does it make sense to call that out explicitly in here?

andy-armstrong:
@mulby great questions. I intended this proprosal to broader than just edx-platform but as you point out I haven't captured such use cases. In fact I'm hoping that the wider Django community might find the library useful too which is why I went with this name and am hosting it in its own repo. I would really appreciate it if you could contribute example use cases for Insights, and I'd love to see some from ECOM too. In fact, I was thinking that this approach might have been a good alternative for the Programs UI that would have allowed it to live in its own repo as Jim had originally intended.

Regarding your second question: yes, these component views are designed purely for HTML APIs. I'll make that clearer.

dsjen:
@andy-armstrong, @mulby -- Use cases for Insights might be for including additional charts or tables in-page or perhaps additional summary cards.

screen shot 2017-01-04 at 12 31 37 pm

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+Use Case 1: Course Tabs
+-----------------------
+
+A very common need for new features is to be able to add a new course tab for
+the learner. Examples of this include Discussions, Student Notes, and Teams.
+There is a supported pluggable entry point for course tabs (see
+`Adding a new course tab`_) but this requires that the feature add a new URL
+for the feature, and render the whole page at that URL using a Mako template.
+It would be preferable for a course tab to be able to register a pluggable user
+interface that will be presented as a tab, and for the implementor not to need
```
bjacobel: > implementor

This might be a British English thing - the more common spelling is implementer.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+much in the way of additional review required to be confident in the behavior
+of the app.
+
+Consider how web components and Django Component Views overlap
+--------------------------------------------------------------
+
+There is a lot of interest in the `Web Components`_ specification and how
+the capabilities could be used to provide reusable components. In particular,
+it seems that web components provide a level of isolation that XBlocks do not,
+and that which could greatly enhance the security of the edX platform.
```
bjacobel: Web Components isolate your CSS and HTML inside a shadow root, but your JS still runs in the main thread on the global execution context AFAIK. I don't think web components can claim any security benefits in the current spec (they'd have to execute JS on a web worker in order to do this).

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+should instead be made into a plugin. One option would be to create a new
+CourseDiscussionXBlock to display the "Course" tab, but as described above there
+are downsides to forcing this to be an XBlock. It would be cleaner if the edX
+discussion library provided a pluggable user interface that can be dropped into
+each of the contexts in which it is needed. The component would be provided
+with the context for authorization and to customize itself appropriately.
+
+An interesting question is how can such components could share course-scoped
+settings that the instructor specified in Studio. See `OEP-6 Context-scoped XBlock
+Fields`_ for the design of context-scoped fields in XBlock.
```
bradenmacdonald: If we're using django apps as a plugin framework, for something like discussions, could it make more sense just to store settings using Django models (with a course key field) or ConfigurationModels?

mulby:
@dsjen I was thinking more along the lines of the Credo Reference work, where they define entirely new reports that are accessible from the course home and the menu system, but I haven't really thought it through.

andy-armstrong:
@bradenmacdonald Thanks for all the great feedback on various drafts of this document. Here are my thoughts:

> Do we want to define standard types of UI blocks? ("This is a discussion system"; "this is a course outline view"). Probably not, but worth mentioning.

I was considering whether to add tagging to this proposal, but I think that it is probably cleaner to keep Django Component View as simple as regular Django Views, and then use a different mechanism (probably Stevedore plugins) to register different types. I did this in my Instructor Dashboard hackathon demo, where I introduced a new ``InstructorFeature`` plugin type:

https://github.com/edx/edx-platform/pull/14122/files

I've added a description of this in my upcoming draft.

> How does a UI block state where it can appear in the UI? For example, if I want to write a UI block that is a course tab, how do I define that it's a course tab? Or how do I create a UI Block that can appear on the home page (Dashboard) but not in any course-specific contexts?

This is definitely where I see Stevedore plugins coming in. We already have a ``CourseTab`` entry point declared, so that plugin can now declare which view should be used to render itself.

> Can different UI blocks be enabled in different courses? For example, if some courses want to use the standard Open edX discussion framework, and others want to use an alternate discussion framework that uses UI Blocks to display Discourse discussions. If so, how is that configured?

This is an interesting question that I don't have a great answer for, other than to say that I feel that the enabling of features is a higher level capability than this OEP. We could solve this through XBlocks, or we might need something even higher than that since discussions might be more than one XBlock.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+
+* There is also the concept of changing course announcements to be a new type of
+  discussion post
+
+In an ideal world, Open edx Discussions would be a separately-installable
+feature that would provide plugins for all of the functionality listed above. An
+Open edX operator could choose not to install edX discussions and choose to
+install different collaboration features instead.
+
+Currently the discussion board is a Django view built into edx-platform, but it
```
andy-armstrong: Good point. I've condensed this down in my next draft.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+   A pluggable user interface should be able to return the HTML, CSS and custom
+   JavaScript necessary to render its content. In addition, it should be able to
+   declare the JavaScript libraries that it depends upon.
+
+   Python and JavaScript APIs will be provided so that the consumer of a
+   component can render it. These APIs will ensure that the necessary libraries
+   have been loaded, and that the necessary CSS and JavaScript files have been
+   added to the page (without duplication).
+
+2. **It should be able to get the data that it needs to render its content**
```
andy-armstrong: Thanks @robrap. 

> 1. Do you want to mention some part of this as part of the "Proposal" section, stating that this requirement will be fulfilled through standard Django mechanisms.

I'll beef up this section to make this more explicit.

> 2. Do we want to mention that there are or will be specific best practices and appropriate services to call inside edx-platform from one of these pluggable views, the details of which are out of scope of this OEP. The main point being to reiterate that someone should not feel free to do anything they please within the limits of Python and Django inside the platform.

I've added this to my upcoming draft.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
++---------------+-------------------------------------------+
+| Type          | Architecture                              |
++---------------+-------------------------------------------+
+| Created       | 2016-12-06                                |
++---------------+-------------------------------------------+
+
+Abstract
+========
+
+There is a long standing desire to allow developers to extend edx-platform by
```
andy-armstrong: Great point @dsjen. My upcoming draft broadens this.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+
+Currently the discussion board is a Django view built into edx-platform, but it
+should instead be made into a plugin. One option would be to create a new
+CourseDiscussionXBlock to display the "Course" tab, but as described above there
+are downsides to forcing this to be an XBlock. It would be cleaner if the edX
+discussion library provided a pluggable user interface that can be dropped into
+each of the contexts in which it is needed. The component would be provided
+with the context for authorization and to customize itself appropriately.
+
+An interesting question is how can such components could share course-scoped
```
andy-armstrong: @dsjen great catch! Fixed.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+1. **It should render its content into a section of a page**
+
+   A pluggable user interface should be able to return the HTML, CSS and custom
+   JavaScript necessary to render its content. In addition, it should be able to
+   declare the JavaScript libraries that it depends upon.
+
+   Python and JavaScript APIs will be provided so that the consumer of a
+   component can render it. These APIs will ensure that the necessary libraries
+   have been loaded, and that the necessary CSS and JavaScript files have been
+   added to the page (without duplication).
```
andy-armstrong: @dsjen what did you have in mind? I have a section on dependency management in the future directions section.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,475 @@
...
+Will Django Component Views ever need to have their own storage?
+----------------------------------------------------------------
+
+The intention is that Django Component Views are purely for rendering user
+interfaces, and so should not have any storage of their own. They will need to
+be configured with data (e.g. the course that is being rendered) and they need
+to be able to fetch data (e.g. the children of an xblock that they are
+rendering). If data is needed to support a feature, then that should be
+implemented as a Django model. Django views do not provide storage mechanisms,
+and hence Django component views, as a subclass, shouldn't either.a
```
andy-armstrong: Fixed

oeps/oep-0012.rst
andy-armstrong: @bjacobel great point. I've removed this sentence since it is inaccurate, and I've replaced it with the following:

> More exploration is
needed to determine whether web components are ready for use within Open edX,
and if so how they overlap with Django Component Views.

oeps/oep-0012.rst
andy-armstrong: @bradenmacdonald We've had some very interesting discussions about this. The challenge with using a regular Django model is that the settings don't live with the course, so they get lost in an export/import round-trip. If that isn't an issue, then absolutely a Django model is the right choice. If they need to live in the course, then either we need something like context-scoped fields, or we need to provide an extension to Django models to declare how they can be persisted into the OLX.

andy-armstrong:
Thanks for all the great feedback, folks. I've published a third draft that I'm hoping can be approved in the next week so that we can start the implementation. Please let me know if you have anything you'd like to see addressed before then.

oeps/oep-0012.rst
dsjen: Gotcha!

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+custom UI without having to add custom blocks to each course, and if the feature
+is removed then the tab will seamlessly disappear without having to remove the
+blocks again.
+
+Use Case 2: LMS Discussions
+---------------------------
+
+Open edX Discussions is a complex feature that integrates into many parts of the
+LMS (currently as an inline XBlock, as a courseware tab, and to provide team
+discussions). Ideally, Open edx Discussions would be a separately-installable
```
ormsbee: Capitalization: edX

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+  available through REST APIs.
+
+Will Django Component Views ever need to have their own storage?
+----------------------------------------------------------------
+
+The intention is that Django Component Views provide new user interfaces, and so
+should not have any storage of their own. They will need to be provided with
+context (e.g. the course that is being rendered) and they will need to be able
+to fetch data (e.g. the children of an xblock that they are rendering). If a new
+feature needs storage of i  ts own, then that should be implemented as a Django
```
ormsbee: Typo: its

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+  stumbling block for adding new XBlocks as the data is usually only made
+  available through REST APIs.
+
+Will Django Component Views ever need to have their own storage?
+----------------------------------------------------------------
+
+The intention is that Django Component Views provide new user interfaces, and so
+should not have any storage of their own. They will need to be provided with
+context (e.g. the course that is being rendered) and they will need to be able
+to fetch data (e.g. the children of an xblock that they are rendering). If a new
```
ormsbee: typo: XBlock

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+----------------------------------------------
+
+Django Component Views should not be constructed as a tree of views as that is
+introducing state that needs to be managed. However, component views can invoke
+other component views to render portions of their content. For example, a Unit
+Student View could ask each of its children to render itself as a fragment, and
+then could return a composite fragment containing all of them. This is
+essentially how XBlocks render their children today, and component views should
+work in the same way.
+
```
ormsbee: I'd like to avoid arbitrary nesting, or at least discourage it until we've proven that we can have it working well in an async manner. One of the biggest performance issue with XBlocks has always been the blocking behavior of having to invoke arbitrary children and grandchildren, and not being able to return until they've all executed.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+course tab to be able to register a pluggable user interface that will be
+presented as a tab, and for the implementor not to need to know how to implement
+the full page.
+
+There has been a long standing request to support rendering XBlocks as course
+tabs (`Allow xblocks to be shown as course tabs`_), but there are conceptual
+issues with this approach:
+
+* XBlocks must be configured and saved into the modulestore which doesn't make
+  much sense for course tabs that don't have their own state
```
pmitros: Many XBlocks don't have their own state.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* Replacing a built-in tab such as "Course" would require a data migration for
+  every course if it was saved as an XBlock
```
pmitros: For the course tab, I believe the schema was designed with the plan this would not require any migration. For the other tabs, it would, but right now, these are hackish schemas regardless. Any cleanup would require a migration.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* It seems confusing to make an XBlock that should never be added into the
+  courseware, but which only exists to be shown as a tab
```
pmitros: The tabs should be part of the OLX. That was an artifact of how the platform was built. The component model didn't come in until some were already built.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+It would be prefable to allow a feature library to register a course tab plugin
+which in turn renders using a component. This allows the feature to provide a
+custom UI without having to add custom blocks to each course, and if the feature
+is removed then the tab will seamlessly disappear without having to remove the
+blocks again.
```
pmitros: A big part of the goal of XBlock decorators was exactly to allow this kind of automatic addition/removal of features. There are similar mechanism which could manage this with XBlocks.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+Open edX operator could choose whether to install edX discussions or an
+alternate collaboration tool.
+
+Currently the discussion board is a Django view built into edx-platform, but it
+should instead be made into a plugin. One option would be to create a new
+CourseDiscussionXBlock to display the "Course" tab, but as described above there
+are downsides to forcing this to be an XBlock. It would be cleaner if the edX
+discussion library provided a pluggable user interface that can be dropped into
+each of the contexts in which it is needed. The component would be provided
+with the context for authorization and to customize itself appropriately.
```
pmitros: That's exactly what an XBlock is intended to be -- a pluggable user interface which can be plugged in where ever needed.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+-------------------------------
+
+EdX Insights is a course analytics application that could be extended in
+a variety of different ways. Some examples include:
+
+1. ** New Charts **
+
+   A plugin could provide new types of charts for existing data types, or it
+   could extend the application to show charts for data that had not previously
+   been visualized.
```
pmitros: This was actually a use case in the design of XBlocks. The idea was that Insights would act as a data service, and XBlocks would grab data and render it. We could then define a front end as a set of XBlocks. Adding component would force us to reinvent solutions to many problems -- for example, we'd need some mechanism for determining which views to render in the dashboard. We would also want some way to compose dashboards from such renderings (we may want more than one per page). OLX lets us do that, as do nested XBlocks. Here, it's the same problem, just pushed into the future.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+to define a component that can be rendered into an existing page, or that can
+be fetched over AJAX and dynamically inserted into a page. This proposal is
+to provide this capability through a new concept called Django component views.
+
+A new `DjangoComponentView` class will be implemented as a subclass of
+`DjangoView` that are then declared in `urls.py` as with other views. They do
+not themselves define user-visible pages but instead provide an implementation
+that can render a component into another page. These views have full access to
+Django so they can use standard mechanisms to access any data that they need to
+be able to render themselves.
```
pmitros: It seems like a lot of the benefit of this could be had by simply allowing edX-specific XBlocks to make use of Django (or even just dropping the cross-platform aspiration of XBlocks). Many already do this, but it's considered bad practice. We could re-consider it good practice.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+Rationale
+=========
+
+Why are we not using an industry framework?
+-------------------------------------------
+
+It seems that the requirements for pluggable user interfaces are not unique
+to Open edX, and hence other solutions are likely to exist. Our research so
+far has not found a solution that covers the full set of use cases, but that
+might change in the future.
```
pmitros: I've actually been pretty impressed with React components for solving much of edX' use case. We'd need to add a little bit to them, but it might be easier than starting from scratch.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+from tens or even hundreds of iframes.
+
+Could we use XBlocks instead of introducing Django Component Views?
+-------------------------------------------------------------------
+
+Note that all of the requirements of pluggable user interfaces are supported by
+XBlocks. However, there are several reasons why XBlocks are not always
+appropriate:
+
+* XBlocks need to have a unique usage id so that instances can be referenced.
```
pmitros: So do the plug-in components, actually (see above discussion about unique IDs) :)

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* XBlocks need to have storage scoped for different purposes (per learner, per
+  course etc).
```
pmitros: They actually don't -- a few access the database directly. It's just a convenient storage mechanism. Per-user storage seems broadly applicable to many use cases described (e.g. keeping forum settings and similar). Where it doesn't make sense, one doesn't need to use it. 

Per-course was always a mistake,

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* XBlocks in the LMS are tied to a course, which doesn't make them great for
+  general user interfaces (e.g. UIs that are global, organization-wide or
+  program-wide).
```
pmitros: It was a goal of XBlocks to be only thinly coupled to courses or decoupled. That's a cleanup still worth doing.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* XBlocks are persisted while component views are ephemeral. This also means
+  that converting an existing view into an XBlock requires a data migration
+  so that each course is updated to contain a block instance.
```
pmitros: I don't quite understand this comment. Both are rendered on-the-fly. Both sometimes make use of permanent storage. I'm not quite sure I catch the distinction.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* XBlocks need to support serialization to OLX while component views don't.
```
pmitros: Are you sure about this? Won't we want some way to capture the configuration of an Open edX platform, or a course in an Open edX platform, if for no reason, at least for course dumps, so researchers can make sense of how we're set up?

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+
+* XBlocks are designed to be platform-agnostic and so in a pure implementation
+  should only fetch data through provided runtime services. This has been a
+  stumbling block for adding new XBlocks as the data is usually only made
+  available through REST APIs.
```
pmitros: Eddie has proposed eliminating this from XBlocks.

oeps/oep-0012.rst
```diff
@@ -0,0 +1,520 @@
...
+There seem to be many benefits to XBlocks rendering themselves using component
+views, but the challenge is that today XBlocks explicitly do not make use of
+Django. A future rethinking of XBlocks based upon Django and utilizing component
+views would allow the following:
+
+* A runtime could allow the configuration of alternate UIs for certain blocks
+  that could be installed as plugins. For example, an Open edX operator could
+  choose to show a different UI for its "Course" tab. As another example,
+  a mobile runtime could show native mobile versions of certain blocks,
+  perhaps using `React Native`_.
```
pmitros: I think, in many cases, this ought to be a course-specific decision. Operator might have a default preference, but the right rendering is designed on how the course content ought to be consumed. Some courses are very linear, and some highly non-linear.

pmitros:
First of all, my comments should not be blocking here. They are my thoughts; I am not an arbiter on this, so do not give them too much weight. The decision is with @efagin, @nasthagiri, and the other architects. 

However, on a high level, reading the proposal, I'm pretty confused as to why we want two independent plug-in mechanisms. The proposal identifies several shortcomings of XBlocks, such as the lack of ability to call into Django, to be extended up the hierarchy to course tab, or to sit outside of a course context. All of those could be addressed within the XBlocks framework. It claims too much complexity in some aspects of XBlocks. For example, the XBlocks fields storage model is not need in some contexts. But XBlocks aren't required to store things; just don't use that where it doesn't make sense. 

The proposal also has a set of philosophical difference with XBlocks -- for example, whether the framework ought to be Django-specific or Django-agnostic. If we are not planning to invest enough in XBlocks to make them usable in other platforms, we could change our mind about this, and tie them to Django. There is nothing tying our hands to older technical decisions. The same could be said of almost every other decision around XBlocks. 

However, two APIs doing almost exactly the same thing isn't better than one. It leads to twice as much maintenance cost, twice as much developer on-boarding, and exponentially more test complexity. If XBlocks have shortcomings, those should be addressed. If they're not possible to address, we should consider making a new 2017-era framework. If they couple too much stuff together (e.g. views, serialization, etc.), that could be broken out into independent libraries. But simply having two APIs, especially as similar as these, doesn't make much sense to me. We'll end up having to reimplement AB tests, ways to get configuration information to course teams/researchers/etc., ways to make this work with mobile, and many other things, in two different ways. That only leads to inconsistency and confusion for everyone -- developers, course teams, and users. 

Again, just my 2c.

oeps/oep-0012.rst
bradenmacdonald: @andy-armstrong Ah, ok, that makes sense. FYI, I'm kind of waiting on some of the architectural questions around UI Blocks and discussion XBlock to settle down before I continue prototyping course-scoped fields. If you think that there may be a need to tie them in to UI blocks in some way, or make UI Blocks be able to read them, let me know.

cpennington:
> If we are not planning to invest enough in XBlocks to make them usable in other platforms, we could change our mind about this, and tie them to Django. There is nothing tying our hands to older technical decisions. The same could be said of almost every other decision around XBlocks.

100% agree, and that was actually part of the discussion that @andy-armstrong, @nasthagiri and I had at the start of this. In some ways, this was designed as a way to proof out how this style of plugin would work to then pave a path towards migrating XBlock (and things like state and runtime services) over to a Django-centric framework.

andy-armstrong:
@pmitros, @nasthagiri and I had a great discussion yesterday to go over his feedback. We came up with a variant of this proposal, which is to move fragments out of XBlock into its own repo (we're thinking called web-fragments). This repo will publish PyPI and NPM packages. The Python package will provide the fragments itself, and optionally include a DjangoComponentView (DjangoFragmentView?) which provides the same capabilities I described here. The JavaScript package will initially provide a JQuery implementation for handling fragments in the browser, which is code we already have in three places (Studio, conditional blocks, and https://github.com/edx-solutions/jquery-xblock). In the future we'll add a React version too.

I'm going to publish another draft that updates to this new recommendation. In the meantime, let me know if you have issues with this, and in particular let me know what you think of the name web-fragments. It has one benefit of being shorter than django-component-views!

pmitros:
@cpennington I wrote the text below, and Andy posted much of the same thing as I was finishing up. Feel free to read or skip, depending on how closely you want to follow this. Andy, Nimishia, and I sat down. We had a very constructive discussion on the best way to build this abstraction, and we reached a consensus. 

The DCV framework isn't much more than a wrapper for something which resembles the `fragment` in XBlock. We would create a new fragment library which would decouple this from XBlock. This would be a general-purpose multi-language multi-framework library in its own github repo, with integration for XBlock, Django, and JavaScript (and there, potentially Backbone, React, etc.). It would be used by both XBlock and edx-platform. In Django projects, there would be a thin decorator or similar for rendering fragments.

This would:

- Provide substantially the same functionality as DCV. The naming of DCV was a little bit misleading -- unlike things like XBlocks, React, JQuery, WebComponents, etc., it wasn't intended to be a full component system. It was intended to essentially bring `fragment` functionality to Django.
- Bring Django, edx-platform, and xblock into greater convergence, rather than divergence. Both would use a common library, so it would be easier to learn, document, maintain, and move code back-and-forth.
- Resolve existing pain points in the platform -- fragments are rendered client-side in three different places.
- Avoids the discussion about whether e.g. we should invest resources to make tabs into XBlocks. If they were to become XBlocks, moving them to a common rendering framework would be a natural step in that direction. If they were not, this would still result in cleaner code. That's a mark of good architecture -- changes like that are easy to make. 

Not as part of the discussion, but if this works well, we could apply the same approach to other aspects of XBlocks as well. For example, XBlocks uses OLX for configuration and serialization of configuration, with things like overrides, inheritance, and so on. This approach has many good ideas, and many inconsistent hacks. In the future, we could create a new library which would abstract that out, clean it up, remove the hacks, and apply it platform-wide. We could repeat for XBlock services, for component discovery/entry points, for front-end, and so on.

Andy: Regarding naming, web fragments sounds fine to me. If that's chosen, DjangoFragmentView sound good to me. Word like "component," "module," "data," "abstraction," and similar are almost always unclear; all those words have too many meanings in different contexts. With DFV, you know what it is from the name.

antoviaque:
@cpennington 

>  In some ways, this was designed as a way to proof out how this style of plugin would work to then pave a path towards migrating XBlock (and things like state and runtime services) over to a Django-centric framework.

Would that change break compatibility? Ie does that mean that existing XBlocks would need to be adapted / partially rewritten?

cpennington:
> Would that change break compatibility? Ie does that mean that existing XBlocks would need to be adapted / partially rewritten?

Hard to say. We're still quite a ways away from that point. However, speaking from experience, maintaining backwards compatibility layers for long periods can definitely be a drain.

andy-armstrong:
Thanks for the great feedback and discussion, @pmitros. I've pushed up a fourth draft that describes our agreed upon design for a new web-fragments library.

Folks who are interested, please re-review.

pmitros:
@antoviaque I'll just echo the: "We're still quite a ways away from that point." Last I looked -- a year or two ago -- there were about 50 XBlocks in the community, a fair subset of which were on edx.org. On the other hand, edx-platform still has XModules. If history is any indicator, the odds of us even having the capacity for a serious backwards-compatibility change to XBlocks and migration of even what's on edx.org in say, the next half-decade, are not all that high.

One of the nice things about this proposal as revised is that it pulls out one of the key concepts which has worked well in XBlocks, and makes it usable platform-wide. It can be a path to either re-imagining XBlocks, or to convergence, and finishing XBlocks and making them more core to the platform.

andy-armstrong:
Thanks @clintonb. 

Folks on this PR, I plan on merging this by end of day tomorrow unless I hear any objections.

antoviaque:
@andy-armstrong @pmitros 

> We're still quite a ways away from that point. However, speaking from experience, maintaining backwards compatibility layers for long periods can definitely be a drain.

> edx-platform still has XModules. If history is any indicator, the odds of us even having the capacity for a serious backwards-compatibility change to XBlocks and migration of even what's on edx.org in say, the next half-decade, are not all that high.

Yes, that's exactly why I'm bringing this up now - that would be a second fundamental architecture change to the plugin system, before XBlocks have even been properly finished.

> Last I looked -- a year or two ago -- there were about 50 XBlocks in the community, a fair subset of which were on edx.org.

That might not look like a lot, but for a young community of our size, that's is actually a lot. The community has already heavily invested in XBlocks (including edX), and people now know how to write them. My gut feeling is that whichever solution we end up with, it shouldn't result in asking the community to rewrite their XBlocks or learn Django. XBlocks work pretty well as is for most uses, people wouldn't understand imho. It might for example be better for the project to have better / more standard ways to handle the border cases - but optionally, to not affect those who don't need the additional complexity? And especially without asking people to rewrite their plugins...

andy-armstrong:
@antoviaque thanks for the great feedback.

> My gut feeling is that whichever solution we end up with, it shouldn't result in asking the community to rewrite their XBlocks or learn Django.

I agree that we shouldn't ask folks to rewrite their XBlocks. In our discussions it seemed to make the most sense to keep XBlocks as they are, but to additionally provide a new Django XBlocks that reuses code where it makes sense, but which benefits from all the functionality of Django too. Eventually we might deprecate the old one, but certainly not until the Django version has really proved itself. 

I disagree about not forcing the community to learn Django. If we decide that Django is the best framework for writing plugins, then it is better for the community to learn it than to have to learn a proprietary framework. I would imagine that most Open edX developers will have to learn Django anyway. There's also the possibility that Python is not the best way to create simple XBlocks and that we should provide a JavaScript-based solution. Anyway, we don't need to decide this here as it will be a great topic for a future OEP.

antoviaque:
@andy-armstrong Thanks for the answers - and we can definitely discuss this in future iterations yep.

PR #35, opened 2016-12-14
update arbiter field of OEP-6

efagin:
fyi @cpennington @bradenmacdonald - no review required, I'm just fixing the header to include the arbiter. you may also want to add something under resolution at some point.


bradenmacdonald:
Thanks @efagin.

PR #37, opened 2017-01-09
OEP-14: Proposed process for archiving edX Github repositories

cahrens:
This is a continuation of discussions that began in https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process.

Please review @e0d @efagin @pmitros @wesmason @robrap 

Also let me know if you would be willing to serve as Arbiter of this OEP.

FYI @staubina @jlajoie @gsong @andy-armstrong @jibsheet

### DIFF
--- a/oeps/oep-0014.rst
+++ b/oeps/oep-0014.rst
@@ -1,0 +1,197 @@
+=========================================
+OEP-14: Archiving edX Github Repositories
+=========================================
+
++---------------+-------------------------------------------+
+| OEP           | :doc:`OEP-14 </oeps/oep-0014.rst>`        |
++---------------+-------------------------------------------+
+| Title         | Archiving edX Github Repositories         |
++---------------+-------------------------------------------+
+| Last Modified | 2017-01-18                                |
++---------------+-------------------------------------------+
+| Author        | Christina Roberts, christina@edx.org      |
++---------------+-------------------------------------------+
+| Arbiter       | Eddie Fagin, eddie@edx.org                |
++---------------+-------------------------------------------+
+| Status        | Accepted                                  |
++---------------+-------------------------------------------+
+| Type          | Process                                   |
++---------------+-------------------------------------------+
+| Created       | 2017-01-09                                |
++---------------+-------------------------------------------+
+| `References`  | `ORA PR Discussion`_,                     |
+|               | `Initial Archiving Discussions`_          |
++---------------+-------------------------------------------+
+
+.. _ORA PR Discussion: https://github.com/edx/edx-ora/pull/187
+.. _Initial Archiving Discussions: https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
+
+Abstract
+========
+
+The `edx organization`_ contains a large number of repositories, most of
+which are active and maintained, but some of which are now obsolete. To
+clarify the status of repositories, a process for archiving a
+repository is defined below.
+
+.. _edx organization: https://github.com/edx
+
+Motivation
+==========
+
+Recently openedx.yaml files were added to edX repositories per `OEP-2`_. In the
+course of deciding owners for those repositories, there was an `ORA PR
+Discussion`_ about how best to handle deprecated or obsolete repositories; in
+particular, do obsolete repositories need owners, and how can repositories be
+clearly marked as present for archive purposes only?
+
+This discussion resurfaced related to edX's usage of Gemnasium to report
+the usage of third-party libraries that have known security issues. All
+repositories under the edX organization were being monitored, but this added
+noise when trying to understand the number of third-party library updates
+required for actively maintained repositories.
+
+.. _OEP-2: https://open-edx-proposals.readthedocs.io/en/latest/oep-0002.html
+
+Specification
+=============
+
+When a repository under the `edx organization`_ will no longer be maintained
+because it is no longer in use, the following steps should be followed.
+
+Transfer to New Owner if Interest
+---------------------------------
+First, if the repository is public, follow these steps to see if anyone would like
+to take ownership of it:
+
+1. Send an email to edx-code@googlegroups.com announcing that the repository
+   will be archived, and inquiring if anyone would like to take ownership of
+   the repo. If there are no responses after 2 work days, skip to `Archive
+   Steps`_.
+
+2. If someone does wish to take ownership of the repository, email the internal
+   edX developers mailing list to see if there are any objections. If there are
+   no objections after 2 work days, post to edx-code@googlegroups.com that the
+   transfer will take place.
+
+3. Create an IT help ticket for the repository to be transferred to the new
+   organization.
+
+4. Once the transfer has occurred, create a fork of the transferred repository
+   into the `edx organization`_ and follow the `Archive Steps`_ below for the
+   forked repo.
+
+Archive Steps
+-------------
+1. Update the README.rst file in the repository to state that it is archived,
+   using the `README Archive Statement`_ below.
+
+2. Add ``archived: True`` to openedx.yaml (creating the file if necessary). Note
+   that it is not necessary for the openedx.yaml file to define an owner when
+   archived is set to True.
+
+3. Create an IT help ticket to update the description of the repository to
+   begin with ``[ARCHIVED]``.
+
+4. Create a DevOps ticket to change the status of the repo to "non-monitored"
+   in Gemnasium, and to remove any integration or webhooks that DevOps may be
+   maintaining.
+
+In the future, the step of changing the monitored status of a repository in
+Gemnasium could be automated using Gemansium APIs and keying off of the
+``archived`` value in openedx.yaml.
+
+README Archive Statement
+------------------------
+Include this statement in the README.rst file:
+
+    This repository has been archived and is no longer supported—use it at your
+    own risk. This repository may depend on out-of-date libraries with security
+    issues, and security updates will not be provided. Pull requests against
+    this repository will also not be merged.
+
+Rationale
+=========
+
+The proposed process leverages the already-existing ``archived`` flag in
+openedx.yaml. It does not require introducing a new organization that is
+maintained by edX, and the source code remains easily visible and searchable
+(see `Rejected Alternatives`_).
+
+Backward Compatibility
+======================
+
+This proposal does not introduce any backward compatibility issues.
+
+Reference Implementation
+========================
+
+The `Discussions Hackathon repository`_ has been updated to conform to the
+`Archive Steps`_.
+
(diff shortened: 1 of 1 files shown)


e0d:
Should we accommodate:
- transfers after archiving
- unarchiving

I think we should have a document somewhere, maybe here, that enumerates the caveat emptor items related to using an archived repo, e.g., at you own risk, security updates will not be provided, PRs will not be considered, etc.

efagin:
Well @andy-armstrong here's another argument in favor of merging empty OEPs to claim IDs. You have a collision with @cahrens here for OEP-11: https://github.com/edx/open-edx-proposals/pull/32.

@cahrens I believe OEP-14 is currently unclaimed by any active PR.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+When a repository under the `edx organization`_ will no longer be maintained
+because it is no longer in use, the following steps should be followed:
+
+First, send an email to edx-code@googlegroups.com announcing that the
+repository will be archived, and inquiring if anyone would like to take
+ownership of the repo. If someone wishes to take ownership, transfer the
+repository to their github organization and hand over rights [Open questions:
+Do we need to outline these steps further? Would there need to be internal
+sign-off before transferring the repo outside of edX? Would we wish to create
+our own fork of the repo?].
```
pmitros: I would propose:

* Sign-off: We let the internal developers mailing list know about this, ask if there are objections, wait 48 hours, and then post to edx-code if there are none. 
* We definitely do make a fork. These repos were deployed on edx.org, and we want to be able to bring up historical versions of the code. Transfer-and-fork seems like a reasonable process. That way, github properly indicates where the main repo is, but we have a copy.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+If nobody wishes to take ownership of the repository then:
+
+1. Update the README.rst file in the repository to state that it is archived
+and that the repository may depend on out-of-date libraries with security
+issues.
+
+2. Add "obsolete: True" to openedx.yaml (creating the file if necessary). Note
+that currently an owner is required in openedx.yaml per `OEP-2 <oep-0002.rst>`_,
+but a proposal will be made to relax this requirement for repositories which
+are marked as obsolete/archived.
```
pmitros: I prefer the name "archive" to "obsolete," as per Facebook's standards. We will want to retire things from edX maintenance which aren't necessarily obsolete; just ones we no longer have a use for.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+repository is defined below.
+
+.. _edx organization: https://github.com/edx
+
+Motivation
+==========
+
+Recently openedx.yaml files were added to edX repositories per
+`OEP-2 <oep-0002.rst>`_. In the course of deciding owners for those repositories,
+there was a `ORA PR Discussion`_ about how best to handle deprecated or obsolete
```
robrap: nit: a => an

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+.. _Initial Archiving Discussions: https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
+
+Abstract
+========
+
+The `edx organization`_ contains a large number of repositories, most of
+which are active and maintained, but some of which are now obsolete. To
+clarify the status of repositories, and to ensure that we are only monitoring
+active repositories via Gemnasium (which alerts about the usage of third
+party libraries with known security issues), a process for archiving a
```
jibsheet: Instead of only calling out Gemnasium - can we just refer to "GitHub Organization-wide tools" and list Gemnasium as an example?  We use VersionEye on other repos, and may switch.  I'm sure there are other tools that in the future we'd let consume all of the repos in edx due to testing/compliance/etc.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+Motivation
+==========
+
+Recently openedx.yaml files were added to edX repositories per
+`OEP-2 <oep-0002.rst>`_. In the course of deciding owners for those repositories,
+there was a `ORA PR Discussion`_ about how best to handle deprecated or obsolete
+repositories; in particular, do obsolete repositories need owners, and how can
+repositories be clearly marked as present for archive purposes only?
+
+This discussion resurfaced related to edX's usage of Gemnasium to report
```
jibsheet: Seems like this mention of Gemnasium is sufficient and we could remove it from the Abstract.

oeps/oep-0014.rst
jibsheet: Agreed, we keep a fork

oeps/oep-0014.rst
jibsheet: +1 for archived

oeps/oep-0014.rst
```diff
@@ -0,0 +1,175 @@
...
+
+First, send an email to edx-code@googlegroups.com announcing that the
+repository will be archived, and inquiring if anyone would like to take
+ownership of the repo. If someone wishes to take ownership, transfer the
+repository to their github organization and hand over rights [Open questions:
+Do we need to outline these steps further? Would there need to be internal
+sign-off before transferring the repo outside of edX? Would we wish to create
+our own fork of the repo?].
+
+If nobody wishes to take ownership of the repository then:
```
mulby: I think using a new organization "edx-archive" that we transferred the repositories to would clarify things even further. That way we could use organization-wide tools without having to manually manage exclusion lists in each one (VersionEye, Gemnasium etc).

pmitros: Both approaches were considered, actually, and had upsides and downsides. See:
  https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
I don't think there was strong consensus in either direction. The tool management issue is not bad if you include an appropriate tag in the description (speaking from experience; I manage my own repos that way).

mulby: Ah, I hadn't seen that page. Helpful context. I don't feel very strongly about this, but I think I like Option 1 the best.

The current proposal feels like it optimizes for the less frequent use case (searching and seeing history) than for the most frequent use case (searching and seeing code under active development). I also don't see a significant difference in the amount of maintenance work and overhead.

That is I would rather have to opt-in to seeing history than to opt-out of seeing it. Particularly given the fact that as far as I can tell, when you perform a code search on github (as opposed to a repository search) it doesn't appear to show the description of the repo in the search result, see:

screenshot from 2017-01-10 07-41-08

This is in the results for the search "ora user:edx". In order to also opt-in to searching the archive, you could simply search for "ora user:edx user:edx-archive". In this example, it isn't abundantly clear what the difference is between "edx-ora" and "edx-ora2".

It also looks like the results are ordered by "best match" not by repo activity, so this could easily return deprecated repos above active ones.

cahrens: @mulby see further down in "Rejected Alternatives" for a discussion of transferring to a different organization. I don't feel strongly either way, but there seemed to be some resistance to maintaining another organization (in the discussion on the confluence page).

The main problem I'm having is that I don't know how we actually come to a decision, as feelings are pretty split among the 3 different proposals.

oeps/oep-0014.rst
cahrens: If we have to keep a fork, then transferring the repo to someone else just adds more steps to the process. I know @pmitros feels that it is a better contribution to the open source community-- do others think that offering to transfer to someone else is worthwhile?

jibsheet: It's what we did in a previous role (offer to the community when we weren't going to develop/support it further) so it makes sense to me.  I agree it does make more work.

oeps/oep-0014.rst
pmitros: @cahrens Regarding:

> The main problem I'm having is that I don't know how we actually come to a decision, as feelings are pretty split among the 3 different proposals.

Short story: My understanding is that at this point, you decide. 

Long story: Broadly speaking, the goal of the process is to give a chance to solicit feedback from as many stakeholders as possible. Having and understanding broad input from diverse viewpoints leads to better design decisions. That doesn't mean there needs to be consensus -- just an understanding of the viewpoints and trade-offs by the author of the PR. Ideally, everyone has a chance to kibitz, but ultimately, the person driving the thing gets to decide. The only people with blocking power are two reviewers (and, at this point, you have three :+1:'s), your lead architect, and the chief architect. 

100% consensus is rarely good design. Some people will not be fully familiar with the problem. In some cases, compromises between people with conflicting styles and design practices lead to worse overall solutions than one consistent with one. Requiring 100% consensus also tends to lead to people soliciting less feedback, which almost always leads to worse design.

cahrens:
FYI, I have updated the proposal based on feedback (with explicit steps for transferring to a new owner if there is interest, and a template for text to put in the README file).

Also, @efagin has volunteered to be the Arbiter.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,196 @@
...
+4. Once the transfer has occurred, create a fork of the transferred
+repository into the `edx organization`_ and follow the `Archive Steps`_ below
+for the forked repo.
+
+Archive Steps
+-------------
+1. Update the README.rst file in the repository to state that it is archived,
+using the `README Archive Statement`_ below.
+
+2. Add "obsolete: True" to openedx.yaml (creating the file if necessary). Note
```
cahrens: I have created #40 to change "obsolete" to "archived" and to remove the requirement for an owner in that case.

oeps/oep-0014.rst
```diff
@@ -0,0 +1,188 @@
...
+4. Once the transfer has occurred, create a fork of the transferred
+repository into the `edx organization`_ and follow the `Archive Steps`_ below
+for the forked repo.
+
+Archive Steps
+-------------
+1. Update the README.rst file in the repository to state that it is archived,
+using the `README Archive Statement`_ below.
+
+2. Add "archived: True" to openedx.yaml (creating the file if necessary). Note that
```
cahrens: I went ahead and updated this to reflect the changes that will be merged with #40.

cahrens:
FYI @e0d @efagin @pmitros @wesmason @robrap @staubina @jlajoie @gsong @andy-armstrong @jibsheet @mulby and other interested parties-- I have fully followed the proposed Archive Steps for the hackathon discussion repo and noted it in this proposal as a reference implementation.

Please let me know by EOD tomorrow (1/12) if you have any objections with this proposal being accepted.

gsong:
@cahrens I’ve provided small feedback via #42.

efagin:
@cahrens as arbiter, I approve updating this to Accepted and merging it - go ahead whenever you're ready. Thanks for pulling this one together!

PR #38, opened 2017-01-09
Add a "Not Applicable" state to OEP-2

mulby:
When updating [openedx.yaml](https://github.com/edx/edx-analytics-pipeline/blob/5b3a6db1e466a18d67cc74c0a4c98e2c3906290d/openedx.yaml) for edx-analytics-pipeline as part of the OEP-5 adoption work. I discovered that many OEPs don't really apply to the data pipeline. However, there was no way for me to mark those OEPs as "not applicable" so any automated reports would just assume that this repo was not compliant.

A reasonable argument could also be made for just setting the `state` to `True` in those cases, but I think I prefer the explicit third state.

FYI @cpennington

### DIFF
--- a/oeps/oep-0002.rst
+++ b/oeps/oep-0002.rst
@@ -64,9 +64,15 @@ following fields:
     ``oep-n``). Each value would be one of ``True``, ``False``, or a
     reason dictionary. ``True`` and ``False`` would indicate accordance (or
     lack thereof) with the specified OEP. A reason dictionary
-    must have ``state`` and ``reason`` keys. The ``state`` key is a boolean,
-    and ``reason`` is a string that should explain why it's ok that the
-    ``state`` is ``False`` (this will be displayed in reporting tools).
+    must have a ``state`` key. The ``state`` key can have a value of ``True``,
+    ``False`` or ``Not Applicable``. If the ``state`` is ``False`` the reason
+    dictionary must also include a ``reason`` key. The ``reason`` is a string
+    that should explain why it's ok that the ``state`` is ``False``
+    (this will be displayed in reporting tools). A ``state`` value of
+    ``Not Applicable`` indicates the OEP will never be applicable to this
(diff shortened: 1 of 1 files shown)


efagin:
In this case, I'd prefer @cpennington review as the original OEP-2 author. I'll remove myself as a formal PR reviewer.

oeps/oep-0002.rst
```diff
@@ -64,9 +64,15 @@ following fields:
     ``oep-n``). Each value would be one of ``True``, ``False``, or a
     reason dictionary. ``True`` and ``False`` would indicate accordance (or
     lack thereof) with the specified OEP. A reason dictionary
-    must have ``state`` and ``reason`` keys. The ``state`` key is a boolean,
-    and ``reason`` is a string that should explain why it's ok that the
-    ``state`` is ``False`` (this will be displayed in reporting tools).
+    must have a ``state`` key. The ``state`` key can have a value of ``True``,
+    ``False`` or ``Not Applicable``. If the ``state`` is ``False`` the reason
```
cpennington: It might be worth clarifying whether these are strings or booleans, and providing some sample snippets of yaml that are correctly formatted.

mulby:
@cpennington I've revised it to make it more clear how this would be parsed. It increases the complexity of the logic in the parser (in terms of enforcing required fields etc), however, it's backwards compatible and takes advantage of the native types in YAML.

oeps/oep-0002.rst
```diff
@@ -129,6 +143,11 @@ The design of the ``oeps`` dictionary was guided by a couple of use cases:
 Change History
 ==============
 
+2016-01-10
```
cahrens: @mulby It's 2017 now. :) I can update this in #40 if you wish.

mulby: Hahaha oops I knew I would do this at least once... I'll make a PR

PR #39, opened 2017-01-09
Fix typo in OEP-1

mulby:
This level of formality seems silly for this type of fix... but I figured it was better than just committing directly to master.


efagin:
You don't need @cpennington's approval here, go ahead and merge. 👍  from me.

efagin:
You know what? I'll just do it now.

PR #43, opened 2017-01-24
OEP-15: Course-wide Custom JavaScript

msegado:
Submitting this as an OEP per [the discussion on edx-code](https://groups.google.com/forum/#!topic/edx-code/T83TDxhH74E/discussion). Here's a copy of the abstract as of January 24, 2017 (see [the current OEP draft](https://github.com/msegado/open-edx-proposals/blob/oep15/oeps/oep-0015.rst) for any changes):

> ## Abstract
> The Open edX platform is a powerful tool for both content authoring and learning-sciences research, in part due to its support for including arbitrary JavaScript in course materials; for example, such scripts can enable researchers to conduct novel experiments and log custom learner interaction events for analysis. Unfortunately, deploying and maintaining JavaScript assets across an entire course (or multiple courses) is a significant challenge, especially in Studio-based workflows where `<script type="text/javascript">` elements must be included manually in every vertical. As an alternative, this OEP proposes a mechanism for authors to include custom JavaScript across an entire course using a policy key. The implementation is fully backward-compatible and leverages existing XBlock infrastructure to include scripts in the LMS.

/cc @dseaton @jolyonb

### DIFF
--- a/oeps/oep-0015.rst
+++ b/oeps/oep-0015.rst
@@ -37,7 +37,7 @@ The Open edX platform is a powerful tool for both content authoring and learning
 Motivation
 ==========
 
-The primary goal of this OEP is to facilitate research and experimentation by removing the bariers to large-scale JavaScript inclusion. Simplifying the inclusion of scripts is especially important for experiments spanning many courses (e.g., all MITx courses): it is not reasonable to expect all course teams at an institution to manually modify every single vertical in their course to include a research script as is the case with the current platform, but it is quite feasible to request that they make a single easy-to-reverse change to their course policy.
+The primary goal of this OEP is to facilitate research and experimentation by removing the barriers to large-scale JavaScript inclusion. Simplifying the inclusion of scripts is especially important for experiments spanning many courses (e.g., all MITx courses): it is not reasonable to expect all course teams at an institution to manually modify every single vertical in their course to include a research script as is the case with the current platform, but it is quite feasible to request that they make a single easy-to-reverse change to their course policy.
 
 As an additional benefit, a course-wide JavaScript mechanism would also simplify the use of course- or institution-wide utility scripts; HarvardX, for example, maintains a custom JS library for course authors.
 
--- a/oeps/oep-0015.rst
+++ b/oeps/oep-0015.rst
@@ -1,0 +1,144 @@
+=====================================
+OEP-15: Course-wide Custom JavaScript
+=====================================
+
++---------------+----------------------------------------------------+
+| OEP           | :doc:`OEP-15 </oeps/oep-0015>`                     |
++---------------+----------------------------------------------------+
+| Title         | Course-wide Custom JavaScript                      |
++---------------+----------------------------------------------------+
+| Last Modified | 2017-01-24                                         |
++---------------+----------------------------------------------------+
+| Author        | Martin Segado <msegado@mit.edu>                    |
++---------------+----------------------------------------------------+
+| Arbiter       |                                                    |
++---------------+----------------------------------------------------+
+| Status        | Draft                                              |
++---------------+----------------------------------------------------+
+| Type          | Product Enhancement                                |
++---------------+----------------------------------------------------+
+| Created       | 2016-08-29                                         |
++---------------+----------------------------------------------------+
+| Resolution    |                                                    |
++---------------+----------------------------------------------------+
+| References    | - `Initial OEP suggestion`_ on the edx-code list   |
+|               | - `An edx-code question`_ from another user at MIT |
+|               |   who was hoping to include course-wide JS/CSS     |
++---------------+----------------------------------------------------+
+
+.. _Initial OEP suggestion: https://groups.google.com/d/topic/edx-code/T83TDxhH74E/discussion
+.. _An edx-code question: https://groups.google.com/d/topic/edx-code/idjPWUIx8Ls/discussion
+
+Abstract
+========
+
+The Open edX platform is a powerful tool for both content authoring and learning-sciences research, in part due to its support for including arbitrary JavaScript in course materials; for example, such scripts can enable researchers to conduct novel experiments and log custom learner interaction events for analysis. Unfortunately, deploying and maintaining JavaScript assets across an entire course (or multiple courses) is a significant challenge, especially in Studio-based workflows where :code:`<script type="text/javascript">` elements must be included manually in every vertical. As an alternative, this OEP proposes a mechanism for authors to include custom JavaScript across an entire course using a policy key. The implementation is fully backward-compatible and leverages existing XBlock infrastructure to include scripts in the LMS.
+
+Motivation
+==========
+
+The primary goal of this OEP is to facilitate research and experimentation by removing the bariers to large-scale JavaScript inclusion. Simplifying the inclusion of scripts is especially important for experiments spanning many courses (e.g., all MITx courses): it is not reasonable to expect all course teams at an institution to manually modify every single vertical in their course to include a research script as is the case with the current platform, but it is quite feasible to request that they make a single easy-to-reverse change to their course policy.
+
+As an additional benefit, a course-wide JavaScript mechanism would also simplify the use of course- or institution-wide utility scripts; HarvardX, for example, maintains a custom JS library for course authors.
+
+Specification
+=============
+
+The proposed implementation adds a new policy key for custom course-wide scripts (name TBD):
+
+.. code-block:: json
+
+   {
+     "global_scripts": [
+       "/static/experiment.js",
+       "//some.cdn.com/library.js"
+     ]
+   }
+
+Since the edX courseware template already receives and renders an XBlock fragment (which includes JavaScript resources), additional script resources would simply be added to this fragment `when the courseware context is created`_.
+
+.. _when the courseware context is created: https://github.com/edx/edx-platform/blob/d497e194623dd32ad5a66f141529129267db645c/lms/djangoapps/courseware/views/index.py#L372-L441
+
+Rationale
+=========
+
+The approach above leverages the existing XBlock/Django infrastructure to handle de-duping and rendering, so little new code should be needed for a working implementation. This also means that changes to script loading in the XBlock runtime (e.g., AMD-style loading `as brought up on the edx-code thread`_) will automatically affect course-wide script loading too.
+
+For the user-facing portion of the feature, a course policy setting (which would appear on the *Advanced Settings* page of edX Studio) seems like a good fit:
+
+- Research script inclusion fits well within the `existing description and warning`_ on the Studio *Advanced Settings* page:
+
+    **What do advanced settings do?**
+
+    Advanced settings control specific course functionality. On this page, you can edit manual policies, which are JSON-based key and value pairs that control specific course settings.
+
+    Any policies you modify here override all other information you've defined elsewhere in Studio. Do not edit policies unless you are familiar with both their purpose and syntax.
+
+- A policy setting requires no additional UI (display and editing is already handled by Studio), reducing implementation and maintenance burden
+
+- Policy settings are conceptually course-wide in scope, which matches the scope of course-wide scripts
+
+.. _as brought up on the edx-code thread: https://groups.google.com/d/msg/edx-code/T83TDxhH74E/cOKZkpkTAQAJ
+.. _existing description and warning: https://github.com/edx/edx-platform/blob/d497e194623dd32ad5a66f141529129267db645c/cms/templates/settings_advanced.html#L83-L86
+
+Why not XBlocks?
+----------------
+
+Custom XBlocks cannot be instantiated at the top level of the courseware; as with script tags, they would need to be added to every vertical in a course for course-wide experiments. The `XBlock course tabs`_ idea on edx-code (originally referred to as `course-level XBlock views`_) would permit blocks to be loaded outside of verticals, but these would only be active when a user explicitly navigates to their tab, not when users are interacting with the courseware as required here.
+
+.. _course-level XBlock views: https://groups.google.com/d/topic/edx-code/Xlfof0JFlMo/discussion
+.. _XBlock course tabs: https://groups.google.com/d/topic/edx-code/ywjXV0wzQiw/discussion
+
+What about security?
+--------------------
+
+The open edX platform already permits arbitrary JavaScript in course content via :code:`<script type="text/javascript">` tags, and such scripts may already be deployed course-wide (albeit tediously) by duplicating them across every vertical in a course. *This OEP merely aims to make these things easier, not to add abilities which authors do not already have.*
+
+Nevertheless, at least one ability difference does exist between the existing and proposed mechanisms: scripts will execute earlier when associated with the courseware fragment instead of a content block fragment, which would give them slightly more control over the loading of subsequent page resources. The implications of this change should be considered carefully, though the author of this OEP is not currently aware of any vulnerabilities that this would introduce.
+
+It is also possible that this feature could increase script usage overall and thus increase the probability of users creating a vulnerability. To mitigate this risk, the description text for the policy key should carry a stern warning to authors similar to the following:
+
+    **WARNING: Custom scripts gain full access to staff and student user accounts**, including the ability to act as these users on Studio, discussion forums, and instructor dashboards. Use this feature with extreme caution. Only include scripts from experienced, trusted JavaScript authors, and only include scripts hosted on secure trusted servers.
+
+(As an aside, it might be valuable for the existing script-tag feature to carry a similar warning, though that is outside the scope of this OEP.)
+
+Security was briefly touched on in the `initial edx-code thread`_ for this feature. One idea mentioned there (but not yet adopted in this OEP) is to allow only :code:`/static/*.js` scripts; this would prevent exploits in which an attacker compromises a server hosting an included script and replaces it with a malicious script, but it would also differ from the behavior of :code:`<script type="text/javascript">` tags which may use any valid :code:`https://` URL. Another comment raised the idea of a feature flag:
+
+    "We might want to make this feature be something that can be disabled if a given installation is not comfortable with giving this power to its authors. As you point out, the power is already there through multiple other mechanisms, so maybe this isn't a concern."
+
+.. _initial edx-code thread: https://groups.google.com/forum/#!topic/edx-code/T83TDxhH74E/discussion
+
+What about performance?
+-----------------------
+
+Several performance-related points were also raised via edx-code. The first concerns the use of this feature to load JavaScript libraries (e.g., for things like course-wide code syntax highlighting):
+
+    "Having the assets tied to the course means that even if the block [that uses these assets] is removed, the assets would still be loaded. It would be better to have them requested on-demand by only the blocks that need them. Having said that, we don't have a mechanism in XBlock to allow multiple blocks to share the same library."
+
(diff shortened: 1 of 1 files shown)


jolyonb:
This would be really useful! :+1:

dseaton:
Absolutely - will forward to some of the folks at HarvardX that I know would be excited.

Colin-Fredericks:
Holy God yes this would be great. Cannot put enough thumbs-up for me. I have plenty of experience adding script tags to every vertical or every assessment in a course, and this would be very helpful.

benpatterson:
I feel compelled to point out that in-course JS has historically been a source of confusion, accessibility issues, performance issues, and potential security vulnerabilities. Our poorest performing pages on edx.org are oftentimes at the hands of custom JS, and inaccessibly components are often at the hands of JS. 

I think turning this on without any failsafes (beyond the 'stern warning') could open up a can of worms. I recognize we already allow custom JS in courses, but it's historically been problematic, and I hesitate at expanding that capability without any attempts to protect a given website from poor JS implementations.

Not saying no (from my perspective), but I am asking for some more girth around protecting us from ourselves.

Colin-Fredericks:
@benpatterson There are definite challenges to writing javascript that plays nicely within edX. Some of this is because there is little to no documentation on the topic, and no one to consult with.

Help us out here. Let us know how to do things better. What causes performance issues for you? What security issues have you seen? What components have hooks for scripts (the way the video component does)? What libraries can we assume are available in what parts of the platform, and which ones might be removed over the next six months?

I definitely hear you on the accessibility side of things. It's something I've worked hard to consider in my code, and I'm still learning. If there needs to be some kind of accessibility review for course-wide code, that seems pretty reasonable to me.

benpatterson:
@Colin-Fredericks I am sure with feedback on anything specific put into place, you'd resolve. :) It is moreso a question for me on how powerful (and vulnerable to errors/issues) we make the content management. Course-authors are not always developers, and we've certainly had instances of copy/paste javascript that a course author picks up and puts on a page for widgets, etc. 

I would just like some discussion on how resilient to mistakes the app should be, and the costs to allowing room for greater author-driven error.

Colin-Fredericks:
@benpatterson What sort of failsafes are you thinking?

jbarciauskas:
Should there at least be a size limit? Some form of linting? I think our options are relatively limited here, but on the flipside so long as we have the HTML block, we're just making peoples' lives harder.

Colin-Fredericks:
@jbarciauskas Linting sounds reasonable to me. I've used JSHint without feeling antagonized by it. :) Haven't tried ESLint yet.

Some sort of size limitation also makes sense, with the caveats that a: some people might try to get around this via minification, and b: I already have hx.js loading multiple other javascript files, and that seems like it would get around a size limit. Either of those might or might not be ok with you; I'm not sure.

msegado:
Thanks all for the input & discussion! In short: I think a good solution would be a per-course script whitelist controlled by ed-techs (or others who are qualified to understand the impact of injected JS). This could eventually apply to `<script>` tags, too. Expect a more detailed reply soon!

benpatterson:
In terms of failsafes, I was thinking along the lines of security vulnerabilities (XSS checking, perhaps?), and a11y issues, depending on what the javascript is doing with the DOM. 

I see @msegado's suggestion, and that brings up the meta-discussion that I was hoping to inspire, which is, in general what level of control do we give course authors in this area. I understand we already have liberal controls in HTML blocks, but since this widens the scope I wanted to ask that question. Perhaps @msegado introduces a new layer between course authors and the platform which could be worth consideration.

Colin-Fredericks:
I encourage you to give course authors as close to full control over their courses as possible.

As a course author, it is somewhere between annoying and infuriating when there is something that I know I should be able to do, but which the edX system prevents me from doing. Programming within edX three years ago was akin to being in a glass maze: you could see where you wanted to go, but you had no idea how to get there, and you ran constantly into barriers that you didn't understand.

This has gotten much better over time. (Also I suck less at coding now.) The vast improvement in documentation has been one big part of that, but the Raw HTML component is another one, because it doesn't delete things that I write or replace them with something else.

Fortunately, as far as I know so far, there is no change I can make to the Raw HTML component that will screw it up so badly that I can't delete that unit from the course outline. Similarly for this proposed feature: if it only applies in the Preview and Live views, it won't change my ability to fix things in Studio. 

I also realize that I might be coming at this with a different set of concerns in mind. If our goal is to have course authors create things that work well within the edX system and provide a good experience for the users, in my opinion we should be doing that with assistance, examples, and guidelines, not with restrictions. If we're worried about what happens if someone maliciously gets course author access and adds malware to the script on every page, then it makes more sense to me to add restrictions.

msegado:
OK, more detailed reply (per my earlier comment) -

@benpatterson, I think you raise some very valid concerns about custom JS in courses, and also think nailed one of the core underlying problems: "Course-authors are not always developers", especially when it comes to nuances like XSS and accessibility. At the same time, I empathize with @Colin-Fredericks and others who want to retain as much control as possible... after all, I opened the PR because I hope to use this feature myself for a multi-course research project 😉 (we can chat offline about details/timeline if that would be helpful).

I think a helpful way to evaluate this issue is to bring it back to [edX's mission statement](https://www.edx.org/about-us):

1. Increase access to high-quality education for everyone, everywhere
2. Enhance teaching and learning on campus and online
3. Advance teaching and learning through research 

Note that there is some tension here: (1) and (2) benefit from high consistency, accessibility, and performance, while (2) and (3) benefit from the flexibility to rapidly experiment with new ideas. Not only that, edX and the many course teams are all working toward these goals from different angles, with different levels of experience and domain knowledge, and with limited coordination. It's not at all surprising that there have been issues!

What we want then, is a solution with the following properties:

* Allow those with domain knowledge to experiment with as few barriers as possible.
* Provide enough safeguards so they don't shoot themselves (or edX, or the users) in the foot.

Automated XSS checking as Ben suggested would be great, but I suspect it's practically impossible: detecting even simple things in [heavily obfuscated JS](https://github.com/neuroradiology/InsideReCaptcha) can be extremely hard, and even if you could I'm not sure you could distinguish between a script's legitimate purpose and accidental/malicious security holes. Accessibility is probably similarly hard to automate.

As I mentioned earlier, another option would be to allow arbitrary scripts but require that they be whitelisted by someone close to the course team with enough web experience to evaluate their effects on performance, security, and accessibility. Whitelisting privileges could be attached to either an existing platform role with a high level of trust (e.g., to all org-level staff, with the understanding that they'll have scripts vetted by someone with experience before approving them) or to a distinct role as was done with `CourseCreatorRole`. There are many open questions with this idea, though:

* What exactly should be whitelisted? URI patterns, or something else?
	* I don't see an obvious way of doing anything other than URIs unless all the content lives on edX servers... [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) would allow content hashes from external sources but it's not yet supported in Safari, IE, or Edge.
* How would inline scripts be handled? These can be helpful for initializing libraries, etc.
* What's the best way to handle whitelister permissions/roles?
* How and where would the whitelist be stored?
	* In a database table with restricted permissions?
	* In the course policy, but secured with with a cryptographic signature from someone with whitelist privileges?
* How do we handle the whitelist when importing a course from XML (e.g. for a rerun)? Will someone with permissions need to re-authorize all the scripts?
* Is any new UI needed for with this approach? (I hope not...)

This is also a much more involved feature than I proposed in the original OEP, and I'm not sure I'd personally have the bandwidth to implement it. It's also likely take more time to land and start being useful. We could consider implementing it in stages...

1. Implement course-wide scripts with the same behavior as existing `<script>` tags (i.e. no restrictions) to make life easier for people like Colin and I. At the same time, deprecate _both_ the existing `<script>` tag behavior and the course-wide script behavior, and warn course authors that these will only allow whitelisted scripts in the near future.
2. Implement a first-pass whitelist (e.g. URI-based), but don't yet enabling enforcement to give existing courses time to add their scripts.
3. After allowing enough time for courses to make changes, switch on enforcement.
4. Continue to add extra security features like hash validation for `/static`-hosted scripts or subresource integrity once it's well supported.

...but of course I realize it's hard to guarantee that 2, 3, and 4 will ever happen with this approach, especially if they're not on edX's official roadmap.

Thoughts?

msegado:
> Automated XSS checking as Ben suggested would be great, but I suspect it's practically impossible

I stand corrected (h/t @jmclaus for sharing this): https://developers.google.com/caja/. This could be a great option for securing both global scripts and existing `<script>` tags in the future, though as I said above, it's a much bigger project than what I was proposing.

I suppose it comes down to whether "make the existing functionality safer" happens before or after "make the existing functionality easier to use". I can see arguments both ways; making global scripts harder to deploy adds at least a small barrier to footguns until concerns with the existing implementation are addressed, but in the meantime it also makes legitimate uses of the feature harder.

@benpatterson, any thoughts from edX's side on how to move forward?

pdpinch:
Sorry to arrive late to this discussion, but I think I can add a couple of thoughts:

I have to acknowledge @benpatterson's concerns, but I think making steps towards centralized management of external JavaScript actually gives you a path forward to managing the risks. As things stand now (and the way scripts have been implemented in practice) you might find yourself chasing hunting down bugs across multiple course units and scripts. If they were centralized it would make it easier to find any misbehaving scripts, and address any issues at the source. 

I think @msegado's comment on Feb 7 gives a reasonable path forward for iterating towards something that's both more useful and safer. Would it help move this OEP along to incorporate that into the proposal? Or does it make the scope too ambitious?

Finally, in defense of @Colin-Fredericks's argument for course-author freedom, there's a huge amount of teaching innovation that JavaScript allows. And no matter how risky it may be, it still doesn't have the same risks (and the necessarily slow review cycle) of Xblock development. If you make innovation too hard, it just encourages course authors to go off-platform.

benpatterson:
Thank you @msegado @Colin-Fredericks @pdpinch 

I recognize that I look at this proposal with one particular lens. I will solicit some additional opinions to get a wider-angle view, as it were. I do have some concerns, but feel you are addressing those.

pmitros:
Most JS is decentralized. The majority use case is embedding an interactive of some form in one place in a course. Similar JavaScript on every page is an important use case, but a minority one. 

My proposal for making JavaScript safe -- perhaps outside of the scope of this OEP -- is that course teams, by default, *cannot* add new JavaScript to their courses, but can be granted that access. 

There is a set of people who can give course authors that permission. On edx.org, that would be the same as the set of data czars. If a course team at Harvard wishes to use JavaScript, they ask the same person they ask for course data. That person could confirm that the course team is appropriately set for any legal, security, accessibility, etc. concerns. If they are comfortable that the course team can manage everything correctly, they grant them appropriate access.

tobz:
We can (and should) make it easier for course authors to manage the web assets they need for courseware.  I think this OEP helps that along in quite a measurable way.  Just wanted to say that upfront. :)

Some of these things are covered in the OEP, but just to review / stream of consciousness some things:
- we _should_ only allow JS that is uploaded (no inline script blocks, no external URLs)
- opt-in access sounds like a good compromise of providing the ability for course authors do the things they need, once we've vetted their needs
- we _should_ strive for sanitized/sandboxed JS (maybe Caja as a one-time-during-upload step? or something homerolled that looks for obvious flaws)
- we _should_ consider dependencies issues; we load JS a certain way, and things need to fit that shape and, also, we don't want libraries double loaded

As an engineer on the platform, one of my biggest worries is: how does the page experience potentially suffer when a course author does something off the beaten path?

To me, this is something like inline script tags (because what happens if we turned course unit navigation into PJAX and we're double, triple, quadruple, etc, loading the same JS) or loading external resources (because their external URL goes down, or now we have an external attack vector for someone to replace with malicious content that is now served from our courseware).

I'm not a JS wizard, but I would like to explore (even if only scratching the surface to figure out the options) whether or not we can sandbox this code client-side with a blunt instrument (like namespace rewriting... protect `window` and another other globals we care about by replacing them with a per-course namespace or something).  That's the one thing that really makes me go "oh mannnn" when thinking of making it *easier* for course authors to include more custom JS in courseware.

A lot of the other bits could be some form of documentation / procedure, I think.  Like, how we load JS, and what that means for the type of module compatibility your JS needs to be loaded properly by the XBlock runtime.

pmitros:
@tobz Sandboxing JavaScript is a rabbit hole. It's essentially impossible to do well, and allow the flexibility people use JavaScript for. As much as I appreciate the argument that people might shoot themselves in the foot, I don't think it's a correct one:

1. We can't stop it. People can shoot themselves in the foot by e.g. uploading hour-long videos. Or having bugs in problems. 
2. Most course teams are typically quite competent. We do stupid things in code all the time too. I don't have any reason to believe course teams will be less competent than us. If you think using edx-only code keeps a good page experience, try unsubscribing from emails from many courses, authoring a drag-and-drop component, or applying for financial aid. I'm not trying to deprecate edX; any engineering team has issues. But we do allow course teams to put up courses on edX.org, and that involved some level of respect, delegation, and trust. 

There is one sandboxing technique which goes a long way -- per-course domains. Early MOOCs did this. The first ones were ai-course.org, db-course.org, ml-course.org, 6002x.mitx.mit.edu, and similar. This means that a JavaScript issue within any course is constrained to just that course. I'd be in favor of that. Cross-course scripting is a vulnerability. 

I'm also in favor of black/white lists. Most course staffs do not write JavaScript. We have 1000+ courses. That's a big perimeter. Simply making the process explicit would prevent casual errors (e.g. a random course grants authoring access to the wrong person). 

I'm also in favor of supported client-side APIs. Course teams scraping our HTML, or modifying CSS is perhaps not the most robust way to do things.

tobz:
All fair points.  Suffice it to say, for better or worse, we (edx.org) must both manage the direction of the platform while simultaneously operating it.  I don't think anything, to be quite honest, is out of bounds for us to consider when there's a potential impact to edx.org.

> We can't stop [people from shooting themselves in the foot].

With certainty and precision, no, we can't.  We can, however, try to limit the distance the that the debris travels.  This is what I was alluding to with sanitized JS that, maybe, rewrites globals in order to protect global edX JS namespaces, or something along those lines.

> We do stupid things in code all the time too.

Let's replace "do stupid things" with "make mistakes," to start.  We all make mistakes, and nobody is arguing the opposite.  Further, nobody is arguing that anyone is incompetent.

Instead, there is a real possibility that the JS that a course team loads will conflict with the base JS of the platform.  This could be double loading a library, or stomping over globals.  There's the performance aspect: adding event listeners with a selector that has very little specificity, and hitting more and more elements as content changes, etc.  There's the code migration aspect: what if we want to perform refactoring that will affect third-party JS in a way that we cannot easily gauge and thus cannot move deftly on platform work?

I don't work on the front-end team(s), so I can't attest to whether or not we've been bit by these types of issues before.  If we're writing an OEP about it, though, they bear mentioning because they are not imaginary problems.  They happen in the wild.  I can regale you with stories from a previous job where we let clients use their own HTML/JS, and all the ways it can (and did) go wrong.

> per-course domains

I think it's safe to say we won't / can't do this.  Opaque keys make this next to impossible.

> black/whitelists

Yes.  I agree.

> client-side APIs

Yes!  We don't have enough good sources of data/hookpoints for XBlocks and client-side JS to provide the rich experience that course authors desire.

pmitros:
I believe sanitized JS would be several times the work of per-course domains. However, I would say both are more ambitious projects than the organization could engage in right now.

Colin-Fredericks:
> Yes! We don't have enough good sources of data/hookpoints for XBlocks and client-side JS to provide the rich experience that course authors desire.

Amen. This leads to more workarounds and grody code. Remind me to show you the visual course progress map I have that auto-updates based on student grades... by iframing in their Progress page and scraping it for their scores. Luckily I know that's too hacky and fragile to put into a real course.

oeps/oep-0015.rst
```diff
@@ -0,0 +1,144 @@
...
+
+This is a valid point; there's a risk of vestigial libraries being loaded with this approach. The simplest way to address this concern is to recommend that authors not use this feature for libraries and instead continue loading these using the existing mechanism (i.e., by adding script tags in any blocks where they are needed).
+
+Another comment mentioned the impact of additional HTTP requests:
+
+    "There are performance implications to loading a number of individual files like this. Having said that, it would be difficult to have individual courses contribute files to the static asset pipeline, since courses can be created/imported after the LMS has been stood up."
+
+Some benchmarking will likely be appropriate here, though hopefully the benefits of this feature will outweigh any performance impact if only a few scripts are included. Caching should help to reduce the duration of each request after the initial load: it looks like edX sets the Cache-Control max-age to 1 year, and external JavaScript CDNs should also have reasonable caching behavior. Hopefully HTTP/2 will also help as it gains adoption.
```
andy-armstrong: Note: only files with MD5 hashes are cached with a 1 year lifetime. Files that have no MD5 hash are uncached as there is no ability to know whether the content has changed. We would need to find a way to get these files to be served with an MD5 hash in order to benefit from caching. It is probably possible to model this on the way XBlocks serve their assets.

oeps/oep-0015.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Motivation
+==========
+
+The primary goal of this OEP is to facilitate research and experimentation by removing the bariers to large-scale JavaScript inclusion. Simplifying the inclusion of scripts is especially important for experiments spanning many courses (e.g., all MITx courses): it is not reasonable to expect all course teams at an institution to manually modify every single vertical in their course to include a research script as is the case with the current platform, but it is quite feasible to request that they make a single easy-to-reverse change to their course policy.
```
andy-armstrong: Typo: bariers

oeps/oep-0015.rst
msegado: Ah, interesting - though I just spent some time digging around and found that this feature was actually added last summer! 🎉 Static assets URLs now include MD5 hashes (provided by Mongo's GridFS) and are served with a 1-year lifetime: https://github.com/edx/edx-platform/pull/12573. [As an aside, that specific PR was later reverted but the feature was reinstated after some corrections; it seems to be working now based on the URLs/headers I'm seeing in my requests.]

oeps/oep-0015.rst
msegado: Whoops, thanks. Fixed it (and ran a spellcheck on the remainder...)

msegado:
(Bumping this issue to make sure it doesn't fall through the cracks. Let me know if there's anything I need to do on my end.)

benpatterson:
Hi folks on this thread. I was the first to reply, but I am not the approver. Just wanted to be explicit about that. I appreciate the dialogue amongst everyone.

CC @msegado

msegado:
Thanks for clarifying, @benpatterson!

I just re-read the OEP process doc and realized that I was supposed to have asked @efagin to assign an arbiter (oops):
> After the Author drafts an OEP in a format in which they are comfortable, they will request an Arbiter from the edX Chief Architect.

So: I am officially requesting an arbiter from the edX Chief Architect 🙂

efagin:
Hi @msegado - no worries. Once we return from the Open edX conference, let's get this one to a conclusion (one way or another). I don't want to keep this in limbo.

I'm copying @scottrish on this OEP as his team ultimately would be reviewing any related implementations. Scott, I'm temporarily assigning you as the arbiter here, but feel free to delegate as needed.

scottrish:
Hi @msegado, just wanted to give a brief update.  We're having some internal discussions regarding the risks already amply discussed in previous comments. I hope to have an update next week.

scottrish:
Hi @msegado.  This OEP is approved. Could you please update the status before it's merged?

Thanks.

msegado:
@scottrish et al., thanks for the great news and for your work on this! I've updated the status/date/resolution and also filled in the arbiter field, but of course feel free to reassign as needed. Let me know if you want me to manually squash the commits too (I believe GitHub now has an option to [squash when you merge](https://github.com/blog/2141-squash-your-commits) but I don't know if it's been enabled for this repo).

PR #44, opened 2017-01-25
Fixes for OEP-12

andy-armstrong:
This fixes a few stylistic issues with the documentation for OEP-12:
 - add missing image
 - fix broken image link
 - fix broken enumerated list
 - fix broken bulleted list


cahrens:
@andy-armstrong this doesn't look correct.

image

andy-armstrong:
@cahrens great catch. I've pushed up a (squashed) fix for that too.

clintonb:
@efagin @cpennington who controls the RTD account for this repo? It would be ideal if we built PR branches so we can catch RST formatting issues before they are merged.

andy-armstrong:
Great point, @clintonb. I considered building it locally but didn't get around to it. 

BTW, RTD regenerated and the issues all look resolved now. Hopefully I won't need a second follow-up!

http://open-edx-proposals.readthedocs.io/en/latest/oep-0012.html

cpennington:
@clintonb: I have management rights (as does the shared edx account, I believe). That said, I don't think RTD has a feature for automatically building PRs (https://github.com/rtfd/readthedocs.org/issues/2465), so we'd need to build them manually.

PR #46, opened 2017-04-21
OEP-16: Bootstrap Adoption

andy-armstrong:
This is the first draft of a proposal for Bootstrap adoption.

You can see the rendered proposal here: 

https://github.com/edx/open-edx-proposals/blob/1fb54d1491fea382521bdec717d1e28f20c3a1cf/oeps/oep-0016.rst

FYI @edx/fedx-team @ormsbee @clintonb @efagin @douglashall


oeps/oep-0016.rst
```diff
@@ -0,0 +1,195 @@
...
+The recommendation is to adopt Bootstrap 4 starting immediately so that new
+features can be built following this best practice. In edX platform, the core
+templates will be parameterized to support the option of using Bootstrap. With
+this option chosen, the page skeleton (header, footer, navigation) will
+automatically render in a mobile-first design using Bootstrap components.
+
+New Bootstrap themes will be implemented in the `edX UI Toolkit`_. These themes
+will be provided as Sass partial files that can be included into any project's
+Sass file. The files will be as follows:
+* A partial defining the standard edX fonts, layouts, margins etc
```
clintonb: Add a new line before this one so the RST renders properly.

oeps/oep-0016.rst
```diff
@@ -0,0 +1,195 @@
...
+* A "uses_bootstrap" context parameter enables a template to add Bootstrap JS
+* The header was reimagined with Bootstrap components
+* The page styles were updated to use Bootstrap mixins and theme variables
+
+For the courseware, a new lms-course-v3.scss was added
+
+* This file included all of the course-specific Sass partials
+* Any global-level rules were excluded to allow Bootstrap's styles to be used
+
+See the `Bootstrap Proof-of-Concept PR`_ for more details, including screenshots
```
clintonb: We should also do a proof-of-concept for theming for both single and multi-tenancy. Theming for a single tenant is relatively easy, but not so much for multiple tenants as evidenced by the fact that we still haven't settled on a single method of theming:

* https://github.com/edx/edx-themes
* https://github.com/edx/credentials-themes

andy-armstrong:
Thanks, @clintonb. I should have been more explicit how I see IDAs working with this, so I'll revises this section. The UI Toolkit is an npm package which any IDA can install, and my proposal is to include the Bootstrap theme Sass files in this package. We discussed having a separate repo, but that seemed like overkill for what we currently see as only being three files. In addition, the reusable React components that @arizzitano and @bjacobel are working on will be published from the UI Toolkit too, so we figured that every IDA will already be including it. Let me know if this seems reasonable to you, or if you see benefits to introducing another npm package.

oeps/oep-0016.rst
andy-armstrong: I'd love to talk to you more about theming, as I'm only familiar with comprehensive theming in edX platform. For that, there is already a separate Sass file per site, so I was thinking that each file would just include its own Bootstrap theme partial. Let's get together with @douglashall, @gsong, and other interested theming folks to come up with a consistent approach.

OmarIthawi:
@andy-armstrong Despite the hate for Bootstrap, I'm a huge fan of it and it's simplicity in both code and UX.

I think it's a better step to include Bootstrap as a foundation to new Open edX features and eventually migrate the existing ones as well.

There's a huge caveat. Bootstrap doesn't support RTL by default despite various (low quality?) [attempts](https://github.com/twbs/bootstrap/pulls?utf8=%E2%9C%93&q=is%3Apr%20rtl). The Bootstrap team is planning to [support RTL once a v4 is out](https://github.com/twbs/bootstrap/issues/19555#issuecomment-197549158).

I'm more than happy to help on this regard.

andy-armstrong:
Thanks, @OmarIthawi. I forgot to add that @bjacobel did an investigation about RTL here: https://openedx.atlassian.net/wiki/display/FEDX/Bootstrap+spike+findings. We're currently thinking of using https://www.npmjs.com/package/postcss-rtl. I'll add this in my next draft.

OmarIthawi:
@andy-armstrong. Thanks for pointing out to the spike findings. I'll share more about my experience in making RTL interfaces.

RTL support might seem to be perfectly automatable, in fact it has a lot more complications that just flipping some margins here and there. Consider the following example that process-all tools like `postcss-rtl` won't support:

 - Flipping font and SVG icons.
 - Showing the same video player regardless of the direction (it's a famous acceptable LTR use in RTL context).
 - Avoid flipping and corrupting logos and branding in general.
 - Yet, require flipping the decorations and any page visuals.
 - JavaScript and CSS RTL support? This really requires the touch of an artist to provide an elegant support.

I'm not trying to make another anti-automation lost case. But I've been working with RTL sites for the past 10 years and the only elegant solution I've found so far is the `bi-app-sass`.

Believe me `bi-app-sass` was a love from first sight, after I tried the following:

 - Duplicate the LTR file to RTL and make minor changes: Maintenance horror.
 - Make RTL override file: Maintenance horror.
 - Use CSS post processing tools: Requires a lot of manual post processing, and it often makes the incorrect guesses (recall Python2's *implicit* Unicode <> str conversions? a similar experience)
 - `bi-app-sass`: Seriously, love from first sight! I tried it and loved it. It works just fine. However you have to be conscious about your CSS, and make *explicit* styling. Much like what Python3 did for str <> bytes. It requires some work, but the result is always predictable. 

Please let me know if we should move the RTL tooling to another thread/OEP.

oeps/oep-0016.rst
```diff
@@ -0,0 +1,195 @@
...
+
+* accessibility
+* internationalization and right-to-left languages
+* theming
+* performance
+* mobile-first responsive designs
+* extensibility / ease of overriding
+* modularity / scoping
+* Open edX community interest
+* size and activity of the framework's community?
```
dsjen: Is the "?" here to indicate this goal might be cut?  It seems to be by the rubric item, "Community project support."

oeps/oep-0016.rst
```diff
@@ -0,0 +1,195 @@
...
+* integration with build systems
+* integration into existing edX pages
+
+Bootstrap is the industry standard web framework with a comprehensive collection
+of web components. Each of these components is designed to be mobile-first,
+meaning that Bootstrap applications look good at phone, tablet, and desktop
+sizes.
```
dsjen: @bjacobel -- agree on the mobile-first comment.  It's about the implementing for the primary use case and thinking of all other cases are exceptions.

andy-armstrong:
@OmarIthawi @clintonb @dsjen @bjacobel et al, thank you for your excellent review comments. Here's where I think we stand right now:
 - I agree with the recommendation that we have a separate edx-bootstrap repo rather than conflating it with edx-ui-toolkit.
 - We need to reach a common understanding as to how to handle RTL support, given that Bootstrap does not yet have a recommended approach. I'm doing a discovery story this sprint: https://openedx.atlassian.net/browse/FEDX-352

Once we've determined an approach to RTL, I will publish a second draft of this OEP to cover these two topics plus the other feedback on this PR.

andy-armstrong:
@dsjen @nasthagiri - @efagin suggested both of you as potential arbiters for this OEP. I'm going to publish a second draft today, so would either of you like to volunteer? Thanks!

dsjen:
@andy-armstrong -- I'm happy to be arbiter!

andy-armstrong:
@OmarIthawi @clintonb @dsjen I've published my second draft based on all of your feedback, and the results of the discovery story for right-to-left support. I'd love your feedback when you have some time.

dsjen:
@edx/fedx-team & @cptvitamin (and others) -- I wanted to check in and see if you have any feedback or questions about this OEP for adopting bootstrap.  We're looking to gather all feedback within a week, by May 24th, if possible. Thanks!

cptvitamin:
@dsjen no objection to using Bootstrap 4 from me. I'm fairly confident that it can still be used to create accessible interfaces.  However, some extra care and attention may be required.  It's also important to note that we will not be able to use any feature that doesn't fully conform (or cannot be  made to conform) with WCAG 2 AA.

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+at this point that the `edX Front End Working Group`_ decided to investigate
+other alternatives.
+
+For more information, see the wiki page `State of the Pattern Library`_.
+
+Material Design
+---------------
+
+Google's Material Design is another very successful web framework that was
+evaluated. It was ultimately considered to be too opinionated to support the
```
gsong: I don't remember when this conclusion was drawn. And I personally think Material design was conceived for a [wide variety of use cases](https://material.io/guidelines/) for material design, specifically

* [Accessibility](https://material.io/guidelines/usability/accessibility.html)
* [Bidirectionality](https://material.io/guidelines/usability/bidirectionality.html)

[Material Components](https://github.com/material-components/material-components-web) are also kept up to date by Google.

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+Technology Selection
+====================
+
+The following rubric was used to assess potential technologies:
+
+* Learning curve / prior experience
+* Options for community
+* Frameworks that are opinionated vs. flexible
+* Ability to easily build accessible front-ends
+* Ease of design-framework integration
```
clintonb: Nit: "design-framework" is confusing upon initial read. Perhaps just, "Ease of integration".

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+* It provides a more flexible grid system that supports all of edX's layouts
+* It uses ems and rems for responsive typography
+
+Another key factor in the switch is the wide availability of Boostrap-based
+React components. With edX's move to adopt React (see
+`OEP-11 - Front End Technology Standards`_), this will provide the community
+with an ever increasing set of components with which to extend the platform.
+
+Finally, a critical benefit to using Bootstrap is that the edX community
+does not need to build or work with a proprietary markup scheme. XBlock authors
```
clintonb: I'm guessing this refers to the custom HTML needed, for example, to render custom components in the [Pattern Library](http://ux.edx.org/components/alerts/).

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+
+* The default is for right/left properties to be switched which is usually
+  the desired result. With bi-app-sass, every property that needs to switch
+  needs to use a function, and it is very easy to forget to do so.
+* For special case situations, rtlcss provides declarative comments that give
+  the developer to control the behavior.
+* Having only a single CSS file that adapts to the desired language
+  direction makes asset management simpler.
+
+Given this decision, the reference implementation was updated to use `rtlcss`.
```
clintonb: What reference implementation? This is the first mention of an implementation. Can we get a link?

Never mind. I see the mention below. It might be worth calling out the fact that details about the reference implementation are below.

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+This proof-of-concept successfully demonstrated pages rendered right-to-left
+which confirmed the choice.
+
+Note: once Bootstrap chooses its own approach, it will be necessary to revisit
+this decision. It might be simpler, for example, to switch to use the same
+technology for simplicities sake.
```
clintonb: simplicity's sake

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+account.
+
+Backward Compatibility
+======================
+
+For edx-platform, the approach taken in the reference implementation was to
+introduce a new v3 style, thus leaving all v1 and v2 styles unaffected. This
+allows pages to be converted one at a time.
+
+An investigation was performed to see whether Bootstrap components could be used
```
clintonb: What about "global" components like the header and footer?

azizur:
What if I want to use something completely different than Bootstrap?

OmarIthawi:
Hey @andy-armstrong, thanks for the update. Since I haven't used `postcss-rtl` I'll trust the results from the discovery story.

caesar2164:
Thought: Bootstrap stuff could be put inside of a new directory inside `lms/static/bootstrap`

azizur:
@andy-armstrong Correct me if I am wrong, wasn't there something about adapting BEM? How will this work with Bootstrap? Because as far as I know, Bootstrap has it's own selector naming conventions and do not do BEM? Would you then need to have an intermediate step to map CSS classes in Django Templates in SASS?

oeps/oep-0016.rst
```diff
@@ -0,0 +1,261 @@
...
+Bootstrap is the industry standard framework for desigining web applications. It
+comes with a comprehensive collection of web components that are designed to be
+responsive, meaning that Bootstrap applications look good at phone, tablet,
+and desktop sizes.
+
+Bootstrap 4 is the first version to support key requirements for adoption
+within the edX platform:
+
+* Bootstrap 4 has switched to using Sass (previous versions used Less)
+* It provides a more flexible grid system that supports all of edX's layouts
```
andy-armstrong: I was thinking of the combination of flexbox plus the improvements they describe here:

https://blog.getbootstrap.com/2015/08/19/bootstrap-4-alpha/

> **Improved grid system.** We’ve added a new grid tier to better target mobile devices and completely overhauled our semantic mixins.

> **Opt-in flexbox support is here.** The future is now—switch a boolean variable and recompile your CSS to take advantage of a flexbox-based grid system and components.

oeps/oep-0016.rst
andy-armstrong: The approach I took in the reference implementation was to conditionalize the header and footer templates with a "use_bootstrap" context parameter. If Bootstrap is being used, the template outputs Bootstrap components. I'll add this explanation.

andy-armstrong:
Thanks for the feedback, folks. Since these were mostly small changes, I'm hoping this will be the final draft. @dsjen, let's see if we can merge next week.

andy-armstrong:
@azizur I'm not aware of any proposal to adopt BEM. My recommendation is to follow the Bootstrap naming conventions. I wouldn't want to have any name mapping process as that would likely lead to challenges with debugging the resulting CSS, even with source maps.

> What if I want to use something completely different than Bootstrap?

For custom extensions to the Open edX code base, you can use whatever technology you like to style your features. However, the core code will all be built upon Bootstrap for ease of development.

andy-armstrong:
@caesar2164 Naming is always hard. I think having a Bootstrap directory is a good idea, so we'll see how that pans out when we start the implementation. My only concern is that there will be shared Sass partials used in both Bootstrap and non-Bootstrap contexts, so I don't know if that will be confusing.

dsjen:
@andy-armstrong -- thank you for all the responses!  Yes, I agree about merging next week as this has been open for feedback for quite some time already, and at the core, it looks like we've coalesced around bootstrap.

PR #47, opened 2017-04-25
Add a statement about LTS dependencies

nedbat:


nedbat:
@efagin @cpennington @doctoryes anyone else?

nedbat:
@clintonb does this make sense?

oeps/oep-0010.rst
```diff
@@ -38,7 +38,7 @@ process.
 Motivation
 ==========
 
-Open edX releases happen infrequently, every four to six months. EdX engineers
+Open edX releases happen roughly every six months. EdX engineers
 are typically focused on the weekly edx.org releases.  This document
```
doctoryes: Nit: `weekly` -> `daily`

oeps/oep-0010.rst
```diff
@@ -38,8 +38,8 @@ process.
 Motivation
 ==========
 
-Open edX releases happen infrequently, every four to six months. EdX engineers
-are typically focused on the weekly edx.org releases.  This document
+Open edX releases happen roughly every six months. EdX engineers
+are typically focused on the daily edx.org releases.  This document
```
efagin: It's not really daily, depends on which system we're talking about - we do multiple releases per day in aggregate. I'd just remove the word entirely to avoid unnecessary complication ("focused on the edx.org releases").

oeps/oep-0010.rst
```diff
@@ -103,10 +102,29 @@ parts of providing a finished quality product, and we don't want to overlook
...
+their developers for the entire Open edX release.
+
+Typically this means choosing Long Term Support (LTS) versions of the
+supporting layers, but it's possible shorter-term support versions will provide
+the support needed.
+
+The layers in question here are Django, Python, and Ubuntu.  Here's a `calendar
+of the known support windows`__ and how they overlap with Open edX plans.
+
+.. __: https://docs.google.com/spreadsheets/d/11DheEtMDGrbA9hsUvZ2SEd4Cc8CaC4mAfoV8SVaLBGI
```
efagin: For now, this will work, but I'd recommend putting a note in the spreadsheet to make it clear that it's linked from this OEP, just so you/others don't make changes there in the future that negatively impact the message here.

PR #48, opened 2017-05-05
Make README link to docs less confusing.

robrap:
I clicked on the wrong link to the wrong docs.  I can't imagine I'd be the last person to make that mistake.  I think it would be better to make the link clear that it isn't for OERs, and let someone find that if they need to.


robrap:
@andy-armstrong mentioned that there may have been an agreement to have a link to the old doc.  If so, I'll add an alternative that I think will contain both, but will be more clear.  
@cpennington Can you confirm that we need the other link?

cpennington:
I had an email agreement with the guy that was publishing to oep.readthedocs.com, but he never added a link on his, and it seems to have stagnated.

PR #49, opened 2017-05-05
Fix README keeping both links.

robrap:
An alternative if we wanted to keep both links.


robrap:
I am going to close this.  I think it is a good alternative if we do ever need the other link.  But, if we don't, it is cleaner.

PR #50, opened 2017-05-22
Include forks in oep-14.

benpatterson:
@efagin @cahrens @jibsheet 

PTAL as per our previous discussions.

### DIFF
--- a/oeps/oep-0014.rst
+++ b/oeps/oep-0014.rst
@@ -110,6 +110,17 @@ Include this statement in the README.rst file:
     issues, and security updates will not be provided. Pull requests against
     this repository will also not be merged.
 
+For Repos that are Forks
+------------------------
+If the repository is a fork of an upstream repository that is not within the edX organization, and will no longer be maintained, it can be deleted with the following steps.
+
+1. Review any differences between the upstream parent and the forked repo. Capture anything that should be kept for historical purposes.
+
+2. Review dependencies on this fork (i.e., ensure no one is using it anymore)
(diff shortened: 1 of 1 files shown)


oeps/oep-0014.rst
```diff
@@ -110,6 +110,17 @@ Include this statement in the README.rst file:
     issues, and security updates will not be provided. Pull requests against
     this repository will also not be merged.
 
+For Repos that are Forks
+------------------------
+If the repository is a fork of an upstream repository that is not within the edX organization, and will no longer be maintained, it can be deleted with the following steps.
+
+1. Review any differences between the upstream parent and the forked repo. Capture anything that should be kept for historical purposes.
```
cahrens: I don't think this will format correctly-- not the spacing on the other numbered things.

cahrens: Also, you will need to update the change history.

oeps/oep-0014.rst
```diff
@@ -110,6 +110,17 @@ Include this statement in the README.rst file:
...
+
+2. Review dependencies on this fork (i.e., ensure no one is using it anymore)
```
cahrens: Do you mean nobody in edX or nobody anywhere?

Assuming edX, I would just replace this line with "Ensure that there are no edX dependencies on this repository".

benpatterson:
@cahrens I updated based on feedback; LMK how it looks.

cahrens:
👍

PR #52, opened 2018-01-22
L'OEP Draft

robrap:
@nasthagiri: The following was my original work when considering introducing the L'OEP.  I have no idea what state it is in, but it may provide a start for a revised OEP-1 proposal?

I didn't review this.  I'm just posting what I had in case it helps you.


robrap:
OEPs themselves have gotten lighter and lighter.

PR #53, opened 2018-02-04
OEP-1: Update and simplify

nasthagiri:
The purpose of this PR is to update and simplify the OEP process defined in OEP-1.

**Review Period**

February 05, 2018 -> February 20, 2018

**Arbiter**

@cpennington 

**Motivation**

Simplify the process for creating/maintaining Open edX Proposals. Going forward, we want to encourage teams to write down and propose their architectural and system-wide decisions so we can align with best practices, processes, and system architectures.

Update the process given recent changes in the org (i.e., the formation of an Architecture team in lieu of a chief architect).

**Changes**
* Simplify process
  * Favor announcing on Slack over emailing edx-code.
  * For Best Practice OEPs, favor updating rather than replacing.
  * Reiterate option to have multiple authors to share the load.
  * Add an explicit "Review Period" so process is finite and clear.
  * Documentation readability
    * Slight rearranging of sections, with further table of contents.
    * Break down submission process in 5 clear steps.
    * Fix a few typos with State transitions.
* Replace edX Chief Architect with Architecture Team.
* Append type and brief title to an OEP's file name.
  * Rationale: Taking this pattern from [architecture decision records](https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records) ([example](https://github.com/npryce/adr-tools/tree/master/doc/adr))
* Remove "Product Enhancement" proposal type.
  * Rationale: Seems like a separate process from developer docs as the product team would want to weigh in on these decisions.
* Remove support for Google Docs for discussion.
  * Rationale: We are trying to consolidate our developer docs and want to move away from the exclusiveness and non-discoverability of Google docs.


oep-template.rst
```diff
@@ -4,42 +4,46 @@ OEP-XXX: OEP Template
...
++-----------------+--------------------------------------------------------+
+| `References`    | <links to any other relevant discussions               |
+|                 | or relevant related materials>                         |
++-----------------+--------------------------------------------------------+
 
 Abstract
 ========
 
-The abstract is a short (~200 word) description of the technical issue that
+The abstract is a short description of the technical issue that
```
feanil: Why remove the length here?  It seems like that is a reasonable guide so people don't have to infer what `short` means to them.

oep-template.rst
```diff
@@ -85,32 +77,20 @@ An OEP that introduces backward incompatibilities must describe the
...
 ========================
 
 The reference implementation must be completed before any OEP is given "Final"
-status, but it need not be completed before the OEP is accepted. While there is
+status, but it need not be completed before the OEP is "Accepted". While there is
 merit to the approach of reaching consensus on the specification and rationale
 before writing code, the principle of "rough consensus and running code" is
-still useful when it comes to resolving many discussions of API details.
-
-The final implementation must include test code and documentation, following
```
feanil: Why remove this requirement?  Do you find that we are not actually holding ourselves to this standard?

oeps/oep-0001.rst
```diff
@@ -226,68 +312,48 @@ Each OEP should have the following parts.
     a short descriptive title, the names, and optionally the contact info for each author.
 
 *Abstract*
-    A short (~200 word) description of the technical issue being addressed.
+    A short description of the technical issue being addressed.
```
feanil: Same as in the template.  I think having a suggested ballpark for what short means is useful.

oeps/oep-0001.rst
```diff
@@ -300,111 +366,84 @@ ReStructuredText [8] allows for rich markup that is relatively easy to read,
...
+| OEP             | OEP-XXXX-<brief-type>-<brief-title>       |
++-----------------+-------------------------------------------+
+| Title           | <OEP title>                               |
++-----------------+-------------------------------------------+
+| Last Modified   | <date string, in YYYY-MM-DD format>       |
++-----------------+-------------------------------------------+
+| Authors         | <list of authors' real names and          |
+|                 | optionally, email addresses>              |
++-----------------+-------------------------------------------+
+| Arbiter         | <Arbiter's real name and email address>   |
```
feanil: Given that an OEP can have multiple arbiters over time.  Do we want some sort of date associated with the aribter to indicate past aribters vs current arbiter?

oep-template.rst
nasthagiri: Overall, I reviewed OEP-1 through the lens of trying to simplify the process to encourage more folks to write down their decisions and have their designs reviewed before implementation.  (Given we have had only 13 OEPs written since its inception 2 years ago.)  So anything that I thought could be an unnecessary barrier, I tried to rephrase or remove.

Given that, having an explicit word count could be viewed as either a clear constraint to follow or an unnecessary restriction that the arbiter and reviewers can objectively comment on.  Right now, I'm erring on the side of less process - thinking we can always add process back in if needed.

oep-template.rst
nasthagiri: See my perspective above regarding the rationale for simplifying the process.

Given that, I find having instructions about the implementation to be orthogonal to the OEP-1 process.  I am thinking each OEP should be focused on its own goals, otherwise, they become too fat when they refer to other related aspects.

oeps/oep-0001.rst
nasthagiri: We don't have dates for authors either - as they can also change over time.
Can we rely on git for such details?

nasthagiri: For OEP-1, I went with providing a link to the PR for each arbitration.

oep-template.rst
```diff
@@ -4,42 +4,46 @@ OEP-XXX: OEP Template
...
-+---------------+-------------------------------------------+
-| `Replaces`    | <OEP number>                              |
-+---------------+-------------------------------------------+
-| `Replaced-By` | <OEP number>                              |
-+---------------+-------------------------------------------+
-| `References`  | <links to any other relevant discussions  |
-|               | or relevant related materials>            |
-+---------------+-------------------------------------------+
++-----------------+--------------------------------------------------------+
+| OEP             | :doc:`OEP-XXXX-YYYY-ZZZZ </oeps/oep-XXXX-YYYY-ZZZZ>`   |
```
cpennington: In the other OEPS, you've used the format `OEP-XXXX </oeps/oep-XXXX-YYYY-ZZZZ>`. Which one did you mean? Is the goal to add info just to the filenames? Or to both the filenames and the visible identifier?

oep-template.rst
```diff
@@ -50,32 +54,20 @@ existing architecture or process is inadequate to address the problem that the
...
 Rationale
 =========
 
 The rationale adds to the specification by describing the events or
-requirements that led the proposal, what influenced the design, and why
-particular design decisions were made. It should describe any alternate designs
-that were considered and identify any related work, for example, how the
-feature is supported in other systems.
-
-The rationale should provide evidence of consensus within the community and
```
cpennington: Is it worth keeping this paragraph, but as a suggestion, rather than a soft requirement? `The rational *could* provide evidence...`

oep-template.rst
bmedx: I agree with @feanil on this, though we could soften the wording to something like "We have found ~200 words to be a reasonable length for an abstract, but recognize that the length will vary based on the complexity of the OEP." to clarify both what we're looking for in length, and that it won't be a barrier to OEP approval.

nasthagiri: Sounds like a good compromise.  I'll wait a bit to hear if there are any other thoughts on this before making the update.

oep-template.rst
```diff
@@ -85,32 +77,20 @@ An OEP that introduces backward incompatibilities must describe the
...
-.. _Open edX Contribution Guidelines: http://edx.readthedocs.org/projects/edx-developer-guide/en/latest/process/index.html
-
+still useful when it comes to resolving many discussions.
 
 Rejected Alternatives
 =====================
 
 This statement describes any alternative designs or implementations that were
-considered and rejected, and why they were not implemented. It should also link
-to the original source of that discussion.
```
cpennington: It seems like a link to other discussions is still valuable to include, if it's available.

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
+
+Once the Authors have asked the Open edX community whether an idea has any chance
 of acceptance, a draft OEP should be submitted as a pull request against the
-`central OEP repository`_. To identify the draft proposal, the Author should
+`central OEP repository`_. To identify the draft proposal, the Authors should
 check the numbered list of previous OEP pull requests and select the next 
-available number. The pull request title must start with "OEP-XXX", which 
-claims that OEP number for the included proposal.
+available number. The pull request title should be of the form 
+"OEP-XXXX-YYYY-ZZZZ", where
```
cpennington: It seems like the title of the PR should be human-centric, and that maybe we should use GH tags for things like process, best practice, or architecture.

oep-template.rst
nasthagiri: **Guess what?**  Here's data about the number of words in the abstracts of the current 13 OEPs:

- **average**: 76 words
- **median**: 39
- **min**: 11 words
- **max**: 217

_I had happened to look at this as I began to write a new OEP today._

**What does this mean?**

1. Arbiters and reviewers have been satisfied with abstracts that have significantly fewer words than 200.
2. Even if there's a word-count requirement, people are using their own judgment.

**Given this,**
I'd rather keep OEP-1 as minimal as necessary.  And it doesn't seem like requiring or even suggesting 200 words for the abstract is necessary.  Different OEPs will tend to have different length of content and density of information in each of the sections.  I like how the current format helps people break it down into a consistent structure.  But then, we can leave it to the authors, arbiters, and reviewers how they choose to relay their information and decisions.

oep-template.rst
nasthagiri: We currently have this in 3 different places:
* **References** section in the header
* In this **Rationale** section
* In the **Rejected Alternatives** section below

I found it confusing and repetitive.  For now, I kept the References header and removed the rest.  Where do folks think it should belong?

I felt the References header was good for reading that information quickly at-a-glance.  If authors/arbiters want to capture further detail, they can always do so.

oep-template.rst
nasthagiri: See [above comment](https://github.com/edx/open-edx-proposals/pull/53/files#r166806793).

oeps/oep-0001.rst
nasthagiri: I see your point.  Do you think bp-proc-arch aren't a quick human-centric way to read the type of the OEP in the title (for developers)?

oep-template.rst
nasthagiri: Good catch.  I'll update the template.  I meant to add info to just the filenames since that is absent today.

oep-template.rst
nedbat: I appreciate the effort to reduce barriers to writing.  I think we should leave something here about links because it would be easy to forget to include them.  How about, "Links to the original discussion will be helpful, if you have them."

oeps/oep-0001.rst
cpennington: I'd much rather have real text in the title, and have tags with real words as well.

Something like "OEP-1: A Process for Enhancing OpenEdX", tagged "process"

nasthagiri: That sounds good.

oep-template.rst
nasthagiri: Sure.  But in which section of the document should discussion links go?

oep-template.rst
```diff
@@ -85,32 +77,20 @@ An OEP that introduces backward incompatibilities must describe the
...
-OEP submissions that do not consider backward compatibility issues are unlikely
-to be accepted.
-
-
 Reference Implementation
 ========================
 
 The reference implementation must be completed before any OEP is given "Final"
-status, but it need not be completed before the OEP is accepted. While there is
+status, but it need not be completed before the OEP is "Accepted". While there is
```
nedbat: Not to nit-pick, but I found `accepted` to read better than `"Accepted"`.

nasthagiri: Feel free to nitpick.  I had made the change to make it consistent with `"Final"` in the 1st phrase.  But I can easily revert the change - as readability was the goal for the consistency.

robrap:
Thanks for doing this @nasthagiri.  I haven't had a chance to read through yet, but one thought in response to your demo presentation, is that it would be great if the description of this PR (and all OEP PRs) started with the due date on which the decision will be made and comments must be in.

oep-template.rst
```diff
@@ -50,32 +54,20 @@ existing architecture or process is inadequate to address the problem that the
...
-OEP submissions without sufficiently clear and compelling motivation are
-unlikely to be accepted.
-
 Specification
 =============
 
 The specification describes the technical details of the Architecture, Best
-Practice, Process or Product Enhancement proposed by the OEP. If the proposal
-includes a new API, specify its syntax and semantics.
+Practice or Process proposed by the OEP. If the proposal includes a new API,
```
pdpinch: I understand and generally agree with your rational for removing Product Enhancements. I do wonder though how those should be brought up and resolved. Can we spare a few words in this document to say what is out of scope for OEPs?

nasthagiri: @pdpinch Yes. Good idea to say where one should go instead for Product Enhancements. I've reached out to the edX Product team on what their current process is. 
FYI  @marcotuts @sstack22

oep-template.rst
nasthagiri: Given my recent experience with [OEP-17](https://github.com/edx/open-edx-proposals/pull/54), the **Rationale** section seems to be a good fit for links to external discussions.  I'll update this PR to remove it from Rejected Alternatives, but keep it in Rationale.

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
 should be Accepted, and as such, the Arbiter should be knowledgeable about 
 the contents of the proposal, and willing to listen to arguments both for 
 and against it by the rest of the community.
 
-The Arbiter is also responsible for helping the Author move the proposal
+The Arbiter is also responsible for helping the Authors move the proposal
 through the OEP process, providing technical and process expertise as needed.
-The Arbiter also assists the Author to solicit feedback from the community on
+The Arbiter also assists the Authors to solicit feedback from the community on
 the OEP, and for helping to move the OEP towards a final decision (whether that
```
efischer19: Tiny grammatical nit, either subject/verb agreement here is weird or I'm parsing the compound predicate incorrectly.

- The Arbiter ... assists
- (The Arbiter) ... for helping to move

The simple fix is to replace `...and for helping to move...` with `...and helps to move...`. A potentially cleaner way to phrase the whole sentence would be `The Arbiter also assists the Authors in soliciting feedback from the community on the OEP and moving it towards a final decision (whether that decision is Accepted, Rejected, or Deferred).`

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
 
-After the Author drafts an OEP in a format in which they are comfortable, they
-will request an Arbiter from the edX Chief Architect. This Arbiter will be
-recorded in the "Arbiter" header on the OEP. The rest of Open edX community 
+Step 4. Request an Arbiter
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+After the Authors draft an OEP in a format in which they are comfortable, they
+will request an Arbiter from the `edX Architecture Team`_ by emailing 
+`arch-team@edx.org`_. This Arbiter will be
```
robrap: Maybe `edX Core Architecture Team` and `core-arch-team@edx.org` would better differentiate from the larger team of all architects?  Renaming proposals are imperfect, but you get the idea. :)

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 For an OEP to be accepted by the Arbiter, it must meet certain minimum
 criteria. It must be a clear and complete description of the proposed
-enhancement. The enhancement must represent a net improvement. The proposed
-implementation, if applicable, must pass the existing
-`Open edX Contribution Guidelines`_.
-
-.. _Open edX Contribution Guidelines: http://edx.readthedocs.org/projects/edx-developer-guide/en/latest/process/index.html
+enhancement. The enhancement must represent a net improvement.
```
robrap: "Enhancement" feels like a confusing term when it comes to best practices, especially when helping to codify existing, but partially followed, best practices.

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
-
 OEP Review & Resolution
 -----------------------
 
 Once an OEP has been accepted by an Arbiter, it is "Under Review". Once this
-state is acheived, we recommend announcing the OEP to the community in relevant
-forums such as the edx-code and more relevant mailing lists, the
-#open-edx-proposals Slack channel, and any other relevant Slack channels. It 
-should also be added to the list of "Under Review" OEPs in the OEP repo.
+state is acheived, we recommend announcing the OEP to the community in the
```
robrap: `acheived ` => `achieved`

oeps/oep-0001.rst
```diff
@@ -300,111 +368,84 @@ ReStructuredText [8] allows for rich markup that is relatively easy to read,
...
-
 .. _reStructuredText: http://docutils.sourceforge.net/rst.html
-.. _OEP template: https://github.com/cpennington/open-edx-proposals/blob/master/oep-template.rst
+.. _OEP template: https://github.com/edx/open-edx-proposals/blob/master/oep-template.rst
 
 OEP Header Preamble
 -------------------
 
-Each OEP must begin with an ReST table with metadata about the OEP. The rows
+Each OEP must begin with a ReST table with metadata about the OEP. The rows
```
robrap: I personally would prefer "an RST table".  I would read "ReST" or "reST" as "rest", which makes me think about the wrong thing.  Wikipedia claims all 3 are used: https://en.wikipedia.org/wiki/ReStructuredText

oeps/oep-0001.rst
```diff
@@ -300,111 +368,84 @@ ReStructuredText [8] allows for rich markup that is relatively easy to read,
...
+* The **Arbiter** field is used to record who has the authority to make the final
+  decision to approve or reject the OEP.
+
+* The **Type** header specifies the type of OEP: Architecture, Best Practice, or
+  Process.
+
+* The **Created** header records the date that the pull request for the OEP was
+  opened. It should be in YYYY-MM-DD format, e.g. 2016-04-21.
+
+* The **Review Period** header specifes the target dates for reviewing the OEP, as
```
robrap: Would it be helpful to have a PR template (automated or otherwise) that includes this info at the top of the description?

oeps/oep-0001.rst
```diff
@@ -414,3 +455,35 @@ Change History
 
 * Add a definition of the *Change History* section.
 * Add a copyright notice.
+
+2016-10-11
+----------
+
+* Add a new "Product Enhancement" proposal type
+* Remove references to arch@ email address.
+* Create "Inital Submission" section.
```
robrap: `Inital` => `Initial`

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
-
-Initial Submission
-------------------
-
-Once the Author has asked the Open edX community whether an idea has any chance
+the potential authors time. Asking the Open edX community first if an idea is 
+original and appropriate helps prevent wasted effort. It also helps to make
+sure the idea is applicable to the entire community and not just the authors.
+
+.. _open-edx-proposals Slack channel: https://openedx.slack.com/messages/C1L370YTZ/details/
```
clintonb: I like Slack but, between work and extracurricular activities, I have five distinct Slack accounts. They can get distracting. As someone who is less active in the community, I rely more on the mailing list to see broader changes. I agree that Slack is a great forum for the conversation; however, at least mentioning the new OEP and linking to the Slack conversation on the mailing list would help keep informed those of us who aren't always on Slack.

oeps/oep-0001.rst
```diff
@@ -414,3 +455,35 @@ Change History
 
 * Add a definition of the *Change History* section.
 * Add a copyright notice.
+
```
robrap: Nice to have: Insert TOC directive that lists dates.

oeps/oep-0001.rst
clintonb: P.S. Coincidentally, the mailing list is how I learned of this request for comments. 😀

nasthagiri:
Thanks for all the feedback so far.  It is great to have a wider audience review these changes.  I plan to update the PR tomorrow (Feb 15th) based on all comments received.

Regarding slack vs edx-code, I'm thinking of updating the PR as follows:
* For `Step 2. Vet your idea` - keep this step minimal by posting on Slack.
  * We want to encourage folks to get over the hump of creating OEPs.  Trying to get a consensus in an email conversation prior to creating the OEP is an extra hurdle.
  * Further conversations can then be captured on the OEP PR itself.
  * The initial "vetting" will still happen, but just on slack, and should be sufficient. (Let's try it.)
* For announcing the OEP, let's continue to use the edx-code mailing list in addition to Slack.
  * Folks aren't always online on Slack and can easily miss announcements.
  * We can create a standard announcement template (with a standard subject line) so folks can filter for OEP announcements specifically if they choose.

johnbaldwin:
I just became aware of this PR late last night and am working on digesting the proposed changes and conversation. Is there a planned schedule for periodic review and updating the OEP proposal process? That is, if members of the community are not able to digest and provide feedback to this PR with a deadline of today, when is the next opportunity?

cpennington:
@johnbaldwin: We've had some late-breaking comments, so I'm going to leave the window for commenting open for a little bit. That said, if there are future comments you have, or have clarifications you think should be made, but this PR has been closed, please do open a new PR with the suggested changes and we can discuss them.

johnbaldwin:
@cpennington Thank you!

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
+The OEP process begins with a new idea for Open edX. Unless it is a Best
+Practice, a single OEP should contain a single key proposal or new idea. Small
+enhancements or patches often don't need an OEP and can be injected into the Open
+edX development workflow with a patch submission.
+
+Step 2. Vet your idea
+~~~~~~~~~~~~~~~~~~~~~
+
+OEP Authors should first attempt to ascertain whether the idea is appropriate 
+for an OEP. Posting to the `open-edx-proposals Slack channel`_ is the best way
```
johnbaldwin: Are there any persons or a group we should DM in the #open-edx-proposals slack channel when we post an idea?

johnbaldwin:
Regarding the license statement in oep-0001: https://github.com/edx/open-edx-proposals/blame/arch/simplify-oep1/oeps/oep-0001.rst#L320

Something that might help is to clarify if just the OEP is under creative commons or the body of work for which the OEP represents is also under creative commons or that the body of work for which the OEP represents takes the license of the target repositories for code PRs (like GNU v3.0 for edx-platform)

johnbaldwin:
Please correct me if I'm wrong, but I assume that this is not a shared repository for folks outside of edx.org. Do you think it would help to make clear to folks outside of edx.org that they need to follow the fork -> branch -> PR against this repo?

see: https://github.com/edx/open-edx-proposals/blame/arch/simplify-oep1/oeps/oep-0001.rst#L151-L152

oeps/oep-0001.rst
```diff
@@ -45,163 +61,211 @@ OEP Types
...
-sure the idea is applicable to the entire community and not just the author.
-Just because an idea sounds good to the author does not mean it will work for
-most Open edX installations.
-
-Initial Submission
-------------------
-
-Once the Author has asked the Open edX community whether an idea has any chance
+the potential authors time. Asking the Open edX community first if an idea is 
+original and appropriate helps prevent wasted effort. It also helps to make
```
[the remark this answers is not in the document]
nedbat: I think "original" is a misleading word.  The intent was, "has this idea been discussed before?"

oep-template.rst
nasthagiri: @pdpinch We, including @johnmark, are working to determine a good process for Product Enhancements (for us to use internally as well as with the open edX community).  Please stay tuned to provide input on this.

oep-template.rst
nasthagiri: reverted.

oep-template.rst
nasthagiri: done.

oeps/oep-0001.rst
nasthagiri: fixed.

oeps/oep-0001.rst
nasthagiri: Good idea.  We created an @oep-team slack group for this - a group who can help with vetting an idea.  I'll update the doc.

oeps/oep-0001.rst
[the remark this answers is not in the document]
nasthagiri: Updated by replacing "original" with "previously discussed".  Also, changed "should" to "may choose to".

oeps/oep-0001.rst
nasthagiri: Good feedback.  I've updated the doc to include Slack as one of the options for Vetting the OEP, but keeping the edx-code mailing list as a place to announce the OEP.

oeps/oep-0001.rst
nasthagiri: fixed.

oeps/oep-0001.rst
nasthagiri: updated.

oeps/oep-0001.rst
nasthagiri: For now, we've decided to keep this as is.  We may change it in the future as needed.

oeps/oep-0001.rst
nasthagiri: fixed.

PR #54, opened 2018-02-13
OEP-17: Feature Toggles

nasthagiri:
**Review Period**
February 14, 2018 -> March 19, 2018

**Arbiter**
@efischer19 

**Motivation**
A feature toggle is a software development technique that decouples deployment of code from release (enablement) of the code.  Given the many benefits of using feature toggles, edX development teams have been using them since the very start. Aligning on a common best practice for feature toggling is critical for both the short-term and long-term health of the system. Using different standards, strategies, testing procedures, etc leads to confusion, production failures,
and long-term maintenance issues.


efischer19:
This looks good on my initial readthrough, let's open it up to the rest of arch-review to get their thoughts.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+load or attack.
+
+Typically, long-lived Ops toggles are useful for gating non-critical features that
+are very expensive on system resources. However, the long-term costs of maintaining
+the added complexity in the code should be measured against the benefits of
+operationally degrading the service when needed.
+
+Use Case 5: Beta Testing
+~~~~~~~~~~~~~~~~~~~~~~~~
+.. _Beta Testing: `Use Case 5: Beta Testing`_
```
feanil: How does this differ from an A/B test?  Isn't this an A/B test where A is the current system and B is the new thing we thing we're trying out?  Is it that our current A/B framework doesn't allow for targetting in the ways you mentionned below or some other reason?

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+.. _Mean Time to Recovery (MTTR): https://www.thoughtworks.com/radar/techniques/focus-on-mean-time-to-recovery
+.. _canary release: https://martinfowler.com/bliki/CanaryRelease.html
+
+Use Case 4: Ops - Graceful Degradation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. _Ops - Graceful Degradation: `Use Case 4: Ops - Graceful Degradation`_
+
+In certain cases, the development team (in consultation with the operations team)
+may choose to extend the lifetime of an Ops toggle in the codebase even after
+releasing its gated feature. A small number of such long-lived Ops "kill switches"
```
feanil: Do we want any best practice around naming conventions for this or other long-lived switches so they are easier to reason about?

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+|   | - Rollout       |                                     |   Flag_ class supports the "Rollout" |
+|   | - Group         |                                     |   toggle type. However, since edX    |
+|   |                 |                                     |   (currently) does not store course  |
+|   |                 |                                     |   and organization relationships as  |
+|   |                 |                                     |   `Django groups`_, the framework    |
+|   |                 |                                     |   must provide support for the       |
+|   |                 |                                     |   "Group" toggle type.               |
++---+-----------------+-------------------------------------+--------------------------------------+
+| 8 | Non-collision   | Feature toggles created by          | **No.**                              |
+|   |                 | independent teams should not collide|   The framework must support         |
```
robrap: You use the following text below, which I found added some clarity: `The namespace should be unique
+to each Django app so it doesn't collide with other installed apps in the system.`

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+warranted to use them in a large-scale system.
+
+Motivation
+==========
+
+Given the many benefits of using feature toggles, edX development teams have
+been using them since the very start. Additionally, since deployment latency
+for edx.org has reduced from monthly->weekly->daily and since teams have been
+bitten too many times by `long-term feature branches`_, teams now have a greater 
+incentive to develop incrementally within the main branch. Updating veteran
```
dsjen: I'm a bit confused by the use of "veteran" here. Is it referring to legacy code/UX?

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+to a *specific subset* of the population before releasing to the rest. This is in
+contrast to the `Ops - Monitored Rollout`_ case where changes are rolled out to a
+*random subset* of users.
+
+In the edX case, the Beta testing program may include the following types of population
+subsets:
+
+* Users - list of specific users
+
+* Courses - users enrolled in any course within a list of specific courses *(for
```
dsjen: I'd also add that course instructors (e.g. users of studio and not necessarily "enrolled" in a course) can also have functionality gated through feature flags. I'm not sure if this bullet point should be augmented with that or an additional bullet should be added to cover this case.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+There are sometimes business requirements for keeping long-term feature toggles in
+order to expose or limit certain features to certain groups.
+
+Use Case 6: VIP / White Label
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. _`VIP / White Label`: `Use Case 6: VIP / White Label`_
+
+The business may choose to modify the product experience for different classes of
+users. For example, the state of a feature toggle may depend on whether the user
+is a paying customer, a bundled feature within a subscription plan, or applicable
```
georgebabey: This sounds like entitlements to me. In light of not having a robust entitlements system baked into our platform, I can see why someone would try to leverage toggles as a short term fix but l would caution against using toggles for this and not sure it belongs in an OEP.

Ideally an entitlement system would be tied back to reporting systems, financial systems, etc. and I have concern if we lean too heavy on long-term feature toggles it may make building a robust system more difficult.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+  and you may end up deploying bugs to your production environment.”
+
+.. _James McKay puts it: https://jamesmckay.net/2011/07/why-does-martin-fowler-not-understand-feature-branches/
+
+Testing Best Practices
+~~~~~~~~~~~~~~~~~~~~~~
+
+Given that, here are best practices for testing a Feature Toggle:
+
+* Tests should run with whatever state is in production.
```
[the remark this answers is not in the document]
nasthagiri: Yes, I agree that production includes both Prod and Edge.  I've updated this line to state both.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+
+* Tests should run in both on and off Toggle states unless they are guaranteed
+  to not be enabled in production.
+
+  * Acceptance or end-to-end tests for Toggles that gate user-facing changes
```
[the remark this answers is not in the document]
nasthagiri: We do plan to create a separate OEP for testing where we will describe the testing pyramid and best practices.  I imagine we may clarify more of this in that OEP.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+.. _Devstack: https://github.com/edx/devstack
+.. _WAFFLE_OVERRIDE: https://waffle.readthedocs.io/en/v0.9/usage.html#waffle-override
+
+Test Plans for Toggle Use Cases
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following diagram summarizes test plans for the various toggle use cases while
+taking best practices into consideration.
+
+.. image:: oep-0017/test_plan_for_use_cases.png
```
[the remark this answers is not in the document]
nasthagiri: 1. Yes, I am thinking of them as 3 different things.  I can clarify this further in the doc once we've come to a common understanding.
2. Not sure what this means.
3. The type of tests (unit vs acceptance vs end-to-end vs contract-based tests) should be explained in a separate Testing OEP.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+   :alt: A diagram that shows the test plans for the various toggle use cases.
```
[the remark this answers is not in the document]
nasthagiri: Yes, exactly.  If needed, I'll schedule a design review meeting for this to make sure we're on the same page and adjust the diagram and document accordingly.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+
+In addition to using the Report as a central tool of overseeing the toggles,
+individual teams should create tickets in their backlogs for removing toggles
+according to their intended expiration dates.
+
+Rationale
+=========
+
+Although feature toggles have been in use from the very early stages of development
+on the platform, the `Feature Flags and Settings on edx-platform`_ wiki was one
```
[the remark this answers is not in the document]
nasthagiri: Regarding documentation location:

- This OEP is the place for best practices on feature toggles.
- We will have an OEP on best practices on tests.
- There will be some overlap and cross-references between the Testing OEP and this one.
- Implementation details about the waffle_utils API should live within the waffle_utils folder.

Regarding the additional details, as I said above, it would be best if we remove the option to have a "default" value for toggles.  I believe it will add more confusion and bugs, which outweigh its convenience.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+.. _long-term feature branches: https://blog.newrelic.com/2012/11/14/long-running-branches-considered-harmful/
+
+Use Cases
+---------
+
+At its core, feature toggles allow teams to deploy alternative code paths and to
+choose between them at runtime. There are multiple scenarios where this capability
+comes in handy.
+
+**Note:** Using feature toggles for experimentation is out of the scope for this
```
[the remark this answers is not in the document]
robrap: We did use it and I can imagine using it again.  Not sure if you want it to be a separate use case, or it could be an exception to the "Incremental Release" use case where we are more ok with it because we plan to quickly remove it?  Or, do you think something else is what we should have done?

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+The following diagram summarizes the various use cases along 2 axes: feature
+maturity and longevity. Feature maturity corresponds to the level of certainty
+that the team has about the feature, including unexpected side-effects such
+as performance and user-behavior regressions. Longevity depicts the lifetime
+of the feature toggle and how long-lived it is expected to be.
+
+The diagram also labels which use cases are primarily driven by engineering
+teams (E) and/or business product teams (B).
+
+.. image:: oep-0017/feature_maturity_longevity.png
```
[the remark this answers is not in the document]
robrap: Not sure if you want to (or did?) add any of these notes to the doc, but it helped me understand why some of these points landed where they did in the image.  Thanks!

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+  * supports the "Switch" toggle type
+  * though recommend using WaffleFlag_ instead since it has more support
+    for testing
+
+* WaffleFlag_ class
+
+  * supports the "Switch" toggle type
+  * supports the "Rollout" toggle type for `Ops - Monitored Rollout`_ with support
+    for percentage rollouts.
+  * supports the "Group" toggle type with `Beta Testing`_ for certain users.
```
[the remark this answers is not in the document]
robrap: This was less about using Feature toggles, but whether or not a "Group" toggle type should be used in this case.  For examples, using Waffle Switches is fine.  This had come up before, and @jcdyer mentioned he ran into complexity issues when trying to use a Waffle Flag in this way (i.e., with a fake request).  I'm not sure what those were.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+| Toggle Type         | One of Switch, Rollout, or Group to    | In code, by developer  |
+|                     | further clarify the toggle's usage.    |                        |
+|                     |                                        |                        |
++---------------------+----------------------------------------+------------------------+
+| Created in Code Date| Required field to specify the date the | In code, by developer  |
+|                     | toggle was added to the codebase; to   |                        |
+|                     | easily find all stale toggles.         |                        |
++---------------------+----------------------------------------+------------------------+
+| Expiration Date     | Optional field to specify target date  | In code, by developer  |
+|                     | of removal; to easily find all expired |                        |
```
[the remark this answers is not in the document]
robrap: Agreed that it should be more general then, but whether it is a link to a ticket or a link to a page with details/action items, etc., it feels like someone should be able to easily find/access the link that provides this further history/status/details that don't below in the code, except as a link (in my opinion).

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,809 @@
...
+|                     | toggles.                               |                        |
++---------------------+----------------------------------------+------------------------+
+| First Modified Time | Date the toggle was first set in the   | Derived from relational|
+|                     | system; to get the starting date       | tables                 |
+|                     | of its use.                            |                        |
++---------------------+----------------------------------------+------------------------+
+| Last Modified Time  | Date the toggle was last set/unset in  | Derived from relational|
+|                     | the system; to easily find all unused  | tables                 |
+|                     | toggles.                               |                        |
++---------------------+----------------------------------------+------------------------+
```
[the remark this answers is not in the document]
robrap: Thanks.
1. If one report per environment, I think it would be good for us to document somewhere what reports need to be reviewed (i.e. Prod, Edge).  Maybe that will always just be these two?  We had questions about other environments, and learned that they are managed through Open edX release process, so we just had to updated appropriate docs in those cases.
2. I can explain our use case for default value, and you can let me know if you have a better idea.
3. I'll explain soon the different values and you can let me know where this will appear.
Thanks!

oeps/oep-0017-feature-toggles.rst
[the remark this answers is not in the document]
robrap: I'm fine with removing the "default" value, but only if we have an alternate suggestion that satisfies the requirements I stated above.  Maybe it is through a migration that adds the flag?  I can't remember why we went down the path we did, but in all cases, I think we should help people understand the path and what to do.

oeps/oep-0017-feature-toggles.rst
robrap: I am suggesting that this be further clarified, because I was confused until I got to the statement that I added above.  Thanks.

oeps/oep-0017-feature-toggles.rst
nasthagiri: Agreed.  In the report section below, I proposed having developers specify "Current Use Case(s)" and "All Use Case(s)" for each of their toggles.  This should indicate whether the toggle is long-term or short-term.  I suggest keeping it decoupled from the name itself - since the use case of a toggle tends to evolve as it is used.

johnbaldwin:
@nasthagiri I've read the OEP and through all the comments, I believe. So my apologies if I'm repeating something already stated.

I have one concern with this OEP. If I understand correctly, we're trying to move in the direction of breaking up the monolithic platform into more of a plugin architecture and pulling things apart into services, libraries, and plugins, with clear interfaces between them.

I fear that this OEP encourages a stronger monolithic pattern by encouraging feature toggles, and with that a tighter coupling of the different parts of the system that build a toggle driven feature, which will make it harder to break up the platform.

nasthagiri:
@johnbaldwin Yes, excellent point.  I added a paragraph to explain this trade-off and strongly encouraged developers to use SOLID design principles when considering long-term feature toggles.

The reason we still need this OEP is that we have many valid reasons for short-term toggles in order to gradually update our system without causing any major issues or downtime.  Long-term toggles are mostly needed to support our varying user base, including open edX.  Hopefully the use cases section explains this in further detail.

nasthagiri:
On March 2nd, we had a real-time discussion regarding this OEP.  Main action items are:
```
* A/B testing
	* Expand on why it is out of scope
* Use cases
	* Clarify that can have multiple at different phases of the feature
* Requirements
	* Include Distributed System (other than waffle)
* Drupal / Ruby apps - outside of this OEP
* Testing image
	* Merge Prod/Edge
	* Simplify on or off, on [+ off], etc
	* Break it up with description of each in the doc
```

Detailed notes captured below:
```
	-Drupal experimentation, re: https://github.com/edx/open-edx-proposals/pull/54/files#r171898753
		-Drupal is temporary/a snowflake, let’s leave it out for now.
	-Tools change, principles of feature flagging still hold.
	-”what if we expose an endpoint for ‘is this feature on’”
		-feels like a big redefinition for this OEP
		-more of a list “here are the tools we use for python. Any new tool (for Ruby, Drupal, etc) should do things X, Y, and Z to be considered fully featured.”
			-this could/will be addressed in a second round update.
	-this (OEP-17) is both a “best practices” and “current technologies” list
	-A/B testing
		-mostly out of scope for this document
		-other tools may be/are better suited for evaluating data after usage - connect affected users to downstream data, etc.
		-”An A/B test is a feature toggle with associated data collection”
		-Optimizely and waffle can/do overlap a bit
		-we should at least mention “waffle can do the same thing as optimizely, kind of. We chose not to use it in that way b/c it doesn’t do A/B testing *well*”
			-where “the same thing as optimizely” = “frontend A/B experimentation with associated data gathering”
	-What about optimizely doing backend-style flags?
		-fits in under “incremental release” in the diagram, kind of
		-known to be short-lived feature specifically for experiments
		-maybe we should highlight “experiments don’t fit into this OEP super well, you may cut across several of the above use-cases while doing that. See OEP-TBD for more”
	-Testing
		-make the diagram bigger (+10%, at least)
		-could clarify/explain some terms a bit further; this is a little “inside baseball” as it currently stands.
		-add a post-launch end state image with everything on
	-we should document the internal facts of life, such as “edge is a production environment”
		-unsure where this should live, but it’ll be a mapping from “general community guidelines” to “guidelines implemented specifically for edx.org”
	-defaults:
		-problem: a flag goes out, gets turned on, and can’t be removed. How do we propagate that “on” value back to devstacks, tests, etc. It is effectively now the default.
		-next time this comes up, let’s try adding a migration that inserts a default value for the toggle into the db during the time we want to keep the toggle around.
			-there might be better practices than this, but we can iterate on it
			-We can have a “waffle.py” inside each app that lists these “defaults” with a paper trail. This could be better, prevents huge # of migrations.
	-distributed systems:
		-hard problem, will likely require a centralized “flag configuration” to exist
		-we should note this (distributed system support) is the requirements section of the OEP
	-follow-up: not clear how it’s possible that we’re using toggle already, but not with a tech that hits all the requirements.
```

oeps/oep-0017-feature-toggles.rst
nasthagiri: Instead of duplicating the statement, I added a note above the table to indicate there's further detail that follows the table.

nasthagiri:
I have now updated this PR to address requested changes and feedback.

@efischer19 How many days from now do you suggest to allow for any final comments?

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,927 @@
...
+evaluated by measuring its impact on users' aggregate behavior.
+
+This is a deeper topic that is worth exploring in a separate OEP (see `Optimizely
+Tips and Tricks`_). For now, suffice it to say that edX uses an external A/B testing 
+platform (Optimizely_) to serve this purpose. Among other things, Optimizely supports
+user segmentation and targeting, data aggregation capabilities, statistical tools,
+and toggled code customizations. At this time, Optimizely is used for edX 
+experimentation, customizing edX code, but without merging any changes to the edX
+codebase. This also contrasts with the use cases that are in scope of this OEP. 
+
```
robrap: I'd embed some sort of clarification like the following:
```
*Note:* An experiment toggle is a very specific toggle type. This section does not imply that the other uses cases may still come in handy while working on an experiment.
```

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,927 @@
...
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Use Case
+     - Test Plan
+   * - `Incremental Release`_
+        Toggle is disabled in all environments, but tested in both states on master.
+     - .. image:: oep-0017/test_release.png
+   * - `Beta Testing`_, `Ops - Monitored Rollout`_
+        Toggle is enabled for some and disabled for others, so should be testing in both states on both master and stage.
```
robrap: `testing` => `tested`

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,927 @@
...
+  ...
+  Your features may not be as isolated from each other as you thought they were, 
+  and you may end up deploying bugs to your production environment.”
+
+.. _James McKay puts it: https://jamesmckay.net/2011/07/why-does-martin-fowler-not-understand-feature-branches/
+
+Testing Best Practices
+~~~~~~~~~~~~~~~~~~~~~~
+
+Given that, here are best practices for testing a Feature Toggle:
```
robrap: Either in this section and/or the section about waffle_utils, you may want to note that it includes testing utilities.

oeps/oep-0017-feature-toggles.rst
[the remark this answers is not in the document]
robrap: Sorry if I missed it @nasthagiri, but I didn't see an update related to the conversation about when and how people need to deal with this situation. We had discussed a preference of migrations over the default functionality (which may need to be deprecated). We also discussed @jmbowman's idea of managing these, but I don't remember the details.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,927 @@
...
+so they can choose the right toggle type(s).
+
+.. contents::
+   :local:
+   :depth: 1
+
+Decision Map
+------------
+
+In order to decide the required toggle type and its required duration for a
```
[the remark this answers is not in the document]
nasthagiri: Yes. Thanks.

oeps/oep-0017-feature-toggles.rst
nasthagiri: done

oeps/oep-0017-feature-toggles.rst
nasthagiri: fixed.

oeps/oep-0017-feature-toggles.rst
nasthagiri: I consider this to be implementation-specific and can go in a good README file within waffle_utils.  Right now, this OEP links to [its __init__ file](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/waffle_utils/__init__.py), which mentions testing utilities, but should be expanded further.

oeps/oep-0017-feature-toggles.rst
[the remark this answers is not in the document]
nasthagiri: I attached @efischer19's notes, which includes possibilities that were discussed.  However, we decided to wait until the next time this comes up before we recommend a best practice for it - as it needs to be further fleshed out - if it comes up again.
From my own multiple experiences with using waffle, we haven't had the need for specifying a default or for running a migration.  The need may be there for a particular Use Case that I haven't run into myself.

efischer19:
> @efischer19 How many days from now do you suggest to allow for any final comments?

@nasthagiri I propose we formally close this out at noon (EDT) on Monday, 19 March. There's been a good amount of discussion, ideally any comments will come in before EOD Friday.

oeps/oep-0017-feature-toggles.rst
```diff
@@ -0,0 +1,927 @@
...
+     - One of Switch, Rollout, or Group to further clarify the toggle's usage.
+     - In code, by developer
+   * - Created in Code Date
+     - Required field to specify the date the toggle was added to the codebase; to easily find all stale toggles.
+     - In code, by developer
+   * - Expiration Date
+     - Optional field to specify target date of removal; to easily find all expired toggles.
+     - In code, by developer
+   * - Current Setting(s)
+     - Summary of the current configuration and value of the feature toggle; to easily evaluate its readiness to transition or retire.
```
robrap: @nasthagiri: The following is not a required change.  Just a note that the reporting tool should report on the total list of coded toggles + the total list of configured toggles, so we can report orphaned configurations, where a value was found, but we have no known corresponding toggle declaration.

We did some flag renaming in the past and forgot to do clean-up of old values.  Also, it is possible that someone could remove a toggle from code and forgot to later remove the config from admin for all environments.

PR #55, opened 2018-03-03
Add test/examples for a feature toggle using multiple use cases.

robrap:
@nasthagiri: Here's a take at some of the enhancements. 

I added a separate commit for my suggested enhancement to the report.  I originally tried to add it to the Current Use Case(s) rows as you suggested, but I realized that it had a different source, and made that row too confusing.  The Current User Case(s) row is manually updated, possibly via admin.  This is not true for what I wanted to add.  

If these two rows remain separate, the report could ultimately provide a warning if you are configuring the feature toggle as opt-out, but don't include opt-out in the Current User Cases(s).  However, this is an enhanced feature that can just be built into the final tool if needed, and doesn't require specs in this OEP.

Let me know what you think.

### DIFF
--- a/oeps/oep-0017-feature-toggles.rst
+++ b/oeps/oep-0017-feature-toggles.rst
@@ -567,6 +567,11 @@ present the following information for each toggle.
 |                     | of removal; to easily find all expired |                        |
 |                     | toggles.                               |                        |
 +---------------------+----------------------------------------+------------------------+
+| Current Setting(s)  | Summary of the current configuration   | Derived from relational|
+|                     | and value of the feature toggle; to    | table                  |
+|                     | easily evaluate its readiness to       |                        |
+|                     | transition or retire.                  |                        |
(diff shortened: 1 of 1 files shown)


oeps/oep-0017-feature-toggles.rst
```diff
@@ -246,6 +247,14 @@ of the feature toggle and how long-lived it is expected to be.
 The diagram also labels which use cases are primarily driven by engineering
 teams (E) and/or business product teams (B).
 
+Note that a single feature toggle may transition through several use cases as a
+feature matures. An example feature toggle might first be used for beta testing a
+feature. While beta testing, the feature may mature from a beta test for
```
robrap: Note: The example could easily be shortened by removing: `While beta testing, the feature may mature from a beta test for individual users to a beta test for several courses.` I personally like that this highlights the complexity involved within a use case, but I can also see an argument for less text.

PR #60, opened 2018-05-02
OEP-0001: Simplify even further for OEP authors

nasthagiri:
This PR updates and simplifies the OEP process defined in OEP-1 even further than [PR-53](https://github.com/edx/open-edx-proposals/pull/53).

**Review Period**

May 02, 2018 -> May 11, 2018

**Arbiter**

@cpennington 

**Motivation**

Going forward, we want to encourage teams to write down and propose their architectural and system-wide decisions so we can align with best practices, processes, and system architectures.

Our hypothesis is that the rigid structure required by the PEP-based OEP template deters and confuses authors.  Also, we find that some authors continue to prefer other documentation and discussion tools.  So this PR introduces 2 other OEP templates to accommodate those needs:

- ADR-based template (much simpler format than PEP)
- External-link catered template (requiring distilled decisions to still be captured in the OEP, but discussions and other details can be captured elsewhere)

**Changes**

  * Reduce steps in the submission process
    * Remove the obvious "scope your idea" as an initial step. 
    * Remove "vet your idea" before creating a Draft.
    * Move "request an arbiter" as the 1st step, in place of vetting and scoping.
  * Support alternative simpler templates.
* Refactored description for OEP status and review.

### DIFF
--- a/adr-based-template.rst
+++ b/adr-based-template.rst
@@ -1,0 +1,61 @@
+======================
+OEP-XXXX: OEP Template
+======================
+
+.. This OEP template is based on Nygard's Architecture Decision Records.
+
++-----------------+--------------------------------------------------------+
+| OEP             | :doc:`OEP-XXXX </oeps/oep-XXXX-YYYY-ZZZZ>`             |
+|                 |                                                        |
+|                 | * <XXXX is the next available OEP number>              |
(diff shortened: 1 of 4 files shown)


nasthagiri:
@cpennington Please take a look through this PR - before I open it up for wider review.

PR #62, opened 2018-05-15
OEP-0022: Caching in Django

robrap:
OEP-0022: Caching in Django

This OEP provides some best practice regarding coding using Django caches. This includes some less proven direction as well, with the expectation that this OEP will be updated over time to keep up with recommendations and better clarify recommendations as share libraries become available.

FYI: @BenjiLee, @nasthagiri

### DIFF
--- a/oeps/oep-0022-bp-django-caches.rst
+++ b/oeps/oep-0022-bp-django-caches.rst
@@ -1,0 +1,248 @@
+===========================
+OEP-0022: Caching in Django
+===========================
+
++-----------------+--------------------------------------------------------+
+| OEP             | :doc:`OEP-0022 </oeps/oep-0022-bp-django-caches>`      |
++-----------------+--------------------------------------------------------+
+| Title           | Caching in Django                                      |
++-----------------+--------------------------------------------------------+
+| Last Modified   | 2018-05-21                                             |
++-----------------+--------------------------------------------------------+
+| Authors         | Robert Raposa <rraposa@edx.org>                        |
++-----------------+--------------------------------------------------------+
+| Arbiter         | Alex Dusenbery <adusenbery@edx.org>                    |
++-----------------+--------------------------------------------------------+
+| Status          | Under Review                                           |
++-----------------+--------------------------------------------------------+
+| Type            | Best Practice                                          |
++-----------------+--------------------------------------------------------+
+| Created         | 2018-05-04                                             |
++-----------------+--------------------------------------------------------+
+| `Review Period` | 2018-06-06                                             |
++-----------------+--------------------------------------------------------+
+
+Context
+-------
+
+We have a variety of tools for caching in Django in various repositories,
+but the solutions are not in a shared repository and don't meet all needs. A
+variety of issues were uncovered while working on caching fixes and
+improvements within one of our services.
+
+1. There is a common caching bug while using Django caches where someone
+   caches a value that is Falsey (e.g., None or []) and it is mistakenly
+   treated as a cache miss.
+
+2. In Django we commonly use the Memcached back-end. There are various
+   places in code where we read the same value many times within a
+   the scope of a request, and each of these is a network call to
+   Memcached.
+
+3. For testing and business purposes, it is sometimes useful to be able
+   to invalidate a cached value on a per-request basis.
+
+4. While refactoring caching code and reviewing potential libraries, there
+   seem to be potential best practices that would hide the boilerplate
+   caching logic and better highlight the business logic.
+
+5. There is potential for better standardizing how we handle cache keys.
+   For example, to generate keys that work with Memcached and avoid key
+   collisions in a global space like Memcached.
+
+Note: When and if caching is warranted in any particular circumstance is beyond
+the scope of this OEP.
+
+Recommendations
+---------------
+
+The following recommendations include patterns that are generally useful. The
+intention is to better codify these recommendations into shared libraries over
+time and update this documentation as the need arises.
+
+Common Caching Defect and Fix
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following is a common defect when caching using Django caches::
+
+    cached_value = cache.get(key)
+    if not cached_value:
+        # calculate value, set in cache, and return
+        # if the calculated value is Falsey (e.g., None or []), it will
+        # be treated as a cache miss above.
+    return cached_value
+
+The simplest fix for this defect is::
+
+    _CACHE_MISS = object()
+
+    cached_value = cache.get(key, _CACHE_MISS)
+    if cached_value is _CACHE_MISS:
+        # calculate value, set in cache, and return
+        # if the calculated value is Falsey (e.g., None or []), it will
+        # no longer be treated as a cache miss above.
+    return cached_value
+
+To solve this problem, use caching utility methods that avoid this defect. See
+`Implementations`_ for more details.
+
+Tiered Cache with Request Caching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use a library that allows tiered-caching calls that use a Request Cache backed
+by a slower Django cache (often Memcached). With a tiered cache, multiple
+calls for the same value within a request will not require additional network
+calls to Memcached.
+
+See `Implementations`_ for more details.
+
+Force Cache Refresh
+~~~~~~~~~~~~~~~~~~~
+
+When using any Django cache it is recommended that caching utilities offer a
+simple method by which to force a cache refresh for a given request. This is
+useful both for testing purposes, as well as for business purposes if a problem
+occurs in Production that requires a capability for refreshing a cached value.
+
+Note: This recommendation comes for free with a Request Cache, which is already
+refreshed at the start of every request.
+
+See `Implementations`_ for more details.
+
+Cache Decorators
+~~~~~~~~~~~~~~~~
+
+The following is not yet a strong recommendation. More work is required to
+assess and clarify.
+
+When caching, create cache methods where the boilerplate caching code is hidden
+behind a decorator. This practice allows the more important business logic to
+stand out.
+
+Note that in some cases, the boilerplate logic hidden behind the decorator
+includes opinionated code regarding how keys are defined and when and if the key
+will change. More assessment is required here to determine the recommended
+decorator(s) to use.
+
+Sample Before::
+
+    def get_and_cache_my_business_value(x):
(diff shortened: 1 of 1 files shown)


oeps/oep-0019-bp-django-caches.rst
```diff
@@ -0,0 +1,228 @@
...
+Tiered Cache with Request Caching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use a library that allows tiered-caching calls that use a Request Cache backed
+by Memcached. Multiple calls for the same value within a request will not
+require additional network calls to Memcached.
+
+See `Implementations`_ for more details.
+
+Force Cache Miss
```
iloveagent57: Can you give an example?  Do you really need to read the cache and get a miss if you know in advance that you want to refresh the cached value?

oeps/oep-0019-bp-django-caches.rst
```diff
@@ -0,0 +1,228 @@
...
+        # calculate my_business_value
+        ...
+        return my_business_value
+
+See `Implementations`_ for more details.
+
+Generating Cache Keys
+~~~~~~~~~~~~~~~~~~~~~
+
+Either use a cache decorator that generates the cache keys, or use a trusted
```
iloveagent57: Does it _have_ to be a decorator?  What's wrong with a private helper method like:
```python
def _cache_key(self, x):
    return 'my_module.my_class.my_function.{}'.format(x)
```

oeps/oep-0019-bp-django-caches.rst
```diff
@@ -0,0 +1,228 @@
...
+  creating a shared library of caching utilities.
+
+* There are common utilities for generating cache keys. These should be be
+  assessed when considering creating a shared library of caching utilities.
+
+**In ecommerce:**
+
+* The `RequestCache and TieredCache`_ both handle the `Common Caching Defect and
+  Fix`_. Although it is simple to refactor standard Django caching code to use
+  these classes, the interface to these classes is non-standard and they
```
iloveagent57: s/they//

oeps/oep-0019-bp-django-caches.rst
```diff
@@ -0,0 +1,228 @@
...
+* Use linting utilities to enforce the best practices.
+
+.. RequestCache and TieredCache: https://github.com/edx/ecommerce/blob/master/ecommerce/cache_utils/README.rst
+.. quickcache library: https://github.com/dimagi/quickcache
+
+Consequences
+------------
+
+As long as we keep all options open for developers, these best practices and
+supporting libraries should only help developers right cleaner and less buggy
```
iloveagent57: s/right/write/

oeps/oep-0019-bp-django-caches.rst
robrap: Will update.

oeps/oep-0019-bp-django-caches.rst
robrap: Will make this more clear.

oeps/oep-0019-bp-django-caches.rst
robrap: Thanks.

oeps/oep-0019-bp-django-caches.rst
robrap: The brain works (or malfunctions) in mysterious ways.

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+| `Review Period` | 2018-06-06                                             |
++-----------------+--------------------------------------------------------+
+
+Context
+-------
+
+We have a variety of tools for caching in Django in various repositories,
+but the solutions are not in a shared repository and don't meet all needs. A
+variety of issues were uncovered while working on caching fixes and
+improvements within one of our IDAs.
```
nasthagiri: "IDA" is not an industry standard term - you may want to expand it - or just say "microservice".

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+~~~~~~~~~~~~~~~~
+
+The following recommendation requires additional assessment.
+
+When caching, create cache methods where the boilerplate caching code is hidden
+behind a decorator. This practice allows the more important business logic to
+stand out.
+
+Note that the boilerplate logic hidden behind the decorator must include
+opinionated code regarding how keys are defined and when and if the key will
```
nasthagiri: It may not be necessary for the decorator to be opinionated about this. Rather, it can take a function as a parameter to compute the key.

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+
+This section will be updated as a shared library is selected and evolved.
+
+**In edx-platform:**
+
+* There is no utility for handling the `Common Caching Defect and Fix`_.
+
+* There is an `edx-platform RequestCache`_, but no Tiered-Cache.
+
+* There are no utilities for forcing cache misses.
```
nasthagiri: To force cache misses, we clear the cache before calling the method that uses and populates the cache.

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+  to use these classes, the interface to these classes is non-standard and
+  the simple refactor doesn't move toward using cache decorators.
+
+* There is both a `RequestCache and TieredCache`_.
+
+* The TieredCache provides functionality for forcing cache misses.
+
+* There is no cache decorator utility.
+
+* There are common utilities for generating cache keys. These should be be
```
nasthagiri: typo: "be be"

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+  enhanced to include the RequestCache, or to extend to use one.
+
+* There is functionality for skipping the cache. However, this functionality still
+  needs to be assessed.
+
+* The library is based on a slick @quickcache decorator. The decorator is
+  opinionated about its cache keys, tied to the code it is wrapping. This needs
+  to be assessed to see if it is a good fit.
+
+* The cache key generation is built into the decorator. Again, this needs to be
```
nasthagiri: How is this bullet different from the one above?

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+
+* Provides a TieredCache, but no RequestCache. The library would need to be
+  enhanced to include the RequestCache, or to extend to use one.
+
+* There is functionality for skipping the cache. However, this functionality still
+  needs to be assessed.
+
+* The library is based on a slick @quickcache decorator. The decorator is
+  opinionated about its cache keys, tied to the code it is wrapping. This needs
+  to be assessed to see if it is a good fit.
```
nasthagiri: Assessed or extended, if it doesn't meet our needs.  As mentioned above, a decorator could theoretically be extended by taking in a function that computes the cache-key.

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,229 @@
...
+.. quickcache library: https://github.com/dimagi/quickcache
+
+Consequences
+------------
+
+As long as we keep all options open for developers, these best practices and
+supporting libraries should only help developers write cleaner and less buggy
+caching code.
+
+The possible downside is that caching looks easier than it is, and developers
```
nasthagiri: Downside to what?  These best practices or to using a shared library?

oeps/oep-0022-bp-django-caches.rst
robrap: Thanks.  I'll go with "services" if that works for you.

oeps/oep-0022-bp-django-caches.rst
robrap: 1. I corrected the language here.
2. I added an intro to this subsection to clarify its status.
3. I renamed "Decisions" to "Recommendations", and added an intro to that section.
Note: I'm using past tense, but all changes are local for now.  Coming soon.

oeps/oep-0022-bp-django-caches.rst
robrap: You can read my update soon and let me know what you think.

oeps/oep-0022-bp-django-caches.rst
robrap: Updated.

robrap:
@nasthagiri: Thanks.  See my last commit for adjustments around references to Memcached.

oeps/oep-0022-bp-django-caches.rst
```diff
@@ -0,0 +1,248 @@
...
+    _CACHE_MISS = object()
+
+    cached_value = cache.get(key, _CACHE_MISS)
+    if cached_value is _CACHE_MISS:
+        # calculate value, set in cache, and return
+        # if the calculated value is Falsey (e.g., None or []), it will
+        # no longer be treated as a cache miss above.
+    return cached_value
+
+Use libraries that makes it simpler to avoid the above defect. See
```
nedbat: I don't understand this sentence (and it is slightly non-grammatical).

robrap: Is this more clear?:
```
Where possible, use caching utility methods that avoid this defect.
```

nedbat: I think what I was getting at was: there's a switch here from "Problem statement" to "Solution" that is sudden and implicit.  Maybe, "To solve this problem, use libraries that ...."

OmarIthawi:
I like the proposing a standard or even a "de-facto" method caching decorators. Thanks! 

I was going to ask for a `with` statement support, but [PEP 377 was rejected](https://www.python.org/dev/peps/pep-0377/) so :man_shrugging:

OmarIthawi:
@robrap would this OEP be a good place to specify some convention for name-spacing caches? No. 5 mentions the collision issue, but there's nothing else about it.

robrap:
@OmarIthawi: It is likely this would be the place, but I’d be more clear if you stayed more completely what you had in mind. If you wrote it up we can decide to include it, or maybe we’ll merge and you can work on an update PR. These docs should evolve over time.

robrap:
Thanks for your comments @OmarIthawi.  When we get better utilities that we want others to use (that may capture some of your thoughts), we can update the docs to point to those utilities.  Thanks again.

PR #64, opened 2018-06-12
OEP-0023: Style Customization (Draft)

arizzitano:


oeps/oep-0023-theming.rst
```diff
@@ -0,0 +1,204 @@
...
+
+4. The core contributor team will provide official documentation,
+      examples, and tools to help other core contributors and open
+      source community members develop themes.
+
+5. Further modification to UI elements or content will not be officially
+      supported within the scope of this theming system. A system to
+      enable content modification may eventually be developed and
+      supported, but its definition is outside the scope of this
+      proposal.
```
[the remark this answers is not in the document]
mtyaka: Content modification is a very common requirement, so I agree with Braden it should be clear how that's going to be supported before the frontend starts being rewritten in React. While content modification with current comprehensive theming system is often kind of messy, modifying a React based interface is going to be even worse if it's not clearly thought out.

oeps/oep-0023-theming.rst
```diff
@@ -0,0 +1,204 @@
...
+      announce official support for the new system.
+
+8. Removal of legacy theming-related code may begin two named releases
+      after the announcement of its deprecation.
+
+Consequences
+============
+
+-  All existing theming systems will be deprecated and support will
+      eventually be removed from the platform.
```
mtyaka: I'm in favour of removing obsolete systems, but we need to have way to do content modification first.

oeps/oep-0023-theming.rst
```diff
@@ -0,0 +1,204 @@
...
+
+-  Theming scope will be restricted to look-and-feel and branding
+      element modification. Content modification must take place outside
+      the scope of theming.
+
+-  The Open edX platform will gain further ability to evolve its
+      frontend architecture.
+
+-  The Open edX platform’s static asset size and build times will be
+      greatly reduced.
```
mtyaka: Would this enable real-time theming (via https://github.com/medialize/sass.js) on near real-time (via background tasks)? It would be great if you could set some color values in the admin, click a button and immediately see the changes.

grozdanowski: Great to see this kicking off! I agree we need a single system that would be properly documented and maintained, however using only Bootstrap theming (only visual part) is, in my opinion, not going to be sufficient for theming needs that vendors and maintainers of instances have. There are a lot of situations that require changing the content of pages, with one of the most obvious ones being setting the contents of static pages and changing the header and footer menu items (those are just the lightest modifications that come to mind, there are a lot more).

I don't see Comprehensive Theming as a system that doesn't cover a lot of needs - the template overriding system is something we definitely need and use. The problem lies more in usage of different types of templates (underscore) and lack of a proper API/documentation on how to include certain elements of the platform into custom templates. So doing customisations and implementing our custom templates is not a process where we refer to docs, but we rather figure out how everything works in the default templates and replicate it in our own. So when further work is done on the frontend by the core team and released in the next named release, a lot of stuff breaks because, naturally, the core team doesn't know what we use and what we don't (and I really wouldn't expect they do, after all we're essentially hacking stuff all around in our work).

What would certainly be helpful is a system where templates are clear and can be overridden (in a way we can override them using Comprehensive theming), where it's very clear and documented how to implement the LMS features in custom templates (for example: _You can call the list of 5 newest courses on the platform using X, passing these arguments and expecting this response_), a centralised place where default SASS variables can be overridden in the theme and documented classes that can be reused (I guess Bootstrap would answer this). A prerequisite for this would, I guess, be a cleanup of existing templates in the platform.

My question would be - what is the relation between this and the planned effort of complete separation of frontend and backend? Regarding how closely related to those future plans and the future architecture the new theming engine is, Is it worthwhile to scrap everything and build a new theming engine from scratch, or maybe just improve the Comprehensive theming engine as an intermediate solution before the full transition to a separated frontend?

I have also started a project to explore the possibilities of a future completely separated LMS (including the caveats, and "what do we need to make it work"): https://github.com/appsembler/react-lms
It is in a very early stage, however as you can see I have addressed theming in it in a very similar fashion to current Comprehensive theming - maybe/hopefully it helps in your future explorations for the platform.

oeps/oep-0023-theming.rst
```diff
@@ -0,0 +1,204 @@
...
+      release of the new theming system and announce the official
+      deprecation of all legacy theming systems.
+
+7. The Open edX core contributor team will continue to solicit
+      contributions and feedback on the theming system until the
+      following named release of the platform, by which point they will
+      announce official support for the new system.
+
+8. Removal of legacy theming-related code may begin two named releases
+      after the announcement of its deprecation.
```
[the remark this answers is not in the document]
arizzitano: @bradenmacdonald we would love nothing more than to deprecate existing systems ASAP -- plus, I think it would create some motivation to ramp up adoption of proposed new systems. What do you think about announcing legacy deprecation effective with the alpha release of the new style theming system, and then removing legacy theming code starting with the next sequential named release?

oeps/oep-0023-theming.rst
```diff
@@ -0,0 +1,204 @@
...
+Plan
+----
+
+1. API definition for the new theming system will be proposed, agreed
+      upon, and documented.
+
+2. The core contributor team will develop an example theme compliant
+      with the new system and use it to build support into the platform.
+      Existing work [8] on such a theme already exists in the
+      edx-bootstrap repository.
```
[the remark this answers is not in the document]
arizzitano: Yes! A Bootstrap theme currently exists for edx.org: https://github.com/edx/edx-bootstrap/tree/master/sass/edx

oeps/oep-0023-theming.rst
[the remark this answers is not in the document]
arizzitano: @bradenmacdonald I think this is an excellent approach and would be interested in seeing it grow into an OEP! The deprecation path is a broader conversation -- I'll leave a more general comment below.

oeps/oep-0023-theming.rst
arizzitano: Real-time theming comes up a lot, and while we agree it's ideal from an admin/maintainer experience standpoint, the technical complexity involved in enabling it for new *and* legacy systems is not a burden we currently have the resources to shoulder. CSS variables are promising, but would end support for IE11; CSS-in-JS is another (controversial!) option, but will not work with legacy non-JS UIs. A while back I even looked into building [an in-browser SASS compiler](https://github.com/arizzitano/theme-generator), but was unable to make much progress.

Like the current style theming system, this approach is SASS-based and will need to compile first; however, it involves a single scoped set of variables (typically in one file). We would certainly be open to implementing a real-time theming system in the future -- technical suggestions are welcome -- but I don't think it needs to be a blocker for approving Bootstrap adoption.

oeps/oep-0023-theming.rst
arizzitano: Definitely agree! My wording around some of this was unclear -- I'll leave a comment below, and clean up the language.

oeps/oep-0023-theming.rst
arizzitano: @grozdanowski excellent points about the utility of comp theming template overrides -- as others have raised this as well, I'm addressing below in a broader comment. As much as I'd love to scrap the legacy systems and concentrate on building the new system instead, internal resourcing constraints will mean this needs to roll out gradually, rather than an all-in-one rebuild.

I envisioned the new system with the primary intention of working with the newer (decoupled frontend) architecture pattern. Achieving compatibility with legacy code may involve some hacky one-off integrations, but since the goal is to eliminate legacy frontends anyway, supporting them is less of a higher-order concern.

Your React LMS has actually been a major rallying point for us internally -- especially considering the timeline within which you completed it! We will definitely check out your theming system as well; looking forward to gathering insight and inspiration from your approach.

oeps/oep-0023-theming.rst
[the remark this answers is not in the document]
bradenmacdonald: > What do you think about announcing legacy deprecation effective with the alpha release of the new style theming system, and then removing legacy theming code starting with the next sequential named release?

@arizzitano I think that would be great! But it's not me you'd need to convince... I think we'd need more input from the community to see if anyone else objects strongly to that plan.

oeps/oep-0023-theming.rst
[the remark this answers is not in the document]
bradenmacdonald: > I think this is an excellent approach and would be interested in seeing it grow into an OEP! 

@arizzitano Great! Do you know if you or anyone at edX is interested in taking that on? I might be able to find someone to do so but I don't have capacity to develop it further myself right at the moment.

CC @mtyaka in case you're interested :)

bradenmacdonald:
Thanks for clarifying, @arizzitano ! Sounds great, and we're excited to see this moving ahead.

oeps/oep-0023-theming.rst
mtyaka: @arizzitano Definitely not a blocker for Bootstrap adoption, but great to know that real-time theming has been under consideration! 

At OpenCraft we did [a short discovery on how a (near) real-time theming could work](https://docs.google.com/document/d/1WvjMZA2YNYKnM5acqT1rPOqCTo8JUDNsMumHL7baCis/edit?usp=sharing). It would be great if you could give it a quick look to confirm whether the proposed implementation makes sense.

nasthagiri:
After checking in with @bill-filler, the arbiter, we are merging this OEP while keeping it in **Draft** state, until the time comes for a team to build a reference implementation and a plan forward.

PR #65, opened 2018-06-12
WIP: OEP-0024: Rate limiting

BenjiLee:
WIP!

### DIFF
--- a/oeps/oep-0024-bp-rate-limiting.rst
+++ b/oeps/oep-0024-bp-rate-limiting.rst
@@ -1,0 +1,139 @@
+===========================
+OEP-0024: API Rate Limiting
+===========================
+
++-----------------+--------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


BenjiLee:
Will have time to work on this in 3 weeks or so (after our teams next sprint)

mikix:
I know this is WIP, but what about the idea that Python-based rate limiting is not super flexible (hard to change quickly, hard for OpenEdx installations to customize)?

I know that Cloudflare supports rate limiting, which I believe we use in some places. Maybe that is a better place to control this? Plus, that way we never hit Django at all.

Controlling at the edge layer is also where most WAP features happen anyway. Rate limiting seems in that same family. I believe we use Cloudflare on some services for DDoS prevention.

(A malicious actor can still find the backend server IP and attack that way. But devops has a TODO item to use an IP whitelist or similar for backend servers so that only Cloudflare can hit them.)

PR #73, opened 2018-07-12
OEP-0026: Real-time Events (xAPI/Caliper)

nasthagiri:
Open edX Proposal for Real-time Events, primarily for Adaptive Learning capabilities.

**Review Period**
November 29, 2018 -> December 20, 2018

**Arbiter**
@brianhw 

**Motivation**
To satisfy emerging use cases that require notifying external systems of LMS events in real-time in standardized formats, supporting real-time events is a natural evolution of Open edX's eventing and API capabilities and its impact on connecting users, organizations, and learning services.


oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,421 @@
...
+
+Initially, we will exclude the "name" field. However, if we find that xAPI JSON parsers
+assume this field always exists, then we can include the field but provide a non-PII
+value, such as a copy of the `Open edX User UUID`_.
+
+Open edX User UUID
+^^^^^^^^^^^^^^^^^^
+An *Open edX User UUID* identifier currently doesn't exist in the platform. It will need 
+to be added as part of implementing xAPI support. There would be a 1:1 mapping between
+a randomly generated UUID_ and a row in the Django User table in the Open edX LMS.
```
jmaupetit: I have performance-related concerns with this 1:1 relationship. Using a custom user model may not be also considered?

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,421 @@
...
+
+For the first iteration, we need the following permissions:
+
+* Course restriction - certain consumers can access events only in certain courses.
+
+In the future, we may need the following:
+
+* User restriction - certain consumers can access all events for certain users.
+* Site restriction - certain consumers are limited to accessing events of certain sites.
+* Activity type restriction - certain consumers can access only certain types of events.
```
jmaupetit: Do you think it must be a layer of the app? IMO it must be a separated component acting as a proxy to forward the right event to the right consumer with appropriate permissions.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,421 @@
...
+to configure the list and permissions of registered xAPI event consumers.
+
+Eventually, the registry of consumers may become a self-service portal where consumers
+may initiate their request for access. However, initially, this can be a simple interface,
+possibly implemented in Django Admin.
+
+.. _Open edX Event Tracking: https://github.com/edx/event-tracking
+.. _Django App Plugin: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/plugins/README.rst
+.. _xAPI Python Open Source Library: https://xapi.com/python-library/
+.. _consumer-driven contract testing: https://www.thoughtworks.com/radar/techniques/consumer-driven-contract-testing
```
jmaupetit: If there is no routing layer, the administration layer should be implemented in the proxy (aka routing component).

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,399 @@
...
+.. _ADL: https://adlnet.gov/
+.. _Experience API (xAPI): https://www.adlnet.gov/research/performance-tracking-analysis/experience-api/
+.. _widespread industry support and usage: https://xapi.com/adopters/
+.. _xAPI Architecture Overview: https://www.adlnet.gov/research/performance-tracking-analysis/experience-api/xapi-architecture-overview/
+
+Learning Record Store (LRS) - Future
+====================================
+
+xAPI includes specification for a Learning Record Store (LRS), which encapsulates a
+data store with APIs for storing and accessing xAPI data. While an LRS could be used
```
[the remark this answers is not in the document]
nasthagiri: haha, nice question.

I am using the convention described by [APA style](http://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html) where we sound out the acronym to determine whether or not it starts with a vowel sound.  So in this case, since saying "L" starts with the short "e" sound, I'm using "an".

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,399 @@
...
+for many different purposes, it can be used to train adaptive engines as shown in the
+diagram below.
+
+.. image:: oep-0026/adaptive_learning_lrs_basic.png
+   :alt: The diagram above is enhanced with a new LRS component that is sent events
```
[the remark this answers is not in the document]
nasthagiri: thanks. fixed.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,399 @@
...
+    (write requests) and accessed by adaptive engines (read requests) for training
```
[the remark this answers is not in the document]
nasthagiri: updated.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: We will need to worry about backward compatibility with large-scale Open edX deployments (like at edx.org) where a migration of a very large auth_user table is not feasible with our current MySQL technology offering.

Another short-term alternative - I've been thinking about - in order to quickly move forward in the first iteration - instead of investing time in this new UUID join table, we could use the `hash value of the LMS user_id` as the unique identifier.  This value would not leak information, but it will still tie us to the database structure.  However, if we ever did migrate to a new auth_user table, we could theoretically add the new UUID column at that time and populate it with the hash of the original id values.  This would allow us to dynamically compute this value without needing to make any database migrations.  What do you think?

@edx/arch-review Any thoughts on this?

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: To be clear, when I use the term "layer" here, I simply mean a separable logical part of the plugin.  I do not mean to imply a specific way of how it is implemented.  

Would it be clearer if I use the term "component" instead of "layer"?

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: See my question above - would it be clearer if I use the term "component" instead of "layer"?  I didn't mean to imply a specific implementation when using the term "layer".

oeps/oep-0026-arch-xapi-integration.rst
sampaccoud: Looks like a solution for outputting messages with user uuids but could be a limitation if you need to come back and query the user table with these IDs? Do you think this could be required or can we rule it out?

nasthagiri: It's a requirement to find an appropriate user identifier for xAPI events.
I didn't understand your first question.  If you need to do a reverse lookup from User UUID to User, you can do so via a SQL JOIN operation, right?

sampaccoud: My remark was regarding your alternative:  "we could use the hash value of the LMS user_id as the unique identifier."

I agree with @jmaupetit that adding a separate table for the UUID is not a viable solution. I'm not expert in mysql but I'm surprised that the size of a table prevents adding a new column?

nasthagiri: For our current edx.org deployment, migrating large SQL tables result in server downtime, which is not acceptable for our situation.

I will go ahead and update the docs to say that we'll send a hash of the LMS user_id - instead of creating a new User UUID value.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+     - Limitations
+   * - username
+     - Often contains PII, such as user's name or year of birth.
+   * - LMS user_id
+     - Information leakage of auth_user table with easily guessable values; tied down
+       to implementation of the housing database.
+   * - email address
+     - Is PII and modifiable by the user.
+   * - anonmyous user_id
+     - It is currently constructed by hasing the user's LMS user_id with the Django
```
jmaupetit: `s/hasing/hashing/`

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+   * - Identifier
+     - Limitations
+   * - username
+     - Often contains PII, such as user's name or year of birth.
+   * - LMS user_id
+     - Information leakage of auth_user table with easily guessable values; tied down
+       to implementation of the housing database.
+   * - email address
+     - Is PII and modifiable by the user.
+   * - anonmyous user_id
```
jmaupetit: `s/anonmyous/anonymous/`

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+
+Initially, we will exclude the "name" field. However, if we find that xAPI JSON parsers
+assume this field always exists, then we can include the field but provide a non-PII
+value, such as a copy of the `Anonymized User ID`_.
+
+Anonymized User ID
+^^^^^^^^^^^^^^^^^^
+An *Anonymized User ID* will be generated as a SHA-256 hash of the Open edX user_id, 
+which is the **id** value of the user's row in the `Django auth_user`_ table in the
+Open edX LMS.
```
jmaupetit: Can we mentioned here that this is a "temporary" solution due to technical limitations on the `auth_user` table migration on large Open edX instances? And that the long term viable solution would be to add a `uuid` field to this `auth_user` table?

jmaupetit:
Thanks for adding the list of events @nasthagiri :tada:

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: fixed. thanks.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: fixed. thanks.  (I'm looking for a spell checker now for VSCode.  I have one for PyCharm, but not one for VSCode, which I've now switched to using primarily.)

ormsbee:
Sorry I'm late to the party, but what were the performance concerns about having a new table with a UUID in it?

jmaupetit:
I might be wrong, but don't you think that this relationship could increase requests performance over the `auth_user` model?

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Yes, I can do so.
@edx/devops  Do you confirm that this (adding a new column to the auth_user table) is an issue for us on edx.org?  By issue, I mean, it's not a trivial change - and would require additional work such as load-testing, etc?  Or has any recent infrastructural changes made this type of migration an easier path for us?

ormsbee:
@jmaupetit: I don't think it will really make that much of a difference. But since we'd expect the user:UUID association to never change, if it ever does become an issue, we can throw caching at it pretty easily.

sampaccoud:
Adding a new column on a huge table should not require down time if it is done in several steps:

- adding the column as optional (requires lock but very fast)
- adding a default
- running a data migration on the existing records
- making the column required (requires lock but very fast)

ormsbee:
I haven't been a participant in a DB-related outage RCA in a while, but my understanding is that you can do in place modifications to the table like that in MySQL 5.6 as long as the tables were created in 5.6 (and not carried over from an earlier install). You suffer a much longer lock for tables from earlier versions, even for adding an optional field. I believe we've seen locks held for about 1 min per million rows or so.

@jibsheet or @feanil might have more recent edX operational perspective to offer.

ormsbee:
Migrations have been one of the leading cause of operational issues on edx.org. There are definitely ways to make MySQL add a column without downtime, but I think it's easier to just make a separate table and join it to `auth_user`.

AnnaCampus:
Hi, so sorry again for joining you in the middle of the discussion. I just understood from Roi today the great progress that was done here with the Spec. I'll review the whole spec later on but I want to join to @ormsbee concern about the addition of a column vs new table. From my previous experience addition of new column to existing table might by tricky and might need a more complex upgrade process

nasthagiri:
@AnnaCampus Thank you for chiming in.  I agree with you.  We recently ran into unforeseen migration issues when we tried to replace an out-of-box OAuth Application class to our own custom class.  Apparently, others on StackOverflow had faced similar issues. We eventually backed out of that approach and embraced an alternative data model change.  The StackOverflow articles referred to similar issues when replacing Django's user model with a custom one (in a live production environment). 

While I also agree with @sampaccoud and @jmaupetit on the merit of updating the user table, I believe there is a lot else to be done with this initiative.  For example, we need to enhance the platform to expose good URNs for users, courses, and enrollments. We may also need to update existing openedX events to include additional data that we're not already including.

Given this, the [current proposal of sending a hash of the LMS user-id](https://github.com/edx/open-edx-proposals/pull/73/files#diff-3a03d2dfc38553dfffe84a603a6b8391R139) should be sufficient for our immediate needs. (Apparently, hash-of-user-id is also the value that we send with our edX data packages.  @brianhw can correct me.)  So we can look into customizing the user table at a later point.  For now, let's focus our efforts on the very many other enhancements to our platform as required by this feature.

sampaccoud:
The concern with adding a new table is that we would make a product design decision based on operational issues.

It may make sense for edx.org (and a few others like us maybe), but it's hard to justify for the mass of all the other users of Open edX. If not impacting performances, it complicates the design, the code and will generate more technical debt.

So +1 for @nasthagiri's pragmatic approach.

ormsbee:
I guess as long as we're sure we don't need to do reverse-lookups to map the anonymous ID back to a real user, the hash works for me too. Even more so if we specifically do not _want_ the ability to do a reverse lookup, for privacy reasons.

@sampaccoud: FWIW, even from a code perspective, I would favor a separate model over tacking on an extra field to the user model. Anonymous user mapping, whether more generally or for xAPI in particular, is distinct enough where I think it belongs in its own space. User models can become a catch-all space for all kinds of random things, and I especially wouldn't want to tack it onto the standard Django contrib auth_user model, given it's owned by an entirely separate space. Granted, we could make a custom user model, point it to the same table via config, and do a little fudging to make that transition work smoothly -- but I think all that combines to more technical debt and confusion than just making a separate model and joining it. It's not as _compact_, but it's pretty simple.

jmaupetit:
@nasthagiri one last remark: in the perspective of a sha256 hash of the user ID "long-term temporary" solution (:trollface:), can we explicitly mention that this hash will be at least salted? or generated with `bcrypt` or alike?

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+  + `edx.grades.problem.submitted <http://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/course_team_event_types.html#edx-grades-problem-submitted>`_.
+      Whenever a learner submits any problem.
+  + `problem_check <http://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/student_event_types.html#problem-check>`_.
+       Whenever a learner's answer to a problem is checked.
+  + `showanswer <http://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/student_event_types.html#showanswer>`_.
+       Whenever a learner is shown the answer to a problem.
+  + `edx.problem.hint.demandhint_displayed <http://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/student_event_types.html#edx-problem-hint-demandhint-displayed>`_.
+       Whenever a learner requests a hint to a problem.
+
+- **Video events**
```
Colin-Fredericks: There are a few video-related events missing from the list. The ones I can think of right now are changing the transcript language, changing the video speed, changing the volume, opening/closing the captions or the transcript, downloading the videos, and downloading the transcripts. We might not need all of those for this iteration, but I think getting the language could be important.

bradenmacdonald:
Very nice proposal!

What about grade-producing XBlocks other than capa? For example, Drag and Drop, or even ORA2. Such XBlocks don't emit any of the "Problem interaction events" listed herein, but they do emit standard grade and/or completion events and could be considered of importance for adaptive learning.

Should the translator provide an xAPI event for generic XBlock grade events?

Colin-Fredericks:
@bradenmacdonald I feel like there needs to be some standard equivalent of the problem_check event and the show_answer event for every grade-returning xblock. Otherwise, new graded xblocks will end up needing to have their own unique events incorporated one-by-one into a translator or into every adaptive engine.

If there's no equivalent to the hint event or the submit event that seems like less of a big deal to me, but I also don't 100% understand why the submit and check events are separate.

nasthagiri:
@bradenmacdonald @Colin-Fredericks 

The `edx.grades.problem.submitted` event is fired for all scorable xBlocks - so that includes Drag and Drop, ORA, CAPA, etc.  Since it is fired by the grading infrastructure, it is generic and doesn't include any problem-specific data.  Does this address your questions?

The `problem_check` and `show_answer` events are legacy events that CAPA fires and we continue to support.  Those seem aligned to xAPI's "problem interaction events" and hence I thought about aligning them.

Note that [Caliper](http://www.imsglobal.org/activity/caliper) includes different profiles, including a Grading Profile and Assessment Profile, that may be better suitable for our events - but we will probably need to support both xAPI and Caliper in the long-run.

I agree that completion events will be super useful to adaptive engines.  I would very much like to add them.  I did not see them listed, however, in our docs.  We should create these events if we don't already have them.  I had listed `edx.problem.completed` and `edx.video.completed` in the [spreadsheet](https://docs.google.com/spreadsheets/d/1Qx-1NkpCHXkWh8AagwHD5vzyBCdRXQVof10Ent_EDms/edit?usp=sharing) as possible future things we will need.

Colin-Fredericks:
@bradenmacdonald @nasthagiri 

Thanks for the details. Having problem-specific data would be incredibly useful. One of the glaring gaps in our (HarvardX's) current adaptive approach is that we can't tell what answer students actually gave to a problem, which really limits us. For instance, a particular wrong answer might indicate that a learner is having a particular issue. If we had information about which answer was given, we could target that issue specifically. Without knowing _why_ a learner got something wrong, we're more in the dark and can't adapt as well.

As far as problem completion events, I just want to caution people that problems may end up firing multiple completion events, so don't treat a completion event as "we can stop collecting data from this problem now."

nasthagiri:
Yes, agree that knowing the learner's answer is very important.  In the spreadsheet, I included a `response` field in the `results` section of the `problem_check` (`answered`) event.  Currently, I have it as returning `submission[answer]`, assuming that's the best value from that event.

For other non-CAPA problems, we may need to enhance the `Scorable` interface to provide additional information in the `edx.grades.problem.submitted` event.  I'd rather not just blindly include the student-state data from CoursewareStudentModule as I don't view it as a public interface that we want to expose to adaptive engines.  Thoughts?

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+
+Integration with Adaptive Engines
+=================================
+
+As an iterative path to this future, the immediate goal is to enable integrations with
+external adaptive engines.  In the far future, we may also implement our own open-sourced
+adaptive engines.  But in the meantime, we would like to establish standard APIs
+that (1) adaptive engines can use to receive real-time events from the Open edX LMS and
+(2) the Open edX LMS can use to query adaptive engines on what to present to learners.
+This document focuses on #1 with the need for scalable and real-time updates to adaptive
```
roishillo: @nasthagiri Another goal is to analyze the student's activity and build models based on the data we collected. It is better to train the models based on xAPI because then these data can be collected from multiple sources and not just the edx tracking events.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+the Open edX LMS and registered adaptive engines. We have chosen xAPI since it has
+`widespread industry support and usage`_ for tracking and analyzing learning activity.
+Please see `xAPI Architecture Overview`_ for further information on xAPI.
+
+.. _ADL: https://adlnet.gov/
+.. _Experience API (xAPI): https://www.adlnet.gov/research/performance-tracking-analysis/experience-api/
+.. _widespread industry support and usage: https://xapi.com/adopters/
+.. _xAPI Architecture Overview: https://www.adlnet.gov/research/performance-tracking-analysis/experience-api/xapi-architecture-overview/
+
+Learning Record Store (LRS) - Future
```
roishillo: I think that the current development of xAPI needs to include (an) LRS. I think that when developing these structure and sends it to the engine it needs to keep somewhere (unless every engine will build his own LRS) and this data will be very valuable in the near future when more engines will arrive. (so they can train their model with your specific data sets). I don't think it is a lot of work to keep these sentences in a mongo collection for now (also we had some issues with the mongo collection - I hope to share more about that later) or to keep it in S3 (this is how we are implementing it today).

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+with a *homepage*-scoped identifier.  One of these is sent along with the Actor's "name".
+To be mindful of learner privacy, we will initially take a conservative approach
+and only send #3, with an Open edX anonymized unique identifier of the learner
+(`Anonymized User ID`_).
+
+In the future, if certain external systems require `Personally Identifiable Information
+(PII)`_, like the learner's email address or name, then those may be conditionally sent
+with appropriate permissions. Adaptive engines, however, do not need PII.
+
+Initially, we will exclude the "name" field. However, if we find that xAPI JSON parsers
```
roishillo: I think we don't need name - we took it off, but let us know.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+   * - anonymous user_id
+     - It is currently constructed by hashing the user's LMS user_id with the Django
+       server's *SECRET_KEY* value. This value will change when the *SECRET_KEY* is rotated.
+
+.. _Django auth_user: https://docs.djangoproject.com/en/2.0/topics/auth/default/#user-objects
+
+Example
+^^^^^^^
+
+Here is an example of an **Actor** JSON value that we would generate:
```
roishillo: When there is a guest user (we have some open activities that can be reached by guess) we are sending "null" as the user id. let me know if it works for you - or if you want to have another solution,

oeps/oep-0026-arch-xapi-integration.rst
sampaccoud: Yes! We are currently testing learning locker (https://github.com/LearningLocker/learninglocker) and planning to use it for our developments. This will allow to further evaluate it and see if it can scale.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+will create a pull request to recommend adding it to the `central github repository of
+xAPI Profiles`_. 
+
+Here is an example of a **Verb** JSON value that we would generate:
+
+::
+
+    "verb": {
+        "id": "http://adlnet.gov/expapi/verbs/answered",
+        "display": {
```
roishillo: I am not sure that we need it - one of the problems with xAPI events is that each event is 2KB... if there are ways to minimize it and still be in the standard, thats great.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+
+The **Context** field allows us to embed additional contextual information in each statement.
+See specifics in the `Open edX Events`_ section below since it varies by event type.
+
+Here is an example of a **Context** JSON value that we would generate:
+
+::
+
+    "context": {
+        "registration": "openedx.org/enrollments/enrollment-v1:<anonymized-enrollment-id>",
```
roishillo: We are using the registration to determine the student's sessions.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+        "contextActivities": {
+            “parent”: [{
```
roishillo: One of the challenges with the parent is the array. analyze array is compute intensive. We keep the most important data in the extension. we want to keep in the parent the actual parent of the object (the unit or the sequence) but it will be hard to extract that later.

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+similar design principles. Essentially, a common validation library can be
+collaboratively maintained by participating consumers, including consuming adaptive
+engines. This ensures an end-to-end integration that can be maintained going forward.
+
+**Note**: Although xAPI specifies a standardized format, it is a low-level transaction
+schema and relies on higher-level "profiles" applied on top of it. So the profiles
+for specific Activities, Verbs, Contexts, etc used by Open edX need to be contractually
+maintained.
+ 
+Router
```
roishillo: What queue are you going to use?
maybe the first consumer of the queue is the LRS.

oeps/oep-0026-arch-xapi-integration.rst
roishillo: Great - we used learning locker. The first version (in PHP) worked very slow and didn't support bulk insert of events. I think this version is better - did you try to have a load test on it?

sampaccoud: Not yet but you're right we should.

bryanlandia:
Hey, FYI,  a bit off-topic in terms of a PR review, but this seems to be where most xAPI discussion is happening, and I think this might be helpful to look at another style of implementation....

A customer of ours needed an xAPI solution ASAP so I started with [ADLNet's archived repo](https://github.com/adlnet-archive/edx-xapi-bridge) for converting tracking log events to xAPI events.  It's a separate python process that monitors the `tail` of the tracking log and queuing/publishing xAPI statements.  We have it running successfully in a test/staging environment with LearningLocker as an LRS, covering course enrollment/unenrollment, certificate-based course completion, problem answers/attempts, link/section/tab navigation, and most video events.  Our next phase of work will be toward making it production-ready.

Our fork is here https://github.com/appsembler/edx-xapi-bridge.  I've used the [Open edX Google Spreadsheet](https://docs.google.com/spreadsheets/d/1Qx-1NkpCHXkWh8AagwHD5vzyBCdRXQVof10Ent_EDms/edit#gid=867169230) for the choice of verbs, activities, results, etc.   as much as possible and made use of the [TinCanPython library](https://pypi.org/project/tincan/) for making Statement classes, LRS client, etc..  There's also an [Ansible role](https://github.com/appsembler/configuration/blob/appsembler/ficus/master/playbooks/roles/xapi_bridge/), currently on a Ficus branch, for installing it.

nasthagiri:
Hey everyone, I'm going to spend some time focusing on this OEP again.  There is interest in moving this initiative forward from many different angles and by multiple development groups in the open edX community.  My plan is to:

* Generalize the OEP so it is not specific to just xAPI - but still include the specifics of xAPI as an example (perhaps moving xAPI details to a sub-document).
* Respond to and update the OEP based on later comments from @roishillo, @sampaccoud, and @bryanlandia that I missed earlier.
* Send the OEP out for review.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Yes, good point.  I added this as a 3rd point.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Ok. I updated the LRS - Future section to indicate that we will not implement an LRS ourselves, but will look into integration with 3rd party implementations.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Great.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Interesting.  We don't have that situation ourselves today - although we do plan to support "anonymous access to courseware".  Though, wouldn't you want to distinguish events from one anonymous user from another anonymous user?

oeps/oep-0026-arch-xapi-integration.rst
```diff
@@ -0,0 +1,483 @@
...
+.. _Using consistent user identifiers in Segment events: https://openedx.atlassian.net/wiki/spaces/AN/pages/144441849/Using+consistent+user+identifiers+in+Segment+events
+.. _`Deep Dive: Actor/Agent`: https://xapi.com/deep-dive-actor-agent/
+
+xAPI Verb
+~~~~~~~~~
+
+The **Verb** in xAPI is a past tensed value, identified by a URI from the `xAPI registry`_
+and paired with a short display string. It denotes the action that was performed by the
+Actor on the Object in the statement. We will use only values found in the registry and
+document which values are used for which edX events in the `Open edX Events`_ section below.
```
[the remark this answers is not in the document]
nasthagiri: Sure.  I mean to say that we will try to use standard and registered Verbs where ever possible - rather than creating our own custom ones.  I'll update the text to make this clearer.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Good point.  I update the spec to exclude the "display" field.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: So in which field do you send the "course" information for the event?  Did you take a look at the specific event types in the [spreadsheet](https://docs.google.com/spreadsheets/d/1Qx-1NkpCHXkWh8AagwHD5vzyBCdRXQVof10Ent_EDms/edit#gid=52329066)?  Currently, we proposed sending only 1 value in the parent - rather than a list.  But the parent is typically either the course or the parent module.

oeps/oep-0026-arch-xapi-integration.rst
nasthagiri: Please see the updated version of the spec - as this has changed thinking about scalability concerns.

nasthagiri:
@bryanlandia Thanks for the info.  It's good to know about the progress on that effort.  For this particular OEP, since our focus is on a real-time eventing API, we chose not to integrate as a downstream client of tracking log persistence.  It would be great to hear your feedback on the mapping of Open edX events to xAPI events - perhaps we can share notes and even consolidate on a common solution as a community.

nasthagiri:
All, I have finally pushed an update to this OEP that generalizes this feature, adds use cases, and responds to outstanding comments.

I am looking for someone from the Open edX community who is willing to help as an Arbiter on this OEP.  Let me know (via Slack) if you would be interested in taking on this role.  Thanks.

oeps/oep-0026-arch-realtime-events.rst
```diff
@@ -0,0 +1,403 @@
...
+Events`_ design document.
+
+.. _Caliper Real-time Events: oep-0026/caliper-realtime-events.rst
+
+Anonymized User ID
+==================
+
+In order to uniquely identify a user in the Open edX system with a long-term
+and non-revealing identifier, an *Anonymized User ID* will be generated as a 
+SHA-256 hash of the Open edX user_id. The Open edX user_id is the **id** value
```
douglashall: How will we prevent user ID collisions when trying to aggregate events sourced from different Open edX installations?

nasthagiri:
Here are the notes (thanks @robrap) from an internal review of this PR. 

**Action Items:**

1. Update the eventing subsystem diagram and separate filter (with access checks) from the router component.
2. Check-in with @mulby on event-tracking’s pluggability model and see whether a new processor makes better sense than a new backend.
3. See whether Canvas has an xAPI schema defined.

**Notes:**

* Reviewed use cases.
* xAPI details to be reviewed by the community.
* Is this used for data-synchronization internally?
    * No - meant for integrations with third parties.
    * Could be a stand-in solution, but meant to be out-of-scope.
    * OEP is ok as-is, but Steve noted that Data Engineering may look at batching vs streaming.
* Is there a plan to naildown the technical details to deal with events at scale?
    * Delivery guarantees?  Identifier questions.
         * Hoping that Translator/Validator are separated out enough that as long as there are no bottlenecks, this can keep the system scalable.
         * Looking at Kinesis and Kafka for data synchronization. Hopefully we can make the routing more interchangeable.  Makes sense to not have this as part of the OEP.
             * What parts of these decisions should be shareable across other similar types of problems?
             * Event-driven APIs should be something someone could experiment.  For example, 2-microfrontends with back-ends communicating.  Might have public vs private services across services.  Maybe 3rd-party Public apis. 
             * We’ve got very few success stories for asynchronous communication.
    * “Tracking backend” is like the Router, and “Processors” like the Translator/Validator, you might have different Processors.
         * Very similar to our current eventing.  Take event, fork and transform it.  Segment is called the “Tracking backend”, but is the router.
    * Discussions of different technologies:
         * Maybe celery is good enough for now for data sync.
         *  Kafka provides data-replay, not just speed.
* Routing vs processing:
    * Can I translate my events afterwards?
    * Where does filtering come in?
         *  Admin/router - course restrictions, future restrictions.
         *  Maybe access control layer should be separate from router.  Filtering before routing.  Third-parties could have their own anti-corruption layer and formatting. 
* Does there need to be replay capabilities of events?
    * good question for product.
    * replay events for training purposes for adaptive learning -> via an LRS as shown on the xAPI sub-page.
* Requires a new ID for users.
    * Should this be persisted? Just hash the lms userid to not leak it.
    * What about integration of events from other sources.  In the event itself, you’d have the source.  Source + anonymous id would be unique.
* A past edX employee was adamant that we shouldn’t use xAPI for event log for some reason.  Concerns about choosing the verbs, etc. for xAPI. Does community have experience?
    * Got useful feedback from community about details of events that weren’t needed and could be dropped for a smaller payload.
    * Caliper is more opinionated around schema, which would be a good thing for us regarding integrations.
    * Look at Canvas xAPI format.  Adaptive engines - what do they expect?
    * The validators are a good idea.
* Maybe make it more explicit that we could introduce a new verb and not be blocked on recommending it gets added.
* Future: may want to re-run events against a store like 40+ terabytes in S3.

oeps/oep-0026-arch-realtime-events.rst
nasthagiri: As discussed and captured in the notes: The event itself has the source information, so the source + anonymous-id would be unique.

nasthagiri:
@mulby and @brianhw I finally got a chance to update the OEP based on our notes from several weeks ago.  Can you both review the latest version?  If all is good, I can merge this with the "Provisional" OEP status.

@mulby [This commit](7bdbd7dcf0c6dced899b02438ee2d077fd0ba95f) in particular relates to your input and feedback. 

Thanks!

nasthagiri:
Thank you all for your input and review on this OEP.   I am now merging it with "Provisional" status.  We are looking forward to having the community help us with the implementation efforts.

PR #78, opened 2018-08-22
Minor updates to match Hawthorn reality

nedbat:
Who should review this? Do we need to change the arbiter?

@nasthagiri @jmbowman ?


nedbat:
I've changed Eddie to Jeremy, and merged.

PR #79, opened 2018-08-23
OEP 0028: Content theming in React (Draft)

taranjeet:
It aims to add guidelines for how to create and use React components, so that content modification becomes easy.

This OEP is based on [this](https://github.com/edx/open-edx-proposals/pull/64#discussion_r195914311) discussion and this [gist](https://gist.github.com/bradenmacdonald/34bf8bfc8f737f9b23a2664091b8f115)

Note: As OpenCraft, our role is to define a proper solution and ensure that the future developments move in the right direction, but not committing on developing any of this at the moment.


edx-webhook:
Thanks for the pull request, @taranjeet! I've created [OSPR-2565](https://openedx.atlassian.net/browse/OSPR-2565) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

mduboseedx:
@taranjeet Thank you for the proposal. Feel free to ping me when you feel it is ready for an edX review

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
++-----------------+----------------------------------------------------------------+
+| Created         | 2018-08-17                                                     |
++-----------------+----------------------------------------------------------------+
+| `Review Period` | 2018-08-17 - 2018-09-02                                        |
++-----------------+----------------------------------------------------------------+
+
+Abstract
+-------
+
+OEP-0023 [1] addresses the styling part of a theming system based on React. This proposal adds guidelines for content modification aspect to the React based theming system
```
bradenmacdonald: Nit: "This proposal adds guidelines for _content_ modification ~~aspect~~ to the React-based theming system."

(remove "aspect", add period)

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+-------
+
+Open edX instances frequently wish to customize not only the colors and styles of the Open edX UI, but also to add, remove, and modify significant parts of the UI to suit their use cases and branding. Doing this in a way that is understandable, clean, and easy to maintain can be a challenge unless the original UI is designed with this in mind.
+
+Another pain of theming system is providing access to unrelated data to any component. This data may not be used by earlier components but is required in a particular theme. It becomes an important concern for writing a theming system where a global access to data to all the components in the UI system should be provided.
+
+Decision
+--------
+
+* We will build small modular components which will be non-customizable, stateless and contain HTML, and certain logic. We can see an example of a smaller component SiteLogo, which will later be used in a customizable component _Header.
```
bradenmacdonald: It's not really explained here that there are meant to be two types of components: internal/non-customizable and customizable ones.

At first read, this sentence just sounds like "we will build everything out of small components that cannot be customized."

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+                </header>
+            );
+        }
+    }
+    // Custom theme:
+    export const Header = MyThemedHeader;
+    export const MainNavWrapper = _MainNavWrapper;
+
+
+* We will provide support to add data or other components by using `props.children`. The `props.children` property is provided from React and can be extended to include specialized children like left and right for a Pane component. An example of this can be
```
bradenmacdonald: I'm not sure that this paragraph adds any useful information since it's just describing a core react feature?

>The `props.children` property is provided from React and can be extended to include specialized children like left and right for a Pane component.

Nit: If you want do do a `left` and `right` pane component, those props are not extensions of the `props.children` property. The `props.children` property cannot be "extended" in any meaningful sense. What you are describing is just using other properties in a similar way.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+            </div>
+
+        )
+    }
+
+    // use
+    <Pane left={<Image />} right={<Explanation />} />
+
+
+* We will use function and placeholder to add additional content in customizable components. We can take an example of the above DefaultTheme and see _MainNav where it has support to add additional nav links by overriding `extraNavLinks` function.
```
bradenmacdonald: Typo: "function**s**" and "placeholder**s**"

Typo: "additional content **to** customizable components"

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+            ];
+        }
+    }
+
+    // Custom theme:
+    export const MainNav = MyThemedNav;
+    export const MainNavWrapper = _MainNavWrapper;
+
+
+* We will provide support for a global redux store which will be Open edX global redux store. This will act as a central place to store data and will have access to all data available to the system.
```
bradenmacdonald: > We will provide support for a global redux store which will be Open edX global redux store. ... system

I think it would be better to say: "Each frontend (e.g. the LMS, or Studio) will have a global redux store that acts as a central place to hold the state of its UI."

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+* We will consider the layout of the data in the Open edX global redux store as a stable API. We will provide support to pre-fill the store with some common data like current user, current course, list of courses enrolled, etc. We will provide the choice for themes to fetch data that's not part of the Open edX global redux store from REST API's using built-in redux actions and store it in their own separate redux store. We will announce breaking changes if the layout of the data changes in global store.
```
bradenmacdonald: The redux store would be specific to each frontend (LMS, Studio, ecommerce, etc.) and not global to all of "Open edX".

> using built-in redux actions and store it in their own separate redux store

Actions tend to be fairly coupled to the reducers and overall store state, so I'm not sure how useful it would be to provide standard action [creator]s but expect them to work with custom stores/reducers. It might be better to just provide a standard API client library for the REST APIs, without redux actions, and let each custom theme that needs to use them include its own redux actions/reducers/state. But maybe this would help reduce boilerplate etc., and most any reducer could be adapter to ingest standard actions. Curious to know what others think.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+* We will use containers [2] to access data from the redux store and provide it to components via props. A container is a react component that has a direct connection to the state managed by redux and access data from the state via mapStateToProps. We will use Container as a mechanism to separate data access functionality from the Component. This way we can keep both non redux connected version as well as redux connected version of the same component.
+
+* We will have support to convert any component into a container if it needs to access any data from the redux store, which it currently does not have access to. We can see this by an example where NavbarHeader component initially displays site title. This component now needs to display authenticated username, which is there in the redux store.
```
bradenmacdonald: I feel like the example below should be showing a NavbarHeader component that doesn't use the username, and an inherited version that does. The example here doesn't work without the container, so it isn't an example of "converting any component into a container".

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+    const NavbarHeaderContainer = connect(mapStateToProps, null)(NavbarHeaderComponent);
+
+    // use NavbarHeaderContainer instead of NavbarComponent as it has access to the username
+
+Consequences
+------------
+
+Theming system becomes more robust to content modification. Any data be it static or dynamic can be easily added to an existing component. It also provides support to request any unrelated data from the global store, thereby giving better customization for a new theme.
+
+However, there will be cases when a component becomes too complete to use which will create the need to rewrite that component as a composition of smaller components.
```
bradenmacdonald: What do you mean by "too complete"? "too complex" ?

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I have updated this to include `NavbarHeaderWithUserName` which extends from `NavbarHeader` but completely overrides its render function. The extension will help in propagating other functionalities from `NavbarHeader` to `NavbarHeaderWithUserName`.
`NavbarHeader` contains support to display username which is overridden by a container. This also brings us to a point where sometimes both the component(to display additional properties) and container needs to be overridden in case we need to display some additional data.

taranjeet:
@mduboseedx : Thanks. Sure I will ping you when the proposal is ready for an edx review. :)

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+        render() {
+            return (
+                <header>
+                    <SiteLogo/>
+                    <MainNav/>
+                    <UserAvatar/>
+                </header>
+            );
+        }
+    }
```
xitij2000: You just mentioned that Customisable components will contain neither HTML nor logic, but this uses the HTML ``<header>`` tag? I think this can be confusing. I know what the previous statement is going for, but perhaps it can be phrased in a different way?

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+* We will follow a 2 step approach to create any UI Component. We will first break down a complex component into smaller non-customizable components. We will then compose these smaller components to form a complex UI.
```
xitij2000: I think this can be rephrased a little bit. Currently, it sounds like you already have a component, which is why you are breaking it down. 

Perhaps:
When creating any complex UI Component we will try to break it down into the smallest usable components, and build the component out of these small reusable pieces.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+    export const MainNav = _MainNav;
+    export const MainNavWrapper = _MainNavWrapper;
+
+
+Now if we want to customize our _Header component, we can easily do it like
+
+.. code-block:: js
+
+    // SiteLogo being updated in Header
+    class MyThemedHeader extends _Header {
```
xitij2000: As @bradenmacdonald mentioned elsewhere, this is using inheritance not composition.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+    const Header = (props) => {
+        return (
+            <h1>{props.title}</h1>
+            {props.children}
+        )
+    }
+
+    // use
+    <Header title={"Open Edx"} >
```
xitij2000: nit: you can remove the curly brackets here unless we want to emphasise that style. Which I don't think we should.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+.. code-block:: js
+
+    // Customizable Main Navigation Area
+    class MyThemedNav extends _MainNav {
+        get extraNavLinks() {
+            return [
+                <a href="/about">About Us</a>,
+            ];
+        }
```
xitij2000: I think a better approach is to use [React fragments](https://reactjs.org/docs/fragments.html) here. So it could be:
```javascript
return (
    <Fragment>
        <a href="/about">About Us</a>,
    <Fragment>
);
```
or
```javascript
return (
    <>
        <a href="/about">About Us</a>,
    </>
);
```

oeps/oep-0028-content-theming-in-react.rst
xitij2000: I don't think we can expect that there will be 100% coverage for each API in the frontend, so people will have to create their own actions. I don't think it's reasonable to have the actions and reducers etc for data not used by the default theme. Where a built-in action exists, chances are that data is already loaded, and if not, the standard actions can perhaps be used.

While an API client could be automatically generated using Swagger or such, I think there should probably be multiple API clients for multiple sets of APIs because the API is huge so why should all that code be loaded if most of it isn't used. If a custom theme uses some API that the default theme doesn't, it can also include the relevant API client. This probably should cover the whole API.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+
+* We will provide support for a global redux store which will be Open edX global redux store. This will act as a central place to store data and will have access to all data available to the system.
+
+* We will consider the layout of the data in the Open edX global redux store as a stable API. We will provide support to pre-fill the store with some common data like current user, current course, list of courses enrolled, etc. We will provide the choice for themes to fetch data that's not part of the Open edX global redux store from REST API's using built-in redux actions and store it in their own separate redux store. We will announce breaking changes if the layout of the data changes in global store.
+
+* We will use containers [2] to access data from the redux store and provide it to components via props. A container is a react component that has a direct connection to the state managed by redux and access data from the state via mapStateToProps. We will use Container as a mechanism to separate data access functionality from the Component. This way we can keep both non redux connected version as well as redux connected version of the same component.
```
xitij2000: Maybe container is a popular enough phrase, but I still think we should probably avoid it. We've already mentioned we should prefer composition, and what is a Container but a Component that composes data access functionality with an existing component. 

So what we want is that a components' functionality should be implemented separately from the data access functionality.

oeps/oep-0028-content-theming-in-react.rst
xitij2000: I also don't like the idea of "converting" we are not really converting anything, we are just adding another component that passes some props by default by fetching that data from the store. So it's standard component composition, you're wrapping an existing component with another.

taranjeet: @xitij2000 : I have updated the "converting" part to "composition" now. Thanks:)

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I have now updated this to 

```
These customizable components will contain little or no HTML nor logic
```

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I have addressed this.

oeps/oep-0028-content-theming-in-react.rst
taranjeet: Sure, I have removed the curly brackets.

oeps/oep-0028-content-theming-in-react.rst
taranjeet: Perhaps I think you are talking about [this](https://github.com/taranjeet/open-edx-proposals/pull/1#discussion_r212421128) comment. That was for another point wherein the point I talked about using "composition" but gave an example of "composition". I updated it then only.

Also here it makes sense to use "inheritance" only since if we use composition we would end up creating components like 'BaseHeader`, `Header`, `SiteLogoHeader` and `MyCustomAnimatedLogoWidgetHeader`.

Let me know if this is not right.

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I am not sure that if I don't write about `Containers` what else should I write? 

My intent was here to describe `Containers` as something where we can do everything related to data and `Components` to keep away from data.

@xitij2000

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,246 @@
...
+
+.. code-block:: js
+
+    const SiteLogo = (props) => {
+        return (
+            <img src={props.url} alt={props.altText} />
+        );
+    }
+
+* We will build the complete UI out of the above small modular components. We will generally use composition to build a customizable component. However, inheritance can be used at times. These customizable components will neither contain HTML nor logic. We can see an example of a customizable component called the _Header component, which is formed by the composition of SiteLogo, MainNav and UserAvatar component.
```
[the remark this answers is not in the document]
taranjeet: I agree. What I have done is added about this in the other point and given an example of both cases. Thanks :)

oeps/oep-0028-content-theming-in-react.rst
taranjeet: @xitij2000 @bradenmacdonald I have updated this point but I am not sure if it covers your concern. Please let me know your thoughts on this. Thanks.

taranjeet:
@xitij2000 @bradenmacdonald This is ready for your review. Thanks :)

oeps/oep-0028-content-theming-in-react.rst
xitij2000: My point is that you mention in the paragraph: "We will prefer composition whenever adding any new component or content and inheritance when replacing or modifying the component." Immediately after that, you are showing an example of using inheritance when that is not the preferred approach. Show both ways of doing this, so that it's clearer which is the preferred approach in what circumstance. 

I would also say this is not a good example of inheritance at all. You are extending an existing component, but there is nothing from the original component that you are actually using. Why couldn't this just be a new component? The only thing it gets from ``_Header`` is the ``getNavLinks`` getter method and that isn't even being used, the ``render`` method is entirely replaced. 

Also, I think if a component has a good chance of needing modification in some area, there should probably be a hook for it. So, in this case, replacing the Logo in the ``_Header`` is a pretty obvious modification, so perhaps there should already be a prop for it, and should just default to the standard logo.

oeps/oep-0028-content-theming-in-react.rst
xitij2000: Arrays are just a way of returning multiple elements, and that is now officially supported by React, [Fragments are a replacement for return an array](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html).

Thanks for pointing out the comma issue, you're right that it would be rendered in this case, I just forgot to clean them up while copy-pasting from the existing code.

oeps/oep-0028-content-theming-in-react.rst
taranjeet: This is addressed now. Thanks

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I have updated the snippet to use `React.Fragments`. Thanks :)

oeps/oep-0028-content-theming-in-react.rst
taranjeet: @xitij2000 : I am not sure here but I think we should skip this difference part between composition and inheritance since that is something more specific to React and not this proposal in general. The line where we can say that here inheritance or composition which one should be used is very blur.

Keeping this in mind, I think its better to just write `via composition/inheritance` and leave it to the one who is implementing it because he/she will be a better judge of the situation. 

I have also commented on the ticket about how we can use composition for the Header component but that turned out to be a discussion on inheritance vs composition.

```
class Header extends React.PureComponent {
    public render() {
        return (
            <header>
                {props.children}
            </header>
        );
    }
}
 
const BaseHeader = (props) => {
    <Header>
        {props.children}
        <MainNav/>
        <UserAvatar/>
    </Header>
}
const SiteLogoHeader = (props) => {
    <BaseHeader>
        <SiteLogo />
    </BaseHeader>
}
 
const MyCustomAnimatedLogoHeader = (props) => {
    <BaseHeader>
        <MyCustomAnimatedLogo />
    </BaseHeader>
}

``` 
I think Braden can help us here.
@bradenmacdonald : Your inputs please :)

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,280 @@
...
+-------
+
+OEP-0023 [1] addresses the styling part of a theming system based on React. This proposal adds guidelines for content modification to the React based theming system.
+
+Context
+-------
+
+Open edX instances frequently wish to customize not only the colors and styles of the Open edX UI, but also to add, remove, and modify significant parts of the UI to suit their use cases and branding. Doing this in a way that is understandable, clean, and easy to maintain can be a challenge unless the original UI is designed with this in mind.
+
+Another pain of theming system is providing access to unrelated data to any component. This data may not be used by earlier components but is required in a particular theme. It becomes an important concern for writing a theming system where a global access to data to all the components in the UI system should be provided.
```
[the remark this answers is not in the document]
taranjeet: @xitij2000 : I am a little afraid about mentioning the example due to the following reasons
* The same Header and Header with username example is explained in the "composing a component into container" point
* It is the [context](https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-templates/adr-based-template.rst#context) section of the proposal. Should it include an example or just simply describe facts?

I have addressed your nit and sentence restructuring comments. Thanks :)

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,280 @@
...
+            <React.Fragment>
+                <MyThemedHeader />
+                <h3>{props.title}</h3>
+            </React.Fragment>
+        )
+    }
+
+We can see that in the above example we are using composition since we are adding a component to the existing MyThemedHeader component.
+
+* We will provide support to add data or other components by using `props.children`. The `props.children` property is provided from React and is helpful when a component doesn't know its children ahead of time. An example of this can be
```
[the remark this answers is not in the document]
taranjeet: I agree that this is a standard React feature, but this is something which is helping us write a system which is more flexible towards content updates. Still, should I remove it?
@xitij2000

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,280 @@
...
+
+* Wherever we are developing a component that needs to use data from the redux store we will never do so directly in the component implementation. A separate component should be created that will be solely responsible for accessing the data from the store and passing it to component via props. In React parlance such a component is called a "Container" [2] component, and this term will be used henceforth in the OEP. A container is a react component that has a direct connection to the state managed by redux and access data from the state via mapStateToProps. This way we can keep both non redux connected version as well as redux connected version of the same component.
+
+* We will have support to compose any component into a container if it needs to access any data from the redux store, which it currently does not have access to. We can see this by an example where NavbarHeader component initially displays site title. This component now needs to display authenticated username, which is there in the redux store.
```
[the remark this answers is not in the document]
taranjeet: @xitij2000 : Same comment about the removal goes here.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,280 @@
...
+Open edX instances frequently wish to customize not only the colors and styles of the Open edX UI, but also to add, remove, and modify significant parts of the UI to suit their use cases and branding. Doing this in a way that is understandable, clean, and easy to maintain can be a challenge unless the original UI is designed with this in mind.
+
+Another pain of theming system is providing access to unrelated data to any component. This data may not be used by earlier components but is required in a particular theme. It becomes an important concern for writing a theming system where a global access to data to all the components in the UI system should be provided.
+
+Decision
+--------
+
+* We will build two types of components:
+
+  1. Customizable components which will compose others components
```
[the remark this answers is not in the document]
taranjeet: This is addressed.

oeps/oep-0028-content-theming-in-react.rst
bradenmacdonald: I believe that inheritance will be necessary any time you want to _remove_ or _replace_ an existing component child such as the `SiteLogo` in the example above, unless there is some other hook for doing so via props (like `logoOverride=<MyCustomAnimatedLogo />`) or modifying a global (like replacing a component globally via `const SiteLogo = MyCustomSiteLogo; // not StandardSiteLogo;`)

xitij2000: @bradenmacdonald How will inheritance help in that case though? I might be missing something but here is how I see it:
Case: You have a component where the logo should be removable/replaceable. 
* Provide a prop that allows replacing the logo. e.g. ``<div> {this.props.logo === undefined?<DefaultLogo />: this.props.logo}</div>``. This allows setting the logo as null or providing a custom logo without inheritance. 
* You inherit from the original component. But now you're overriding the original component's render method. So what exactly do you gain by inheriting vs creating a new component? The use for inheritance would be with components that have a lot of logic that will be inherited, but customisable components won't have that.

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
xitij2000: Actually, I think that while this won't be the definitive dev guide, it is something people might refer to during development, so having an example might be helpful. I think showing the same example in differnt contexts can be useful, in fact having a consistent example is probably better. This is looking at that code in a completely different light.

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
xitij2000: I think ``props.children``  and what it does should be familiar to anyone with a basic knowledge of React. So it's probably better to link to the official React development guide and just mention here the areas where we differ from them.

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
xitij2000: I think it's better to link to the Redux+React guide in the beginning of the OEP and only mention places where we differ from the standards.

oeps/oep-0028-content-theming-in-react.rst
xitij2000: > Maybe we should come up with more concrete examples to figure this out better? It's still fairly theoretical at this point.

Yup, I'm just having trouble seeing a component that has a such a straightforward render function, but it still makes sense to inherit, vs create a new component. I can see the case if there is some lifecycle code or other kind of complex code in the component, but I assumed that customisable components wouldn't have that.

xitij2000:
@bradenmacdonald Yup, that is what I meant. Although now that I think about if the theme definition is the "entry-point" then that's a more flexible way of doing what I suggested because the theme definition can point to the base theme or elsewhere.

oeps/oep-0028-content-theming-in-react.rst
taranjeet: I have removed the point where `props.children` and `props.left/right` were being mentioned. My reason to remove is as follows
* These are standard react features, anyone developing in React must be already familiar with it
* We are not doing anything additional from the standard convention.

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
taranjeet: Removed this point since we were not doing anything different from the standard convention.

taranjeet:
@xitij2000 @bradenmacdonald @mtyaka : This is ready for your review. I have now added some use cases of why should we opt for inheritance, updated some confusing examples, added some points and examples as well.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    // NavbarHeader container
+    function mapStateToProps(state) {
+        return {
+            title: state.title,
+            username: state.username
+        }
+    }
```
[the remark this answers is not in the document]
bradenmacdonald: I think the example that Matjaz shows here would work fine? Why wouldn't it?

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
bradenmacdonald: I don't think this paragraph adds much useful info; it is describing what is the main/clear way to do it using React+Redux anyways. You could include a code example if you want, to demonstrate this.

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
taranjeet: My only concern here was how will the theme access this function? These functions(mapStateToProps and mapDispatchToProps) are generally internal to a component and not exported. Will we be exporting them as well?

oeps/oep-0028-content-theming-in-react.rst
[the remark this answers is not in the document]
taranjeet: I have included the example and merged this point with the previous one.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,334 @@
...
+    export const Header = _Header;
+    export const MainNav = _MainNav;
+    export const MainNavWrapper = _MainNavWrapper;
+
+Now if we want to customize our ``_Header`` component, and use ``MyCustomAnimatedLogoWidget`` instead of ``SiteLogo``, we can do it as
+
+.. code-block:: js
+
+    const MyThemedHeader = (props) => {
+        return (<Header props={<MyCustomAnimatedLogoWidget/>} />)
```
mtyaka: Should this be `logo={...}` instead of `props={...}`?

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,334 @@
...
+
+
+* We will provide support via props to control parts of the component when composing components. An example of this can be ``Button`` element
+
+.. code-block:: js
+
+    class Button extends React.PureComponent {
+        render() {
+            return <button
+                color={this.props.color}
```
mtyaka: I would remove `color`, since colors and other style modifications will be better supported through CSS/bootstrap.

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,334 @@
...
+                <CardBody>
+                    <CardTitle />
+                    <CardText />
+                    <Button onClick={this.viewCardDetail(this.props.courseId) />}
+                </CardBody>
+            </Card>
+        }
+    }
+
+3. Inheritance often becomes necessary and useful, when we want to override the rendering functionality of any component, and still maintaining access to lifecycle code. Overriding functionality can include removing, re-ordering, replacing or inserting children. An example of this can be Navbar, where the default Navbar has a SearchForm which is left aligned. This Navbar component can now be inherited to place SearchForm as right aligned.
```
mtyaka: > Overriding functionality can include removing, re-ordering, replacing or inserting children.

I think this covers all three examples? Example 1 inserts a child, example 2 replaces one child, and example 3 reorders it. All three examples are basically only rewriting the `render` function, so I don't think we need all three of them - example 3 is enough. Another good example of inheritance is the one from before this section, where we are overriding `extraNavLinks`. I think we should put these two examples in the same section, and remove redundant examples 1 & 2 from this section.

taranjeet: Sure, I have refactored the point 3 and removed point 1 and 2. Thanks :)

taranjeet:
@mtyaka : I have added some detailed points about what makes a customizable component customizable and something similar for non customizable component. I have also documented that we will mark components customizable and document the methods and props available. Please review.

CC : @xitij2000 @bradenmacdonald

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,279 @@
...
+                <header>
+                    {this.props.logo ? this.props.logo: <SiteLogo/>}
+                    <MainNav/>
+                    <UserAvatar/>
+                </header>
+            );
+        }
+    }
+
+* We will specify what are the customizable components, document whether the customization is available via custom props or overriding superclass method. We will announce breaking changes if any changes are made to the component. We will prefer that customizable components be only customized, however it is totally up to to the theme owner to customize any components that is not marked as customizable by edX.
```
mtyaka: I suggest changing this to:

"We will specify which components are customizable and document whether the customization is available via custom props or overriding superclass methods. We will announce breaking changes if any changes are made to customizable components. Users will still be able to override components which are not explicitly marked as customizable, but edX will provide no backwards compatibility guarantees for them."

taranjeet:
@mtyaka : Thanks a lot for the review.

@mduboseedx It is ready for an edx review. :)

mduboseedx:
Also including the @edx/arch-review team

oeps/oep-0028-content-theming-in-react.rst
```diff
@@ -0,0 +1,279 @@
...
+                <React.Fragment>
+                    <a href="/about">About Us</a>
+                </React.Fragment>
+            );
+        }
+    }
+
+* Each frontend (e.g. the LMS, os Studio) will have a global redux store that acts as a central place to hold the state of its UI.
+
+* We will consider the layout of the data in the redux store specific to each frontend (LMS, Studio, ecommerce, etc.) as a stable API. We will provide support to pre-fill the store with some common data like current user, current course, list of courses enrolled, etc. We will provide the flexibility for themes to fetch data that are not part of the redux store from REST API's using custom redux actions and store it in their own separate redux store. We will announce breaking changes if the layout of the data changes in global store.
```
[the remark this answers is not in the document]
xitij2000: @nasthagiri One way to do that could be to create [selector functions](https://redux.js.org/recipes/computingderiveddata) for accessing any state data. 

A simple example would be: 
```javascript
const getUserProfileImage = (state, userId) => state.userProfiles[userId].profileImage;
```

If the state shape ever changes this function can be modified to still return backwards-compatible data. With this approach, the only supported/stable API will be these functions. You can still use the raw state but at your own risk. 

The functions themselves could probably be marked as stable/unstable using documentation, a prefix like _ or the package they are placed in.

nasthagiri:
@mtyaka After consideration, we'd prefer not to merge OEPs while they are still in "Draft" status since that will pollute and confuse OEP readers.  So I've just opened a [PR to add a new "Provisional" status to OEPs](https://github.com/edx/open-edx-proposals/pull/85).  A "Provisional" status will allow us to generally agree upon a direction and then transition to an "Accepted" status once we have viable reference examples and have adopted the OEP into the platform.  However, a "Provisional" status will still need to undergo a Review process with a dedicated Arbiter.

Given that, for these foundational Theming-related OEPs, since both edX and the community are keen stakeholders - with edX as the primary maintainers/contributors and the community as the consumers/contributors - I would like to assign an Aribiter from edX.  As I mentioned earlier, the edX Architecture team plans to [include Theming in its Roadmap](https://openedx.atlassian.net/wiki/spaces/AC/pages/787087363/Architecture+Roadmap+2018-19) in the 1st half of next year.  So I'd like to propose that we keep this OEP in Draft form as a running open PR until an edX team is able to focus on helping carry these ideas through - which at this time, seems like early 2019.

This will require us to keep this OEP hanging around for a while in Draft state (in PR form), but once we have a focused effort on this, the work you have done so far will be valuable input for us to be effective in moving forward rapidly.

mtyaka:
Sounds good @nasthagiri, thank you for the information. Nice to hear the Architecture team is including theming in their Roadmap soon :)

natabene:
@nasthagiri  @ormsbee Is this a good time to return to this PR or not yet?

nasthagiri:
@natabene We are aware of these 2 theming-related proposals.  We plan to address them as part of our  [replatforming runway](https://openedx.atlassian.net/wiki/spaces/AC/pages/948175209/Replatforming+Runway+Plan).  As indicated on our [Roadmap Progress](https://openedx.atlassian.net/wiki/spaces/AC/pages/787087363/Architecture+Roadmap+2018-19#ArchitectureRoadmap,2018-19-ArchitectureRoadmapProgress), we are currently focusing on our "Design System" (library of reusable components).  We plan to address theming only after that - but still in the 1st half of this year as previously indicated.

nasthagiri:
Also, for any of you following this PR, please contact me (via slack/email) if you'd like to join an Open edX "Theming Advisory Group".  We plan to consult this group on requirements and thoughts on "theming" on the Open edX platform.  Thanks.

natabene:
@nasthagiri Thank you for the update, super helpful to me.

edx-webhook:
Thanks for the pull request, @taranjeet! I've created [OSPR-3593](https://openedx.atlassian.net/browse/OSPR-3593) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

mtyaka:
@nasthagiri I don't have access to the [replatforming runway](https://openedx.atlassian.net/wiki/spaces/AC/pages/948175209/Replatforming+Runway+Plan) or [roadmap progress](https://openedx.atlassian.net/wiki/spaces/AC/pages/787087363/Architecture+Roadmap+2018-19#ArchitectureRoadmap,2018-19-ArchitectureRoadmapProgress) docs.

Has there been any progress on made on theming?

mtyaka:
@nasthagiri There has been some exciting development around micro-frontends in the past year. Is there any documentation or guidelines on how how to theme micro-frontends?

nasthagiri:
@bradenmacdonald Are there elements of this proposed OEP that you believe are still valid given the latest developments in Open edX MFEs?  If not, should we close this old Draft and reopen with the latest considerations?

bradenmacdonald:
@nasthagiri My understanding is that newer guidance is [here in extension_points.rst#Theming Microfrontends](https://github.com/edx/edx-platform/blob/master/docs/guides/extension_points.rst#theming-microfrontends), and specifically that the proposals in this OEP would instead be replaced by "Frontend Plugins" or entire custom MFEs (as well as [Overriding Brand Specific Elements](https://edx.readthedocs.io/projects/edx-developer-docs/en/latest/developers_guide/micro_frontends_in_open_edx.html#overriding-brand-specific-elements)).

AFAIK frontend plugins still needs to be defined, but there's a proposal at https://github.com/edx/frontend-app-discussions/pull/5

I think it makes sense to close this and re-open when there's a solid, tested frontend plugins approach that we're ready to recommend more widely across MFEs.

nasthagiri:
Thanks, @bradenmacdonald.  I agree that the various efforts you enumerated will collectively form a basis for the extensibility and configurability of our frontends.  Once those efforts are matured enough past their proof-of-concept phases (in ADR forms), we can create/update a frontend OEP that provides a cohesive narrative.

Given that, I will close this PR. I'm looking forward to seeing these other efforts move forward.

Cc: @abutterworth @davidjoy

edx-webhook:
@taranjeet Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/open-edx-proposals&entry.867055334=https://github.com/edx/open-edx-proposals/pull/79&entry.1484655318=https://github.com/taranjeet&entry.752974735=2018-08-23+21:25&entry.1917517419=2020-09-04+22:55&entry.2133058324=No) so we can improve your experience in the future.

PR #81, opened 2018-09-19
OEP-30: PII Markup and Auditing

bmedx:


oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+- Run the PII documentation tool to update the PII documentation when they add, remove, or update a PII annotation (see `Documentation Tooling`_ for details)
+
+
+Code Reviewer Responsibility
+----------------------------
+It becomes the responsibility of code reviewers to confirm the developer assertions about the presence of PII in their pull request are accurate, that retirement steps and annotations are present and correct when necessary. This is especially important with pull requests coming from outside of edX, where the original developer may not know of this OEP and their responsibilities in regards to PII.
+
+Responsibility for Third-party Service Integrations
+---------------------------------------------------
+When dealing with third-parties that may store PII (ex. Optimizely, Google Analytics, Sailthru) the implementing developer(s) or team members should work with the drivers of the feature and legal to ensure that:
```
iloveagent57: I think "legal counsel" works better here to address maintainers of open edX instances that aren't edx.org - they may not have a legal department (which "legal" is presumably in reference to).

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+
+Github Pull Request Templates
+-----------------------------
+In order to assist developers in remembering to check all new data for PII, each Open edX repository that might store such data will have a GitHub pull request template that reminds the developer and reviewers to check for the addition of such data in their commits and asks them to affirmatively state that either no such data exists or that it does exist and that appropriate retirement steps are, or will be, ready to retire that data before the request is merged.
+
+Docstring Annotations
+---------------------
+**Python**
+
+When adding or modifying **any** data storing models (ex. Django model, MongoDB model) an annotation must be added stating whether the model does or does not store PII. The annotation should be added to the comments of the storage class where the data storage is defined, or the calling calling method / function if there is no storage class. Calls to third-party services that store data must be annotated to indicate the presence of PII.
```
iloveagent57: duplicate "calling calling"

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+Example 1::
+
+    class ApiAccessRequest(TimeStampedModel):
+        """
+        Model to track API access for a user.
+
+        .. pii:: Contains website and employer information about a linked User. Retired: Directly in the LMSAccountRetirementView endpoint.
+        """
+
+If a project requires another project that stores PII, such Django being used in edx-platform, the developer must annotate the place(s) in code where that package is being called to store the PII with the same docstring annotation as if it were a storage class.
```
iloveagent57: s/such Django/such as Django/

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+        <!-- begin segment footer -->
+        <!-- .. pii:: The user is identified to Segment by username and email here, analytics data will be associated with this user based on their browsing. See Segment documentation for details. Retired: In Segment retirement step in Tubular -->
+        <script type="text/javascript">
+        % if user.is_authenticated:
+            ...
+
+Example 2::
+
+    <script type="text/javascript">
+    // .. pii:: The user's address email address is sent to the billing provider here. Retired: This information is not retired as it is necessary to keep for legal and financial reporting reasons.
```
iloveagent57: Is it worth calling out this case independently of this example?  I feel like this is the number one reason that PII information would be kept around.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+
+This tool may take the form of a test run inside the project which looks at all installed apps and their models for docstrings containing PII. Given that this list will contain many third party packages we will also need to maintain a list of packages and models that will not be checked, as well as a list of PII stored in those packages / models. These lists will need to be hand maintained by the developers adding or modifying packages. This mechanism will also allow the rollout of the annotations to take place over time.
+
+Documentation Tooling
+---------------------
+A tool will be created that reads the annotations in each PII-containing repository and generates a reStructuredText (reST) file named ``pii.rst`` which will be located at the top level directory of the repository or with the repository's documentation and linked to from the top-level README file. This file will gather all of the PII annotations for the project in one place so that the PII load of any given project can be quickly seen and understood. Projects that do not have PII may have their top level README file updated to reflect that.
```
iloveagent57: Wow!  That will be nice to have.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,268 @@
...
+Several ways of making the locations of PII storage auditable were tested in forming this OEP (see `Rejected Alternatives`_). Annotations have the following benefits:
+
+- Clearly show PII locations when working with source
+- Set us up for easily putting this information into automatically generated documentation in the future
+- Do not create Django migrations
+- Do not incur runtime costs, aside from tiny growth of code file sizes
+- Are relatively low-effort to implement and maintain
+- Have a very low likelihood of causing bugs
+
+Sphinx-style annotations were chosen due to Sphinx's wide adoption in the Python, Django, and edX ecosystems. While we have had challenges using Sphinx to document edx-platform several other Open edX repositories already use Sphinx to generate documentation. Even if we never update edx-platform to use Sphinx these identifiers are unique enough to allow us to audit them with a high degree of confidence.
```
iloveagent57: put a comma between "edx-platform" and "several".

oeps/oep-0030-arch-pii-markup-and-auditing.rst
bmedx: I addressed your other comments and added a line detailing this under the "Developer Responsibility" section.

trevorbass:
This OEP makes a lot of sense overall. Smart implementation, and generally strikes a good balance for a challenging topic.

A few recommendations/questions:

(1) [Primary recommendation:] We should consider mandating annotations for models storing user activity as well. Users may be identified creatively by bad actors even without PII as identified here. For example, if someone has a list of certificates for an individual and that individual has shared the certificates on LinkedIn it will not be difficult to deanonymize that data. For some additional examples/context, see "Person Course Deidentification.pdf" referenced here: https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/26147&version=10.0,

(2) Should user id be included as an example of PII?

(3) Do we have a sense of what fraction of models contain PII?

(4) [Not a recommendation, more of a question:] Have we considered multiple levels of PII? For example, attributes like Job Title, Gender, and Sex are much less specific to individuals; models that contain only demographic features like this but not attributes like name or address or IP address might be considered less risky. I'm not necessarily suggesting we do this.

(5) Do we have concrete plans of how we will use this information (PII classification of models) once we have it?

bmedx:
@trevorbass thanks for the feedback!

(1) This sounds like a reasonable addition. I'll take a look through the PDF and try to get a better understanding of what the boundaries of "user activity" might be. I'll probably have some additional questions for you as I get into that.

(2) User IDs are a special case and probably deserve their own section. Would you expect user ids to be annotated in every location we use them throughout a system? My instinct would be to annotate them when being stored in a separate system (ex. Ecommerce storing LMS user ids), but not necessarily everywhere internal (ex. LMS storing LMS user ids with no other PII). Definitely open to other thoughts on this.

(3) By the definitions we've been using for retirement purposes it's a pretty small number. Part of the reason for this OEP is to be able to answer questions like that more precisely, though. :)

(4) I considered having a severity level assigned to an annotation when drafting this, but could not think of an immediate use case for it. My assumption is that the annotation consumers will have the context to assess the risks based on their specific use cases. Once the code is annotated it should be fairly easy to go back and add them if we find a need for them.

(5) The most immediate use will be to create a development process that puts thought about PII front and center, and helps keep our retirement processes up to date as new data comes in. Other immediate uses will be for the open source community to be able to identify user data that they need to address as part of their own retirement processes, and allowing consumers of edX code (internal and external) to easily audit the PII load of any repository we maintain.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+- The process for cataloging PII stored in non-operational systems such as analytics databases
+
+Motivation
+==========
+
+The primary goal of this OEP is to allow Open edX administrators and developers to understand which information needs special handling, and identifies our best practices for how to make those designations obvious. Respectful handling of PII depends on the following principles, which must be measured against our ability to run the service and meet our research goals and partner obligations:
+
+#. Storing the minimum personal data necessary
+#. Sharing the minimum personal data necessary
+#. Removing the maximum amount of stored personal information when requested
```
nasthagiri: How is #3 in this list different from #2?

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+- Date of birth (including partial such as year of birth)
+- IP addresses
+- Job title
+- Gender
+- Sex
+- Social media accounts
+- Website
+- Biography fields (free-form text fields intended to introduce one's self)
+- Forum posts
+- Notes service entries
```
nasthagiri: Does the inclusion of the last 2 (`Forum posts` and `Notes service entries`) imply that we consider any open-ended text entries as possible PII?  For example, are ORA submissions also in a similar category?

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+Specification
+=============
+
+Developer Responsibility
+------------------------
+The responsibility for identifying and appropriately labeling PII rests on the developers working in Open edX code. When any new information is being saved to a persistent storage medium (ex. MySQL, Mongo, S3, reporting services, 3rd party marketing tools) the developer must identify whether any of that information may be (alone, or in conjunction with other stored data) PII, seeking legal assistance if necessary. Specifically the developer's responsibilities are:
+
+- Avoid storing PII when it is not necessary
+- Ensure that any PII that is stored will be retired upon learner request *before* that information is stored in a production environment
+    - Exceptions may be made for classes of PII that at need to be kept for legal, financial auditing, or research purposes. Consult legal counsel for approval and annotate appropriately if you encounter such a case.
```
nasthagiri: Typo? `that at` -> `that`

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+- Ensure that any PII that is stored is annotated appropriately (see `Docstring Annotations`_ for details)
+- Run the PII documentation tool to update the PII documentation when they add, remove, or update a PII annotation (see `Documentation Tooling`_ for details)
```
nasthagiri: Nit (for consistency): `they` -> `you`

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+
+        .. pii:: Contains website and employer information about a linked User. Retired: Directly in the LMSAccountRetirementView endpoint.
+        """
+
+If a project requires another project that stores PII, such as Django being used in edx-platform, the developer must annotate the place(s) in code where that package is being called to store the PII with the same docstring annotation as if it were a storage class.
+
+Example 2::
+
+    # ..pii:: Learner email is sent to Segment in the following line and will be associated with analytics data. Retired: In Segment retirement step in Tubular.
+
```
nasthagiri: Although intentionally not covered in this OEP, is the retirement process/tooling (including Tubular) intended to be available for Open edX installations?

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+        <!-- begin segment footer -->
+        <!-- .. pii:: The user is identified to Segment by username and email here, analytics data will be associated with this user based on their browsing. See Segment documentation for details. Retired: In Segment retirement step in Tubular -->
+        <script type="text/javascript">
+        % if user.is_authenticated:
+            ...
+
+Example 2::
+
+    <script type="text/javascript">
+    // .. pii:: The user's address email address is sent to the billing provider here. Retired: This information is not retired as it is necessary to keep for legal and financial reporting reasons.
```
nasthagiri: Typo: `address email address` -> `email address`

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+
+**Javascript**
+
+When adding in Javascript that results in storage of PII to a location that is not covered by other annotations (ex. Segment), annotations should be added to the location(s) in script where the data is being sent. The annotations should take the same form as in Python as Sphinx can also operate on Javascript for documentation.
+
+Example 1::
+
+    % if settings.LMS_SEGMENT_KEY:
+        <!-- begin segment footer -->
+        <!-- .. pii:: The user is identified to Segment by username and email here, analytics data will be associated with this user based on their browsing. See Segment documentation for details. Retired: In Segment retirement step in Tubular -->
```
nasthagiri: Has the team considered expanding on the annotation somewhat so the tooling can more easily parse out the PII data?  For example, instead of listing `username and email` within an English sentence, we could additionally list them in a separate `.. pii_data::` line right below.  This has the benefit of adding more **consistent** structure around the annotation and allows for better **discovery** - for example, finding all code that touches a user's email address.

nasthagiri: More structure may also require developers to specify an explicit retirement strategy of the PII: deleted, anonymized, retained.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,269 @@
...
+
+Enforcement Tooling
+-------------------
+A tool will be created and integrated into the Open edX test / build systems that will examine all Django models in the project and ensure that they have PII annotations. It is acknowledged that this tool will not handle all cases where PII is stored, but represents an effort to enforce best practices on the majority of places where PII is stored in the Open edX ecosystem.
+
+This tool may take the form of a test run inside the project which looks at all installed apps and their models for docstrings containing PII. Given that this list will contain many third party packages we will also need to maintain a list of packages and models that will not be checked, as well as a list of PII stored in those packages / models. These lists will need to be hand maintained by the developers adding or modifying packages. This mechanism will also allow the rollout of the annotations to take place over time.
+
```
nasthagiri: For our own open edX code, doing static analysis makes sense.  For 3rd party database models/tables, is there a more automated approach that we can use at runtime?  Maintaining the information by hand - branching down the dependency hierarchy - every time we upgrade a library - will be costly.  If there is no alternative, then so be it.  But I'm wondering if there's tooling that can be a backup in case we err.

iloveagent57:
Notes from edX Arch Weekly Review: https://openedx.atlassian.net/wiki/spaces/AC/pages/854950584/2018-10-01+OEP-0030+Review

oeps/oep-0030-arch-pii-markup-and-auditing.rst
bmedx: @nasthagiri The second point refers to the sharing of information with other entities, while the third is about retiring as much as possible.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
bmedx: Free form text is a bit of a gray area. Where we can we retire it, but ORA submissions have wider ranging implications inside a course and for partner data sharing. @stroilova may have more thoughts on this.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
bmedx: Yes. Tubular is a public repo, so we provide all of the pieces we use for retirement except for secrets and our unique configuration. We have some documentation for that process, but also have more tickets to flesh that out with more detail in the backlog.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
bmedx: We discussed this in the arch review meeting, and updates for this should be in the next version which I hope to push today.

bmedx:
@nasthagiri @wesmason @trevorbass @iloveagent57 Thanks again for all of your feedback! I believe this latest commit addresses all of the feedback you and others have given to date. The diff is here if you'd like to review just the changes:

https://github.com/edx/open-edx-proposals/pull/81/commits/68564ccf7f9c6703e6dd92e96491e1358183edf4

@wesmason I took an initial stab at the `pii_types` based on what we've found in the platform and other things that have been mentioned. I know you were working on getting a list as well, so I'm happy to integrate or change as necessary, but I think the specifics of that list will be occasionally updated and shouldn't necessarily hold up the OEP process. Let me know if you feel differently!

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -135,26 +143,72 @@ It is important to note that under this OEP all Django model classes must be ann
...
 
-``.. pii:: <description of the PII>. Retired: <location / process of retirement>``
+``.. pii:: <required description of the PII>``
+
+``.. pii_types:: <comma separated list of the types of PII stored here, required if the pii annotation exists>``
+
+``.. pii_retirement:: <comma separated list of retirement types, required if the pii annotation exists>``
 
-Or, if there is no PII:
+``.. no-pii <:: optional description>``
```
pwnage101: probably use an underscore to be consistent.

oeps/oep-0030-arch-pii-markup-and-auditing.rst
```diff
@@ -0,0 +1,340 @@
...
+        .. pii_retirement:: local_api
+        """
+
+Example 2::
+
+    class NoPiiHere(Model):
+        """
+        This is an example model.
+
+        .. no-pii
```
wesmason: If we update the underscore elsewhere, should update here too.

PR #83, opened 2018-10-28
OEP-21: Update based on recent experience

nasthagiri:
This PR updates OEP-21 (Deprecation and Removal Process) as follows:

- Describes the process in specific ordered steps for one to follow.
- Adds references to external docs.
- Updates information on the JIRA ticketing process.


oeps/oep-0021-proc-deprecation.rst
```diff
@@ -22,295 +22,323 @@ OEP-0021: Deprecation and Removal
...
--  Accepted
+.. _edx-code: https://groups.google.com/forum/#!forum/edx-code
+.. _edx-code example: https://groups.google.com/d/msg/edx-code/779kAKas2Yw/jU7vGHu8CgAJ
 
-   -  Community feedback has been addressed and there is alignment on timeline.
+To openedx.slack.com
+^^^^^^^^^^^^^^^^^^^^
 
--  Deprecated
+Post the following in the #open-edx-proposals and #dev `openedx slack`_ channels:
```
pdpinch: Just curious: why #dev and not, say, #general or #ops?

oeps/oep-0021-proc-deprecation.rst
```diff
@@ -22,295 +22,323 @@ OEP-0021: Deprecation and Removal
...
+#. Include a description with links to what is being removed, in a
+   **Removal** section.
+#. Include a description with links to what it is being replaced by, in a
+   **Replacement** section.
+#. If you plan to mark the code for deprecation, explain how in a
+   **Deprecation** section. See Deprecated_ for considerations.
+#. If automated migration will be needed, explain your migration plan in a
+   **Migration** section.
+#. If there is any additional publicly shareable information or data from your
+   earlier analysis, include that on the ticket.
```
pdpinch: Can you add a bullet to check that the ticket is open for all JIRA users?

oeps/oep-0021-proc-deprecation.rst
nasthagiri: hmm.. someone recommended #dev to me (as it corresponded with an internal edX dev channel) but if these others are more appropriate, then we can change this.

@nedbat @mduboseedx  What do you think are the appropriate openedx slack channels for announcing deprecation and removals?

oeps/oep-0021-proc-deprecation.rst
nasthagiri: So.. when you ran into this access problem, the entire JIRA board for DEPR tickets were inaccessible to non-edX users.  Since then, we have updated the permissions.  Now it would actually be extra effort for someone to change the permissions on the ticket.  Do you think it's still worthwhile for us to add a bullet to have the ticket author double check the permissions?

oeps/oep-0021-proc-deprecation.rst
nasthagiri: @pdpinch I confirmed with the open edX team internally - and they agree with you that #general is more appropriate for this.  Thanks for pointing it out!  I'll update the doc.

oeps/oep-0021-proc-deprecation.rst
nasthagiri: I double checked - there is not even a "Security Level" setting on these DEPR tickets - so one cannot change the access level of these tickets.

oeps/oep-0021-proc-deprecation.rst
```diff
@@ -22,295 +22,323 @@ OEP-0021: Deprecation and Removal
...
+Allocate time to be responsive to any and all feedback and input on your
+**DEPR** ticket. Update the ticket and the proposal, if neccessary, with any
+information that should be captured from the ongoing feedback. Continue to
+iterate and do this until the announced target *Accepted* date.
 
--  Abandoned
+If during this time, there is a large amount of churn or concern, be open to
+adjusting the target dates and revisiting the proposal. If community alignment
+seems difficult, reach out to the `edX Architecture team`_ for directional
+guidance. In some cases, the proposal may need to be *Abandoned* entirely.
```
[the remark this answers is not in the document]
nasthagiri: In this initial phase of rolling out this process, I'm fine with concerns/issues being escalated to myself and the Arch team.  We can later update this as we see how it pans out.  Thanks.

oeps/oep-0021-proc-deprecation.rst
```diff
@@ -22,295 +22,323 @@ OEP-0021: Deprecation and Removal
...
 
+During this phase, remember the following:
 
-Timeline
---------
+* Implement the proposed and agreed upon migration path.
+* Remove related code from all places, including the frontend, APIs, and
+  the backend, perhaps even in that order.
+* Remove any related documentation on docs.edx.org_ and elsewhere.
+* Continue to update the ticket with any delays or issues that may arise.
```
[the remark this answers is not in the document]
nasthagiri: Good point.  I added the phrase "being mindful of the target *Removed* date set on the **DEPR** ticket."

PR #95, opened 2018-12-21
OEP-21: Recommend codebase impact review.

mattdrayer:
Based on recent breakage of the system due to removal of deprecated code, add a suggestion recommending a quick, simple codebase-wide impact analysis via GIthub's org+code search feature.

### DIFF
--- a/oeps/oep-0021-proc-deprecation.rst
+++ b/oeps/oep-0021-proc-deprecation.rst
@@ -45,10 +45,10 @@ development and developer onboarding.
 This is a problem shared by the software engineering industry and many have
 learned the benefits of investing in code removal:
 
-* It is undeniable that unnecessary code, like any other code, requires 
+* It is undeniable that unnecessary code, like any other code, requires
   maintenance over time. It costs time and money. [OREILLY]_
 
-* Extra code also makes it harder to learn the project, and requires extra 
+* Extra code also makes it harder to learn the project, and requires extra
   understanding and navigating. [OREILLY]_
 
 * It is harder to refactor, simplify, or optimize your program when it is bogged
@@ -62,7 +62,7 @@ learned the benefits of investing in code removal:
 * Dead code discourages a culture of treating the software as soft, and therefore
   always open to revision and improvement. [INFOQ]_
 
-Given the above, we define a common language and process to remove unneeded 
+Given the above, we define a common language and process to remove unneeded
 code. That code may be superseded by a new implementation or may be deemed no
 longer necessary and supported. Following this process guarantees alignment
 across all stakeholders and full execution.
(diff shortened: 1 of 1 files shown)


nasthagiri:
FYI @dianakhuang as arbiter.

oeps/oep-0021-proc-deprecation.rst
```diff
@@ -138,7 +138,9 @@ Analyze
 When proposing a removal, consider the following analysis:
 
 * Usage - Which users and services are currently using the code proposed for
-  removal on your own open edX instance?
+  removal on your own open edX instance?  Perform a quick search across the edX
+  codebase to gauge the level of impact and identify potential stakeholders.
+  https://github.com/search?q=org%3Aedx+sample&type=Code
```
smagoun: I suggest we add an additional step to the "Communicated" phase, which is to notify the stakeholders identified in this search. Seems kinda pedantic, but can't hurt.

mattdrayer:
@nasthagiri do we need to wait for approval by @dianakhuang before merging this change?

oeps/oep-0021-proc-deprecation.rst
mattdrayer: Good call -- I added some info to the `Communicated` section

dianakhuang:
From what I understand @mattdrayer , my role is to merge the PR once all the reviews and comments are in. I'll wait for @georgebabey to submit his review before merging.

oeps/oep-0021-proc-deprecation.rst
```diff
@@ -80,16 +80,18 @@ Process States
...
 As shown in the diagram above, the code removal process flows through the
 following states.
 
 *  Proposed_ - The starting state when removal is considered, analyzed, and
    documented.
-*  Communicated_ - The proposal is announced and its corresponding JIRA
-   ticket is updated with community feedback until the target Accepted date.
+*  Communicated_ - The proposal is announced to the general Open edX community
+   via the broadcast mechanisms defined below.  Specific notifications are
+   sent to all identified stakeholders.  The corresponding JIRA ticket
```
nasthagiri: _> Specific notifications are sent to all identified stakeholders._

Can we add a clause to the effect of "as best as possible"?  The real ideal solution for identifying and contacting all consumers of an API - is via our planned effort on Consumer-based Contracts.  Once we have consumer-based contract testing along with information of each consumer (internal and external), it will make identifying and contacting reliant consumers much more reliable.

In the meantime, a codebase search will have to do - but it won't catch everything - especially if consumers are outside of the `edx` Github organization or construct the API call in a way not anticipated by the deprecator.

PR #96, opened 2019-01-08
Add rationale for rST, and link to mdrst cheatsheet

nedbat:


nedbat:
@nasthagiri @robrap OK?

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -288,7 +288,24 @@ This section summarizes the requirements for long-term developer documentation.
 Format
 ~~~~~~
 
-We will use `Restructured Text (rST)`_ for all developer documentation.
+We will use `reStructured Text (rST)`_ for all developer documentation.
```
nasthagiri: Since the rationale was added in this section, can you add a back reference from the `Markdown Format` section to this one?   We can then remove the more shallow statement of `Here is an interesting blog post on..` in that section.

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -288,7 +288,24 @@ This section summarizes the requirements for long-term developer documentation.
...
+
+* rST scales up: it can be used for anything from single-page documents to
+  multi-volume book sets.
+
+* rST has a rich feature set.
```
nasthagiri: Can you add an example of a *useful* feature that's in rST but not in MD?

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -288,7 +288,24 @@ This section summarizes the requirements for long-term developer documentation.
...
+
+* rST is the standard markup format for Python project documentation.
+
+* rST does not suffer from incompatible implementations.
```
nasthagiri: Can you expand on what this means?  Like, it will look the same when rendered in one tool versus another?

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -302,6 +319,10 @@ Best practices
 Learning rST
 ^^^^^^^^^^^^
 
+* We have a `cheatsheet comparing rST to Markdown`__.
```
nasthagiri: One more thing: do you think it would be useful to have a back reference from the `Markdown Format` section back to this?  So folks reading about Markdown specifically know about our effort to create this cheatsheet?

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -319,7 +340,7 @@ Helpful rST-related tools
 
 * Other options are listed on stackoverflow_.
 
-.. _Restructured Text (RST): http://docutils.sourceforge.net/rst.html
+.. _reStructured Text (RST): http://docutils.sourceforge.net/rst.html
```
robrap: 1. `reStructuredText` is one word, as mentioned above.
2. While we are on the topic of Markdown, can we update the bullet point just above this (not yet in the PR) to:
```
* Pandoc_ can be used to convert between many formats, like Google Doc or Markdown to rST.
```

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -288,7 +288,24 @@ This section summarizes the requirements for long-term developer documentation.
 Format
 ~~~~~~
```
robrap: FYI: After this merges, I'd like to change the header underlines to match those in the cheatsheet, which also match the Python recommendations.  I know that rST allows for anything that is self-consistent in a doc, but when in the middle of a long doc, I don't like having to scroll and memorize the underlines to try to figure out what we are using for each heading.

nedbat:
@robrap I've updated based on your feedback.

oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -302,28 +327,37 @@ Best practices
 Learning rST
 ^^^^^^^^^^^^
 
+* If you are familiar with Markdown, here is a `cheatsheet comparing rST to Markdown`__.
```
robrap: Nit: I'd organize as follows.  If you like it, happy to do after merge as well.
```
* If you are familiar with Markdown:
  
  * Here is a `cheatsheet comparing rST to Markdown`__.

  * Here is `another comparison of Markdown and rST`__.

* See the `reStructuredText (rST) Primer`_.
```

PR #97, opened 2019-01-09
Update header formatting.

robrap:
Update header formatting to match Python rules and cheatsheet.
Possible consistency makes it simpler to understand section
depth at a glance in a long doc. Additionally, these header
characters give a visual queue of importance based on height.


robrap:
@nasthagiri: If you are good with this, I'd like to merge it before you start making any changes to avoid conflicts.
FYI: @nedbat

PR #100, opened 2019-01-23
OEP-0031: Micro-frontend Internationalization

albemarle:


oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Technology Choice
+=================
+
+We will use `react-intl <https://github.com/yahoo/react-intl>`_ to internationalize React components, in agreement with `Paragon ADR-2 <https://github.com/edx/paragon/blob/f5b9f876800e5154ef01d04f682b8cda21b3e439/docs/decisions/0002-react-i18n.rst>`_. react-intl is in line with current industry standards in i18n and satisfies all currently identified i18n needs, including multiple plurals.
+
+Alternatives Considered
+-----------------------
+
+An alternative choice we considered was `i18next <https://react.i18next.com/>`_. **[TODO]**
```
bderusha: For your consideration: https://facebookincubator.github.io/fbt/

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Locale
+======
+
+Localized content for new micro frontends should be at a URL that includes the language as the first part of the path after the domain. (Example: ``account.edx.org/es/``) The language string should be in the form of a `BCP 47 language tag <http://tools.ietf.org/html/rfc5646>`_. If the user enters the site at a URL that doesn't include a language string, we will determine a locale and then forward them to the corresponding URL. For consistency, this applies even if their language is English: ``account.edx.org/`` should be forwarded to ``account.edx.org/en/`` . We will use the most general, two-letter language codes to start with, and re-evaluate as necessary.
+
+When choosing what locale to use for a page, the locale in the URL always takes precedence. If there is none, check the user's cookies for a locale; if there isn't one, fetch it from the user's account settings and store it in their cookies for reuse. If there is no locale (or no locale we support) in the account settings, we will use the first locale we can support in the ``Accept-Language`` header that was sent with the request. Failing everything else, we will fall back to English.
```
bderusha: Is the plan to redirect users to the most appropriate URL path if not specified, or to render a language based on other heuristics at their un-locale-pathed URL?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Alternatives Considered
+-----------------------
+
+Previous i18n implementations have sometimes used the English text as the message id. While this saves developers a step in their work, it has the drawback that a change in the English copy invalidates all existing translations, even if they would not otherwise have needed to change. (For example, capitalizing a noun in English has no impact on the German translation, because nouns are capitalized in German anyway, or an English change might be correcting a typo that was correct in the translations.) Do not use English copy as the message id.
+
+Dates And Numbers
+=================
+
+All dates and numbers should be localized.
```
bderusha: Currency is often a [special](https://medium.com/dennis-nerush/localization-sounds-simple-but-it-is-not-part-2-supporting-multiple-currencies-204d5bf60e38) case.  We should have a documented opinion here for clarity.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+==========================
+
+Translate all non-display text that is present in a tag for accessibility reasons (e.g., ``aria-label``), using the same locale as the rest of the page.
+
+There is a special case where a message id doesn't have a translation in the user's preferred locale, so we fall back to a default locale. In this case, that string needs to be wrapped in a ``<span locale="....">`` tag to maintain the correct locale information for accessibility tools. In order to make sure that this string wrapping always happens, edX has written ``FMessage``, a simple wrapper around the standard react-intl ``FormattedMessage`` component, which will be the standard component used to format strings. ``FMessage`` can be found in the **[TODO]** repo.
+
+Message Ids
+===========
+
+Message ids, or message keys, uniquely identify a string to be internationalized. They should consist of one or more words to scope the message, plus one or more words to identify the message. All words will be lowercase and separated by periods. Here are examples of message ids used in a header and in a login modal, respectively:
```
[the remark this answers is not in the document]
albemarle: Different schemes might work best for different teams / components / pages.  We could standardize, but I'm not sure there's a big benefit, as long as a human can look at it and track it down.

oeps/oep-0031-arch-i18n.rst
albemarle: Yes, my intent in the paragraph above was to specify the "change the URL to what you think it should be" strategy.  Would it be clearer if that part of the earlier paragraph ("If the user enters the site...") came after this paragraph?

oeps/oep-0031-arch-i18n.rst
albemarle: Good point!

albemarle: And that does bring us back to country vs. language no matter how much I tried to escape that.  :)

oeps/oep-0031-arch-i18n.rst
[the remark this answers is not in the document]
albemarle: *nod*  @rock345  's team has a specific requirement to show a given locale regardless of the user's settings, which is how this came about.  Maybe there's a better way to do it?  Either way, we could certainly offer the user a small banner notification asking if they'd like to change to their preferred locale if we saw that they were on a URL that wasn't what was in their cookie.  That would cover the case of users getting a bookmark from a friend, as well.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Alternatives Considered
+-----------------------
+
+An alternative choice we considered was `i18next <https://react.i18next.com/>`_. **[TODO]**
+
+Locale
+======
+
+Localized content for new micro frontends should be at a URL that includes the language as the first part of the path after the domain. (Example: ``account.edx.org/es/``) The language string should be in the form of a `BCP 47 language tag <http://tools.ietf.org/html/rfc5646>`_. If the user enters the site at a URL that doesn't include a language string, we will determine a locale and then forward them to the corresponding URL. For consistency, this applies even if their language is English: ``account.edx.org/`` should be forwarded to ``account.edx.org/en/`` . We will use the most general, two-letter language codes to start with, and re-evaluate as necessary.
```
davidjoy: Chiming in without full context here, so forgive me if this is naive.  Does the two-letter language code inform number formatting in localizations, for instance?  Might be worth going straight to en_US, en_UK, en_CA etc... it feels like it's likely to become a stumbling block almost immediately.  If not determined by the URL path, by what criteria do we determine which country-specific localizations to use?

oeps/oep-0031-arch-i18n.rst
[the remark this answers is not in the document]
nasthagiri: Not giving precedence to the user's own settings does seem confusing.  Can we reference a product decision document or such that explains the rationale for this decision?

nasthagiri: If the user's preference doesn't exist, is it the frontend or the backend that fetches and stores the preference?  If it's the frontend, then we'll need to use another mechanism other than cookies - like [LocalStorage](https://www.w3schools.com/HTML/html5_webstorage.asp).

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+=================
+
+We will use `react-intl <https://github.com/yahoo/react-intl>`_ to internationalize React components, in agreement with `Paragon ADR-2 <https://github.com/edx/paragon/blob/f5b9f876800e5154ef01d04f682b8cda21b3e439/docs/decisions/0002-react-i18n.rst>`_. react-intl is in line with current industry standards in i18n and satisfies all currently identified i18n needs, including multiple plurals.
+
+Alternatives Considered
+-----------------------
+
+An alternative choice we considered was `i18next <https://react.i18next.com/>`_. **[TODO]**
+
+Locale
```
nasthagiri: Have we decided on what should happen when a user changes their language preference?  Whether and how does that change get propagated to other browsers and frontends?

oeps/oep-0031-arch-i18n.rst
nasthagiri: Nit: Can we use the term "namespace" rather than "scope" here?  In usage, a scope has generally referred to a region of a program.  Here, you are not enforcing any scoping rules but recommending usage of namespaces.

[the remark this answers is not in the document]
nasthagiri: When a single page is composed of multiple reusable components, is there a potential issue with collisions if 2 different components happen to use the same namespace?  Or is that not an issue if the translation process happens before the components are composed together?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+-----------------------
+
+Previous i18n implementations have sometimes used the English text as the message id. While this saves developers a step in their work, it has the drawback that a change in the English copy invalidates all existing translations, even if they would not otherwise have needed to change. (For example, capitalizing a noun in English has no impact on the German translation, because nouns are capitalized in German anyway, or an English change might be correcting a typo that was correct in the translations.) Do not use English copy as the message id.
+
+Dates And Numbers
+=================
+
+All dates and numbers should be localized.
+
+Right-To-Left Locales
```
nasthagiri: Will there be a future section here regarding deployment pipeline of translation files?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+   :local:
+   :depth: 2
+
+Motivation
+==========
+
+Internationalization (i18n) is a vital part of unlocking content for users around the globe and effectively supporting our business partners. In order to help us reuse localized components from app to app, the different parts of the edX codebase should align on a consistent handling of i18n. This document covers our approach to internationalizing new front-end applications and components written in React.
+
+In this OEP we use the terms locale and language more or less interchangeably. However, note that this document addresses primarily making code multilingual, i.e., the issue of presenting the same content in different languages. Considering ways to make it multi-regional, to allow for presenting different content for different countries, is out of scope for this OEP.
+
```
rock345: Note: Our current marketing site header supports this functionality.  There is currently active work happening on the shared footer to support this.  I know this is why this is being called out as not supported but wanted to note the reason why.

oeps/oep-0031-arch-i18n.rst
rock345: If our preference is two character lang we should make sure we have a common function to convert en-US, en-GB to en.  I think when we are talking China we are going to have to make decisions re: 5 character codes zh for traditional and simplified as supporting both maybe an ask

oeps/oep-0031-arch-i18n.rst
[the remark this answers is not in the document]
rock345: For the marketing site (or any site that support anonymous browsing) they should be shown /lang-code/ if that is the URL they have arrived at.  This should not redirect to /en or another language regardless of their user preference

marcotuts: Thoughts here on how this impacts shared urls? If I share a url to an edX page that includes my locale and someone with a Spanish language clicks the link, it sounds like the url precedence rule would mean they would be switched over to English instead of being redirected to the Spanish version of the page (which is their default / account setting)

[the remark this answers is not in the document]
marcotuts: moving my comment into this thread now that I see it has been discussed already: 

"When choosing what locale to use for a page, the locale in the URL always takes precedence."

Thoughts here on how this impacts shared urls? If I share a url to an edX page that includes my locale and someone with a Spanish language clicks the link, it sounds like the url precedence rule would mean they would be switched over to English instead of being redirected to the Spanish version of the page (which is their default / account setting)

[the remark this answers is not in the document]
albemarle: @marcotuts Your last paragraph correctly describes the current proposal.  I am happy to give you and @rock345 giant q-tips and have you battle it out.  :-)

image

oeps/oep-0031-arch-i18n.rst
albemarle: Do you mean from a language switcher widget, as opposed to changing it permanently in their user settings?

oeps/oep-0031-arch-i18n.rst
bderusha: It is not clear to me that this proposal recommends redirecting to a new URL based on this logic.  If we could make that more explicit that would be great.  It sounds like that is your intent and I agree with that as a proposal.

oeps/oep-0031-arch-i18n.rst
bderusha: Yes, as we've discussed, localization is more than strictly language.  It's language, context, formatting, etc.

I think the big questions here I'd love for us to have an opinion on are: 
- Should currencies and value be linked ( 1,00 USD for `fr_fr` because it's French, or 1.00 USD for `fr_fr` because it's dollars).  IMHO the former is better, but the argument can be made for either.
- Should we localize currency (1.00 USD ~ 1.80 GBP for `en_gb`).  IMHO we should explicitly not localize currency in this way as part of this OEP.  It is potentially a nice feature, but it is a feature so long as you are still paying in USD in the end.

oeps/oep-0031-arch-i18n.rst
nasthagiri: No, I didn't mean to suggest introducing a new way for users to change their language settings.  Question is what the behavior is cross-browser(-tab) and frontend when a user changes their language setting.  The answer to this may indicate requirements for invalidating caches/cookies/etc.

oeps/oep-0031-arch-i18n.rst
matthugs: It looks like that takes us out of scope for the oep: https://github.com/edx/open-edx-proposals/pull/100/files#diff-00db08b214255fa07b73c64617cebe2dR35 (although I'm worried we may want to rename the oep to specify that it's not about Internationalization writ large, given that omission).

oeps/oep-0031-arch-i18n.rst
matthugs: It may be considered unclear whether this applies to phone numbers in addition to counting numbers. Is it unacceptable scope creep to suggest that we should always give phone numbers in the format of `+<country code> <number with spaces>` without the American-traditional dashes and parentheses?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,122 @@
...
+.. contents::
+   :local:
+   :depth: 2
+
+Motivation
+==========
+
+Internationalization (i18n) is a vital part of unlocking content for users around the globe and effectively supporting our business partners. In order to help us reuse localized components from app to app, the different parts of the edX codebase should align on a consistent handling of i18n. This document covers our approach to internationalizing new front-end applications and components written in React.
+
+In this OEP we use the terms locale and language more or less interchangeably. However, note that this document addresses primarily making code multilingual, i.e., the issue of presenting the same content in different languages. Considering ways to make it multi-regional, to allow for presenting different content for different countries, is out of scope for this OEP.
```
schenedx: I believe because this is specifically talking about i18n standard for frontend application implemented by reactjs, we should also call out those application using other frontend framework, or web pages rendered by django mako templates, as out of scope. We should make this explicit call out because some of the standards within this document, i.e. Message IDs, could also apply to those implementations.

oeps/oep-0031-arch-i18n.rst
schenedx: @nasthagiri There are multiple ways user can change their language preference. Are you talking about user changing their language preference in Browser? Or user changes their language preference from their user profile edX page? Or a combination of them? If we call out the way we recognize the locale in a prioritized list, then we can infer the possibilities when user switch their language preference at certain places.

oeps/oep-0031-arch-i18n.rst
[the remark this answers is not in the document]
schenedx: @albemarle I think everyone here is looking for the reason behind why the locale in the URL take precedence over user account setting value. If there is no good reason for this priority, why can't we have user account setting value take higher precedence?

oeps/oep-0031-arch-i18n.rst
nasthagiri: @schenedx Good point.  For this particular design question, I was asking about when a user changes their persisted language preference in their edX account.  They may change it in a dropdown in the edX footer/header, in their profile page, or whatever other UX we support in the future.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,208 @@
...
+**********
+react-intl
+**********
+
+We will use `react-intl <https://github.com/yahoo/react-intl>`_ to internationalize React components, in agreement with `Paragon ADR-2 <https://github.com/edx/paragon/blob/master/docs/decisions/0002-react-i18n.rst>`_. react-intl is in line with current industry standards in i18n and satisfies our currently identified i18n needs, including multiple plurals. In order to add some browser shims, the Architecture team has published a thin wrapper around react-intl in the `frontend-i18n repo <https://github.com/edx/frontend-i18n>`_. For consistency, use this wrapper, instead of importing react-intl directly.
+
+Alternatives Considered
+=======================
+
+An alternative choice we considered was `i18next <https://react.i18next.com/>`_. It comes with plug-and-play functionality to determine a user's locale, babel scripts for various translation management schemes, etc. However, it is fairly large, and doesn't seem to offer compelling benefits that would outweigh the fact that studio-frontend and Paragon already use react-intl.
```
abutterworth: Paragon doesn't have an i18n library FYI.

oeps/oep-0031-arch-i18n.rst
albemarle: I think the existing HOWTO is a better place for that.

oeps/oep-0031-arch-i18n.rst
albemarle: D'oh!  I think that was just where the original react-intl ADR was.  Thanks!

oeps/oep-0031-arch-i18n.rst
albemarle: These are all really good points.  I've punted discussion of locale codes to a future update to this OEP.

oeps/oep-0031-arch-i18n.rst
albemarle: Good question!  I'm not sure we have a decision right now, but if one comes up, we will add it in.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,210 @@
...
+Right-To-Left Locales
+=====================
+
+For languages that are written right-to-left (RTL), such as Arabic and Hebrew, you will need to add a ``dir`` attribute to your ``html`` tag, and your CSS must swap graphics and margins. The frontend-i18n library will make this straightforward; see the `i18n HOWTO <https://github.com/edx/frontend-i18n/blob/master/docs/how_tos/i18n.rst>`_ for more information.
+
+*****************
+Locale Resolution
+*****************
+
+For pages meant to be seen by logged-in users, we will check the user's cookies for a locale; if there isn't one, we will fetch it from the user's account settings and store it in their cookies for reuse. If there is no locale (or no locale we support) in the account settings, we will use the locale that the user's browser is set to. Failing everything else, we will fall back to English.
```
schenedx: I saw that the `URLs` section mentioned about adding the locale to the URL, how does that affect locale resolution here? Why is that mechanism not mentioned?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,210 @@
...
+Message Ids
+===========
+
+Message ids, or message keys, uniquely identify a string to be internationalized. They should consist of one or more words to namespace the message, plus one or more words to identify the message. All words will be lowercase and separated by periods. Here are examples of message ids used in a header and in a login modal, respectively:
+
+  | ``header.search``
+  | ``login.no.such.password``
+  | ``login.privacy.policy``
+
+The part of the id used for namespacing is purely to reduce collisions and help human readability; it's not intended to be programmatically parsable. That said, a helpful convention is to make it the name of your application or repo. Note that if you end up with a duplicate message id in the same translation project in Transifex, and they have different content, one of them will "win" when you download your translated strings, leading to an out-of-place string in your application.
```
schenedx: This is the first mention of `Transifex`. Should we have a section just to talk about about translation service vendor before this section? In the future, there might be other translation service vendors, and how can edX integrate with those?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,210 @@
...
+
+Alternatives Considered
+-----------------------
+
+Previous i18n implementations have sometimes used the English text as the message id. While this saves developers a step in their work, it has the drawback that a change in the English copy invalidates all existing translations, even if they would not otherwise have needed to change. (For example, capitalizing a noun in English has no impact on the German translation, because nouns are capitalized in German anyway. Or an English change might be correcting a typo that was correct in the translations.) Do not use English copy as the message id.
+
+Cross-Site Scripting Vulnerability
+==================================
+
+Because translated strings often contain placeholders that will be filled in at render time, possibly using user input, localization is a potential opening for cross-site scripting (XSS) attacks. The react-intl functions exposed by frontend-i18n are be XSS-safe.
```
schenedx: "The react-intl functions exposed by frontend-i18n are `to` be XSS-safe."

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,210 @@
...
+Missing Translations
+====================
+
+Missing translations can be a disorienting experience for a user. We strongly suggest a careful evaluation of translation coverage, or a clear indicator of beta status, before making a new language available to the public.
+
+****************
+Server-Side i18n
+****************
+
+Have your server send symbolic values or error codes, rather than translated display text, to convey information, and then localize on the front end. This keeps all of your localization together, simplifies testing your server code, and makes your server response reusable by other front ends, which may wish to display the same information but with different text.
```
schenedx: Is this section really just mean we should localize everything on the backend server, and have no logic on the javascript side to do translation? If my understanding is correct, I feel the existing way to phrase is confusing.

albemarle: @schenedx No, the opposite.  The server should provide programmatic information only, and leave localization to the front end.

oeps/oep-0031-arch-i18n.rst
albemarle: @nasthagiri Is this still a concern for you for this version of this OEP?  The current behavior is that other open tabs will just not see the change until the next page refresh, and to my knowledge changing that is not a high priority.

oeps/oep-0031-arch-i18n.rst
albemarle: Will do!

oeps/oep-0031-arch-i18n.rst
albemarle: Good thought, I'll add a brief mention earlier.

albemarle:
> I like this version much better than the last one I reviewed.
> I was hoping you can clarify the difference between `localized` and `translated`. If you are doing that interchangeably, please call it out.

@schenedx They are not quite interchangeable, no.  You localize an application by translating all the strings.  You can also localize a number, for instance by changing from "1,900" to "1.900".

Is there a particular sentence where you thought the usage could be confusing?

oeps/oep-0031-arch-i18n.rst
schenedx: I see. I think a graphical representation of what this mechanism is may bring more clarity. Like draw two boxes, one the backend server, one the frontend. Then have the non-localized data transfered to front end. And have the localized data shown to users.

schenedx:
@albemarle Since you believe there is a difference between `Translated` vs `Localized`. I think we should also call it out and insert your explanation into the document just below the paragraph where you explain the difference between `internationalized` vs `localized`.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
+###################################################
+OEP-0031: Internationalization For React Front Ends
+###################################################
+
+.. list-table::
+
+   * - OEP
+     - :doc:`OEP-0031 <oep-0031-arch-i18n.rst>`
+   * - Title
+     - Internationalization
```
nasthagiri: Reminder to update this title to match the title above.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Not Everything Is English
+=========================
+
+Use longer strings with placeholder variables, rather than trying to create messages by concatenating two or more short messages together. Even if it looks like it's working in English, it won't necessarily be grammatically correct in other languages. This includes uses such as a pagination display: you cannot translate only the word ``of`` and then glue together ``{number} + {translation of "of"} + {number}``. Instead, your string for translation should be ``{page_number} of {total_pages}``, so your translators can move the placeholders into whatever position is required for their language.
+
+When creating forms, keep in mind that there are many different ways of writing an address besides "street number and street name" as one field, and city and state as two others. Also, not every culture puts given name and family name in that order. If possible, allow free form, unstructured addresses, name fields, etc.
```
nasthagiri: Nit: Does this paragraph contain examples of different cultures and "not everything is English"?  That is, the name and address examples are relevant even for English-speaking countries.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Accessibility
+=============
+
+Translate all non-display text that is present in a tag for accessibility reasons (e.g., ``aria-label``), using the same locale as the rest of the page.
+
+Dates And Numbers
+=================
+
+All dates and numbers should be localized.
```
nasthagiri: Is there an example of how to localize these?  Or a link to a more detailed best practice or an example?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Currency
+========
+
+When displaying a price in a foreign (to the user) currency, localize the number according to the user's locale. That is, for a user in France viewing a price in U.S. dollars, show it as ``USD 19,99``, using the ``fr`` convention for the decimal point.
```
nasthagiri: Here too, perhaps a link to an example or a link to a more detailed how-to, would be useful.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+We considered but rejected a fallback step where we would choose a language based on the geographic location of the user's IP address. This is generally not reliable: many countries have speakers of multiple languages, users might be traveling, and users might be living or studying in a country but still be more comfortable in a different language.
+
+****
+URLs
+****
+
+For pages intended to be viewable by the public (including search engines), we will use path prefixing to identify the locale of the page being shown. This means that the URL will include the locale as the first part of the path after the domain. (Example: ``account.edx.org/es/``) This language code should be in the form of a `BCP 47 language tag <http://tools.ietf.org/html/rfc5646>`_.
+
+ If the user enters the site at a URL that doesn't include a language string, we will determine a locale and then forward them to the corresponding URL. For consistency, this applies even if their language is English: ``account.edx.org/`` should be forwarded to ``account.edx.org/en/`` .
```
nasthagiri: Consider linking "determine a locale" to the `Locale Resolution` section above.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+When displaying a price in a foreign (to the user) currency, localize the number according to the user's locale. That is, for a user in France viewing a price in U.S. dollars, show it as ``USD 19,99``, using the ``fr`` convention for the decimal point.
+
+Right-To-Left Locales
+=====================
+
+For languages that are written right-to-left (RTL), such as Arabic and Hebrew, you will need to add a ``dir`` attribute to your ``html`` tag, and your CSS must swap graphics and margins. The frontend-i18n library will make this straightforward; see the `i18n HOWTO <https://github.com/edx/frontend-i18n/blob/master/docs/how_tos/i18n.rst>`_ for more information.
+
+*****************
+Locale Resolution
```
nasthagiri: To distinguish from the `URLs` section, consider renaming this section to `Locale Resolution for logged-in pages`.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+For pages meant to be seen by logged-in users, we will check the user's cookies for a locale; if there isn't one, we will fetch it from the user's account settings and store it in their cookies for reuse. If there is no locale (or no locale we support) in the account settings, we will use the locale that the user's browser is set to. Failing everything else, we will fall back to English.
+
+Alternatives Considered
+=======================
+
+We considered but rejected a fallback step where we would choose a language based on the geographic location of the user's IP address. This is generally not reliable: many countries have speakers of multiple languages, users might be traveling, and users might be living or studying in a country but still be more comfortable in a different language.
+
+****
+URLs
```
nasthagiri: To distinguish from the `Locale Resolution` section, consider renaming this section to `URLs and Locale Resolution for public pages`.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Alternatives Considered
+-----------------------
+
+Previous i18n implementations have sometimes used the English text as the message id. While this saves developers a step in their work, it has the drawback that a change in the English copy invalidates all existing translations, even if they would not otherwise have needed to change. (For example, capitalizing a noun in English has no impact on the German translation, because nouns are capitalized in German anyway. Or an English change might be correcting a typo that was correct in the translations.) Do not use English copy as the message id.
+
+Cross-Site Scripting Vulnerability
+==================================
+
+Because translated strings often contain placeholders that will be filled in at render time, possibly using user input, localization is a potential opening for cross-site scripting (XSS) attacks. The react-intl functions exposed by frontend-i18n are XSS-safe.
```
nasthagiri: Can you clarify - is there anything further that a developer needs to do regarding XSS-safety - if they use `frontend-i18n`?  My reading of this indicates no, but I think it could be made more explicit.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+When a server response can contain one of a set of enumerated values, it can be tempting to assemble your message id on the fly by concatenating a common prefix and the server-provided value:
+
+``const id = 'myapp.errors.${errorCode}';``
+
+However, an unexpected value from the server can cause this technique to generate a message id that doesn't exist. If you choose to do this, it's best to whitelist your server-provided value against a list of expected responses.
+
+Missing Translations
+====================
+
+Missing translations can be a disorienting experience for a user. We strongly suggest a careful evaluation of translation coverage, or a clear indicator of beta status, before making a new language available to the public.
```
nasthagiri: Consider linking to the Beta Language feature as an example of how we are already doing this in the platform.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,212 @@
...
+Reference Implementation
+************************
+
+The reference implementation of this OEP is the Order History app in `frontend-app-ecommerce <https://github.com/edx/frontend-app-ecommerce>`_, which uses the `frontend-i18n <https://github.com/edx/frontend-i18n>`_ library.
+
+###########
+Future Work
+###########
+
+There is a special accessibility use case where a message id doesn't have a translation in the user's preferred locale, so we fall back to a default locale. In this case, the string would ideally be wrapped in a ``<span locale="....">`` tag to maintain the correct locale information for accessibility tools. In order to make sure that this string wrapping always happens, studio-frontend uses ``WrappedMessage``, a simple wrapper around the standard react-intl ``FormattedMessage`` component.
```
nasthagiri: Can you clarify?  Does this mean that `studio-frontend` handles this accessibility case, but `frontend-i18n` doesn't yet?

nasthagiri:
@albemarle @davidjoy @sarina Can we move forward on pushing this OEP through? 
There were questions raised today by a team on MFE i18n - and it would be good to have a published reference.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,270 @@
...
+     -
+   * - Arbiter
+     - Simon Chen <schen@edx.org>
+   * - Status
+     - Draft
+   * - Type
+     - Architecture
+   * - Created
+     - 2019-01-07
+   * - Review Period
```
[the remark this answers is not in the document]
davidjoy: Think you just pick a reasonable timeframe. 

https://github.com/edx/open-edx-proposals/blob/master/oeps/processes/oep-0001.rst#oep-header-preamble

"The Review Period header specifes the target dates for reviewing the OEP, as agreed by the Authors and Arbiter. The recommended duration of the review is 2 weeks. However, if the review exposes areas of the proposal that need further discussion and fleshing out, then the Arbiter may choose to extend the review period."

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,270 @@
...
+
+HTML Considerations
+===================
+
+Set the language of your document with the ``lang`` attribute on your ``html`` tag. Use the UTF-8 encoding, and declare it with a ``meta`` tag in the head of your HTML document. Prefer using characters (``á``) rather than character entities (``&aacute;``) or escapes (``&#225;``).
+
+Grammar and Cultural Considerations
+===================================
+
+Use string interpolation and placeholder variables instead of creating messages by concatenating two or more short messages together. Even if it looks like it's working in English, it won't necessarily be grammatically correct in other languages. This includes uses such as a pagination display: you cannot translate only the word ``of`` and then glue together ``{number} + {translation of "of"} + {number}``. Instead, your string for translation should be ``{page_number} of {total_pages}``, so your translators can move the placeholders into whatever position is required for the grammar rules of their language.
```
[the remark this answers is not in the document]
davidjoy: Looking at the How-To, I'm not seeing this sort of example in there; it's more about how to set up i18n in general, with none of the guidelines and best practices in this document.  

IMHO the example here is pretty clear; I'm sort of inclined to leave this part as-is, since a link to the How-To won't add much more clarity.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,270 @@
...
+
+Cross-Site Scripting Vulnerability
+==================================
+
+Because translated strings often contain placeholders that will be filled in at render time, possibly using user input, localization is a potential opening for cross-site scripting (XSS) attacks. The ``react-intl`` functions exposed by frontend-platform are XSS-safe.
+
+We explicitly don't export ``react-intl`` components that are not XSS safe, such as ``FormattedHTMLMessage``.  See `Don't Use FormattedHTMLMessage` <https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/preventing_xss/preventing_xss_in_react.html#don-t-use-formattedhtmlmessage>`_ for more information.
```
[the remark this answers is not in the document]
davidjoy: After reviewing what they changed in more recent versions, I think this section is technically correct and a useful warning.

The change in more recent versions of react-intl is to deprecate FormattedHTMLMessage in favor of a new form of FormattedMessage which allows you to explicitly opt-in to HTML Parsing: https://formatjs.io/docs/react-intl/upgrade-guide-4x/#breaking-api-changes

After we upgrade react-intl this section will be _meaningless_ rather than _wrong_, which I think is acceptable until we come back and remove it. :)

oeps/oep-0031-arch-i18n.rst
davidjoy: Ugh.  So Beta Languages only manifests in the old account settings view... which means we've effectively blown away the feature as we permanently migrate to the account settings MFE.

Since edx.org has been running the account settings MFE in production for the last two years, Beta Languages seems like a candidate for DEPR rather than something we should include here.  Thoughts?

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,270 @@
...
+Currency localization
+*********************
+
+We should correctly localize currency, which is an issue of country rather than language, and reliant on having an accurate currency conversion service.  This is beyond the scope of our i18n libary.
+
+**********************
+Consistent terminology
+**********************
+
+We should settle on a consistent set of language and locale codes across edX.
```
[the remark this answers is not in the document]
davidjoy: I'm inclined to remove this "Consistent terminology" section; not sure it's adding anything at the moment.  I'm not even sure what about it all is inconsistent.

oeps/oep-0031-arch-i18n.rst
```diff
@@ -0,0 +1,270 @@
...
+Default platform locale as a final fallback
+*******************************************
+
+As noted above in the Locale Resolution on Pages for Authenticated Users section, today we fall back to 'en' if we are unable to determine what language to use for a given authenticated user.  We would prefer this to match the server-side i18n behavior of falling back to the default platform locale, rather than assuming English.
+
+***********************
+Beta languages for MFEs
+***********************
+
+Micro-frontends have a hard-coded list of available languages with no concept of beta languages like the server-side i18n platform.
```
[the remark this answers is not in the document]
davidjoy: Oh, wow, is that in the account settings MFE!?  I stand corrected, I guess we do still support it.

oeps/oep-0031-arch-i18n.rst
davidjoy: Apparently I'm wrong, seeing your screenshot below.

sarina: yeah. I think there's value in leaving these beta lang discussions open in case another reviewer can provide any more insight

sarina:
@davidjoy I'm going to open up the review period through November 15th. It all looks good to me, in the meantime it'd be good to get RTD building.

davidjoy (approved):

sarina (approved):

davidjoy:
Note: We need to move the OEP's RST file.  This will blow away our PR comment history.  To mitigate this, we're going to close this PR and open a new one and perform the move there.  The new PR will just link back to this one.  Hopefully that preserves our history.

PR #102, opened 2019-01-28
Remove gemnasium step from OEP-0014

michaelyoungstrom:
I was going through these docs to archive a tools repo and noticed it still references Gemnasium in the steps. Gemnasium is no longer an active service, so we should remove that from the docs (https://docs.gitlab.com/ee/user/project/import/gemnasium.html).

It should probably stay in the motivation/ rejected alternatives, as it was still a contributing factor to this OEP.

### DIFF
--- a/oeps/oep-0014-proc-archive-repos.rst
+++ b/oeps/oep-0014-proc-archive-repos.rst
@@ -93,14 +93,6 @@ Archive Steps
 3. Create an IT help ticket to update the description of the repository to
    begin with ``[ARCHIVED]``.
 
-4. Create a DevOps ticket to change the status of the repo to "non-monitored"
-   in Gemnasium, and to remove any integration or webhooks that DevOps may be
-   maintaining.
-
-In the future, the step of changing the monitored status of a repository in
-Gemnasium could be automated using Gemansium APIs and keying off of the
-``archived`` value in openedx.yaml.
-
 README Archive Statement
 ------------------------
 Include this statement in the README.rst file:


nasthagiri:
@michaelyoungstrom FYI -  I'm in search of a new arbiter for this OEP for you.  Stay tuned.

oeps/oep-0014-proc-archive-repos.rst
```diff
@@ -93,14 +93,9 @@ Archive Steps
 3. Create an IT help ticket to update the description of the repository to
    begin with ``[ARCHIVED]``.
 
-4. Create a DevOps ticket to change the status of the repo to "non-monitored"
-   in Gemnasium, and to remove any integration or webhooks that DevOps may be
+4. Create a DevOps ticket to remove any integration or webhooks that they may be
```
feanil: I think you can just remove this step until we add a new security integration.  As of right now, we're getting notifications via Github automatically so there is nothing for devops to maintain.  In the future I believe the tools team is considering looking at requires.io but that tooling doesn't exist right now.

oeps/oep-0014-proc-archive-repos.rst
```diff
@@ -93,14 +93,6 @@ Archive Steps
...
    begin with ``[ARCHIVED]``.
 
-4. Create a DevOps ticket to change the status of the repo to "non-monitored"
-   in Gemnasium, and to remove any integration or webhooks that DevOps may be
-   maintaining.
-
-In the future, the step of changing the monitored status of a repository in
-Gemnasium could be automated using Gemansium APIs and keying off of the
-``archived`` value in openedx.yaml.
-
```
nasthagiri: FYI @jmbowman @dianakhuang

PR #103, opened 2019-01-28
OEP-32: Unique Identifier for Users

robrap:


oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -0,0 +1,153 @@
...
+As noted in the `User Identifier Candidate Comparison`_ above, using the LMS user_id has the following two drawbacks:
+
+* Information leakage of auth_user table with easily guessable values.
+
+* The id is tied down to implementation of the housing database.
+
+Cost of an alternative to the LMS user_id
+-----------------------------------------
+
+If we were to try to eliminate the drawbacks of using the auto-incremented LMS user_id as our unique identifier, an alternative might be to introduce a new id, like a UUID (Universally Unique IDentifier), that did not have the `Drawbacks of the LMS user_id`_.
```
bradenmacdonald: What about sticking with the existing `user_id` column, but changing the insertion algorithm to generate a random user ID instead of a sequential one for new users? Obviously that wouldn't help existing users, but that would significantly reduce this drawback for new ones as well as new Open edX installations.

bradenmacdonald:
Interesting... Would this change, once fully implemented, also open the door to allowing users to change their usernames at any time? That would be a nice improvement.

oeps/oep-0032-arch-unique-identifier-for-users.rst
robrap: Hi @bradenmacdonald. I had thought of this, but I was under the understanding that the column and all foreign-keys across the entire system would need to change size to binary(16). That doesn't seem small.  And it ignores the fact that you'd have additional questions about when and where to use the human readable form of the UUID, and other complications of dealing with proper indexing, etc.  Thoughts?

robrap:
Thanks @bradenmacdonald.  I'm not holding my breath for when this change would be fully implemented, but I guess anything is possible.

oeps/oep-0032-arch-unique-identifier-for-users.rst
bradenmacdonald: The column is currently a 32-bit integer. I agree that changing it to a 64-bit int or a 128-bit UUID would be prohibitively difficult (due to the foreign keys and the size of auth_user; you'd need to use pt-online-schema-change for the large tables and it would take ages).

However, what I'm suggesting is just keep the schema unchanged (ID remains a 32-bit integer) but when there's a new user, assign them a random 32-bit integer as their ID, rather than the next integer in the auto_increment sequence. (Plus some logic to deal with conflicts so insertions never fail).

robrap: I see.  Nothing in this OEP prohibits this idea if someone wants to take that on.  The major disadvantage for UUIDs is a potential performance issue with the index because the values are not sorted.  There are solutions like [rearranging the time-related bits](https://mysqlserverteam.com/mysql-8-0-uuid-support/) in the MYSQL solution.  Someone would have to find the right way to make something like this work with 32-bits for it to ever work for edx.org.

I could acknowledge this idea, but I don't want to over explain it and I don't want to seem like anything is getting promised in this OEP along that front.

robrap:
Note to self: An exception should probably be noted for the User Profile API that will continue to be based on username and should **not** include the LMS user_id.
FYI: @nasthagiri

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -0,0 +1,153 @@
...
+
+
+Context
+=======
+
+In 2016, the analytics team wrote a decision to use the *LMS user_id* in `Using consistent user identifiers in Segment events`_.
+
+In addition to needing a consistent user identifier for internal events, we also need one for public events, REST APIs and JavaScript APIs.
+
+.. _Using consistent user identifiers in Segment events: https://openedx.atlassian.net/wiki/spaces/AN/pages/144441849/Using+consistent+user+identifiers+in+Segment+events
```
pdpinch: Is it possible to make this wiki page public?

robrap: All necessary information exists in this OEP, so I killed the pointer to Confluence, since it pointed to other docs, and I didn't want to deal with all of that. :)

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -0,0 +1,125 @@
...
+     - 2019-01-25
+   * - `Review Period`
+     - 2019-04-15 - 2019-04-29
+
+Decision
+========
+
+We will use the *LMS user_id* to uniquely identify a user in the Open edX system. The *LMS user_id* is the ``id`` value of the user's row in the `Django auth_user`_ table in the Open edX LMS.
+
+The *LMS user_id* should be used for all internal and public events, REST APIs, and JavaScript APIs.
```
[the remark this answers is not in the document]
robrap: Yes. This can be updated to ‘can’.

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -0,0 +1,125 @@
...
+
+
+Consequences
+============
+
+A consequence of the decision to use the *LMS user_id* as the unique user id is that some additional changes are needed to make the *LMS user_id* more globally available across all services:
+
+* The `LMS OAuth Dispatch App`_ will now include the *LMS user_id* in the JWTs (JSON Web Tokens) used as part of the `Open edX Authentication`_ upon request.
+
+* Additionally, some further investment is required to make the *LMS user_id* available to all services for all users, like the Ecommerce Service.
```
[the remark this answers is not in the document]
robrap: Yes. We are a lot further, but there is additional work that I think the Revenue team plans to pick up. At some point, we hope this will be resolved in all cases.

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -98,6 +98,8 @@ A consequence of the decision to use the *LMS user_id* as the unique user id is
 
 * The `LMS OAuth Dispatch App`_ will now include the *LMS user_id* in the JWTs (JSON Web Tokens) used as part of the `Open edX Authentication`_ upon request.
 
+* When exposing user information to other users, like in the case of a public Profile page, the associated *LMS user_id* should not be shared with other users.
```
robrap: @nasthagiri: FYI: I added this warning text.  Not sure if it will do anything, but maybe it will help someone think about this somewhere it is needed.

oeps/oep-0032-arch-unique-identifier-for-users.rst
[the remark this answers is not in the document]
robrap: @nasthagiri: I updated the language, but split out internal events, since that was made as a separate decision already.

oeps/oep-0032-arch-unique-identifier-for-users.rst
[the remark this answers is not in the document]
robrap: Leaving for now.  We can update later when this is no longer needed.

oeps/oep-0032-arch-unique-identifier-for-users.rst
robrap: @bradenmacdonald: I went for simple on the update, so I did not mention this.  As noted, I don't think the decision states anything against this possibility.

oeps/oep-0032-arch-unique-identifier-for-users.rst
kdmccormick: Could you elaborate on the rationale for adding this warning text? It seems like one's *LMS user_id* is equally as sensitive their username, and, as such, should be considered essentially public.

oeps/oep-0026-arch-realtime-events.rst
```diff
@@ -130,32 +130,12 @@ For details on integrating with Caliper, please see the `Caliper Real-time Event
 
 .. _Caliper Real-time Events: oep-0026/caliper-realtime-events.rst
 
-Anonymized User ID
-==================
+Unique User ID
+==============
 
-In order to uniquely identify a user in the Open edX system with a long-term and non-revealing identifier, an *Anonymized User ID* will be generated as a  SHA-256 hash of the Open edX user_id. The Open edX user_id is the **id** value of the user's row in the `Django auth_user`_ table in the Open edX LMS.
+The *LMS user_id* will be used to uniquely identify a user in the Open edX system. This decision is detailed in `OEP-31: Unique Identifier for Users`_.
```
felipemontoya: Should this be OEP-32?

robrap: Good catch.  Fixed.

pdpinch:
I feel like there's an unstated judgement here about the anonymous user id. I appreciate the specific call out about LTI and the judgement that the LMS User ID _could_ be used instead of the anonymous user id in the future. 

Is it too much of a digression to make a similar statement about the very common practice of using the anonymous user id in Qualtrics surveys (via the [%%USER_ID%% keyword](https://github.com/edx/edx-platform/blob/master/common/djangoapps/util/keyword_substitution.py)).

robrap:
@pdpinch: Thanks.  As you point out, LTI is just one of many possible examples used to try to clarify the point of the OEP.

I don't know the details of all past decisions.  For the Qualtrics surveys, or any other use of the anonymous user id, I hope the OEP is clear that the _LMS user_id_ is an acceptable solution unless there is an additional requirement for the use case that is not addressed in this OEP.

1. I think it is perfectly valid to use this OEP to challenge if _LMS user_id_ could now be used in a place where it wasn't previously.
2. For any such use case, there is probably some digging needed to ensure that there were no additional requirements for that use case outside of the scope of this OEP.
3. Finally, a more local decision (possibly ADR) could be written that refers to this OEP and documents why and where a switch might be warranted.  Note: if an ID is being switched, there would be separate questions around when and where that can be done without causing backward compatibility issues, and that could all be addressed in an ADR (whether the final decision is to switch or double-down on a past decision not to use the _LMS user_id_ for a certain use case).

Let me know if this makes sense and/or whether you think some of this response is important to have in the OEP itself regarding process and legacy.

oeps/oep-0032-arch-unique-identifier-for-users.rst
robrap: @kdmccormick: The `LMS user_id` is not considered PII, so it is not equally sensitive.  Given that, this note was updated and you can let me know if it is more clear.  Thanks.

kdmccormick: Ah, I see. It's the _association_ that we don't want to make public. Makes sense.

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -0,0 +1,127 @@
...
+
+* Additionally, some further investment is required to make the *LMS user_id* available to all services for all users, like the Ecommerce Service.
+
+This decision is also a simple acknowledgement of our tie to the database implementation and its auto-incremented value, given its already pervasive usage.
+
+The *LMS user_id* is also considered a safe and anonymous option for third-party integrations as described above. One example of a third-party integration where the *LMS user_id* can now safely be used is in the `Realtime Events API`_, which has been updated with this clarification.
+
+Note: The legacy implementation of LTI uses what is called an "anonymous id", but is no more "anonymous" than the *LMS user_id* from a PII perspective.  It uses an id made of a combination of user id and course id, which makes it impossible for a third party system to build a user model in its own system across courses, and limits the capabilities a third-party system might offer a user. For future LTI implementations, this decision simply states that the *LMS user_id* is an option unless these additional limitations are a requirement of the integration.
```
pdpinch:
```suggestion
Note: The legacy implementations of LTI and the %%USER_ID%% keyword uses what is called an "anonymous id", but is no more "anonymous" than the *LMS user_id* from a PII perspective. They use an id made of a combination of user id and course id, which makes it impossible for a third party system to build a user model in its own system across courses, and limits the capabilities a third-party system might offer a user. For future LTI implementations, this decision simply states that the *LMS user_id* is an option unless these additional limitations are a requirement of the integration.
```

nasthagiri: @robrap I agree with @pdpinch that the `%%USER_ID%%` is an additional example of a legacy feature that can be changed with future implementations, given the decision in this OEP.

One suggestion: to minimize the possible confusion, you can refer to `%%USER_ID%%` as an additional example (in a separate note) after this LTI note.

PR #105, opened 2019-02-08
Make track-pulls option obsolete in OEP-2

michaelyoungstrom:
Per @nedbat This is now obselete and should be removed from any existing `openedx.yaml` files.

Let me know if there's a more preferred way of stating this in the doc.

### DIFF
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -7,7 +7,7 @@ OEP-2: Repository Metadata
 +---------------+-------------------------------------------+
 | Title         | Repository Metadata                       |
 +---------------+-------------------------------------------+
-| Last-Modified | 2016-10-13                                |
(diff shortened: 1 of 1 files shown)


oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -78,10 +78,8 @@ following fields:
     it is assumed to be ``False``, unless the reporting tool can auto-detect
     accordance.
 
-.. track-pulls should be moved to a separate OEP
-
-``track-pulls``: boolean (optional)
-    Whether or not this repository should be tracked for creation of OSPR tickets.
+``track-pulls``: obsolete
+    This key is obsolete, please remove if found.
```
nedbat: You should update the Last Modified date at the top of the file.

nedbat: (and then merge it! 👍 )

PR #108, opened 2019-03-12
Draft OEP-34: Lint Amnesty

cpennington:


oeps/oep-0034-bp-lint-amnesty.rst
```diff
@@ -0,0 +1,78 @@
...
+that the warning is on. For instance, a code change might delete a class
+that is imported in a number of other code files. Those imports would now
+be in violation, but they were not changed, and therefor would not be
+flagged by differential linting.
+
+Decision
+--------
+
+In all repositories, we will grant amnesty to all existing linting violations.
+This includes pylint, xsslint, pycodestyle, and any other linting programs
```
robrap: Generally love the idea, but I have some hesitation around xsslint, especially when it comes to repos that haven't yet employed it.

cpennington: I guess my feeling is that adding the ignores is no worse (and probably better) than having cap on the number. Either way, the issues aren't being immediately fixed, and I think flagging them in the code makes them more visible.

oeps/oep-0034-bp-lint-amnesty.rst
```diff
@@ -0,0 +1,78 @@
...
+currently in use. Lints being granted amnesty will be marked with on a
```
pdpinch: is this what you mean?

```suggestion
currently in use. Lints being granted amnesty will be marked on a
```

oeps/oep-0034-bp-lint-amnesty.rst
robrap: The xsslint issues are being fixed. I'd still recommend not listing this, but I'll leave that to the security team to determine.

regisb:
This is a great step forward for the Open edX codebase. Among other benefits it will make life much easier for first time-contributors.
On a related note, as a big fan of static checking, I recently setup type checks with mypy across the entire tutor code base: https://github.com/overhangio/tutor/commit/0a670d7ead3297030e15685cfd6c08bafefd17c4
If anyone is interested I can provide some feedback (spoiler:the experience was very positive).

oeps/oep-0034-bp-lint-amnesty.rst
robrap: @feanil: It looks like this got merged, but I still propose that xsslint gets dropped from this list.

PR #109, opened 2019-04-24
WIP: Rough draft of Pagination OEP-0035

ormsbee:
Still missing a lot of parts. I'm probably going to put this work down until next week, but wanted to park what I had somewhere public.

### DIFF
--- a/oeps/oep-0035-bp-rest-api-pagination.rst
+++ b/oeps/oep-0035-bp-rest-api-pagination.rst
@@ -1,0 +1,257 @@
+========================================
+OEP-0035: REST API Pagination Guidelines
+========================================
+
++-----------------+-------------------------------------------------------------+
+| OEP             | :doc:`OEP-0035 </oeps/oep-0035-bp-rest-api-pagination.rst>` |
++-----------------+-------------------------------------------------------------+
+| Title           | REST API Pagination Guidelines                              |
++-----------------+-------------------------------------------------------------+
+| Last Modified   | 2019-04-21                                                  |
++-----------------+-------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


oeps/oep-0035-bp-rest-api-pagination.rst
```diff
@@ -0,0 +1,257 @@
...
+Your API should accept a ``page_size`` parameter, by defining
+``page_size_query_param``. However if you allow clients to specify the desired
+page size, your *must* specify a ``max_page_size``.
+
+Finally, you *must* limit the number of items in your result set. You can do
+this by modifying your view's ``queryset``. For example::
+
+  class UserViewSet(viewsets.ModelViewSet):
+      serializer_class = UserSerializer
+      queryset = User.objects.all().limit(1000)  # <-- Limit is here
```
bradenmacdonald: @ormsbee do you have a working DRF example of limiting the result set that I can check out?

Django querysets do not have a `.limit(...)` method, and using a lazy slice limit (`queryset = User.objects.all()[:1000]`) does not work because Django "[cannot](https://stackoverflow.com/a/30020676) filter a query once a slice has been taken." and DRF needs to filter the queryset later on.

The DRF paginator could be modified to have a `max_page` argument, but would still try to determine the `count` of results in the queryset, so that doesn't fully help.

sarina:
Is this an active PR/should it stay open, or at minimum, convert to draft?

ormsbee:
I haven't touched this in two years, and I'm unlikely to get around to it soon. I'll close for now. I can always re-open later. Thank you for the reminder. 😄

PR #111, opened 2019-04-30
OEP-0036 Microservice routing layer

fredsmith:


oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,98 @@
...
+new openedx sites quicker and easier than they currently can.  The router
+will provide a map for people to understand which services they need to deploy.
+
+
+Specification
+=============
+
+A layer 7 routing application will be chosen that meets the following
+requirements:
+* High performance connection management capabilities, including support
```
jdmulloy:
```suggestion

* High performance connection management capabilities, including support
```
RST needs the empty line here to render the first bullet properly

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,98 @@
...
+  * ecommerce would occupy /basket
+  * studio would occupy /studio 
+
+links between services would begin to point to the new endpoints, and old endpoints
+serve 301 redirects.
+
+Rationale
+=========
+
+This architecture is a common way to deploy microservices.
```
coryleeio: I think you stated many of them above such as abstracting various concerns from the applications, terminating (and configuring) TLS only on the router, etc, but I would expand with some of those reasons, or move the contents of the rationale section up into the motivation section.

coryleeio: Some of the things that are possible are listed in this article:
https://www.getambassador.io/about/why-ambassador/

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,93 @@
...
+that only manage a small number of endpoints, without requring developers to 
+build out load balancing or CDN infrastructure. 
+
+We are currently using a layer 7 routing framework for prospectus, allowing
+us to replace drupal generated pages with gatsby generated pages quickly and
+seamlessly.
+
+This change will allow developers, openedx operators and resellers to deploy
+new openedx sites quicker and easier than they currently can.  The router
+will provide a map for people to understand which services they need to deploy.
```
coryleeio: Some thoughts and questions: 

Will the router handle auth in any way or just routing?

Is the router a required component to run our stack now? (because it does stuff with auth)

It is important to note that if the router is a required component that it definitely cannot be k8 native.

If it is not a required component we can use a k8s native one like ambassador, but it will not provide this 'map for people to understand which services they need to deploy' at that point.

Do we configure the services to talk to one another with or without knowledge of the router?

If we want to look at how services discover one another as well as how services are routed from the outside world we might consider a service mesh and a L7 router.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+and regularly causes problems with cross service user information sharing. In
+Addition, modifying or splitting a monolithic service into microservices resuls
+in confusing URL changes and costly migrations.
+
+A layer 7 routing layer would allow us to separate the user facing URL or API
+endpoint from the backend implementation. It would allow microservices to exist
+that only manage a small number of endpoints, without requring developers to 
+build out load balancing or CDN infrastructure. 
+
+We are currently using a layer 7 routing framework for prospectus, allowing
```
tuchfarber: While this is good to point out during internal conversation, prospectus is not an open source / Open edX feature, so I don't know if it makes sense in this document.

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I've addressed the capabilities questions, in that, it might.  We should figure out if auth is something we want from this component.  We have this capability with our current platform (nginx), as well as request throttling, and have typically pushed that role up the stack to django.

The routing layer is a required component.

I would suggest that all inter-service communication go through this routing layer, to minimize areas of configuration and access control.

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I think prospectus, or some similar marketing site CMS, should be a component in openedx.  There were a handful presented at Openedx conf.

tuchfarber: I agree. It came up a lot during the theming discussion during the summit. I think the conclusion was:
> Everyone wants different features in their marketing site. We (edX) will have our own and will expose the features to build your own. If the community wants to get together and build a common marketing site w/ cms, they are welcome to.

Since edX, wouldn't be using it, we would be bad maintainers of that Open edX project.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+requirements:
+
+* High performance connection management capabilities, including support
+  for slow clients or backends, and wire-speed throughput.
+* SSL termination on the router.
+* Flexible configuration backends including container service discovery,
+  flat files, and filesystem routing.
+* Path based routing, including path subsets and regular expression matching
+
+This layer 7 routing application will be deployed on a common hostname.
```
tuchfarber: Does this imply that we won't be able to use custom subdomains with this routing layer?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+We are currently using a layer 7 routing framework for prospectus, allowing
+us to replace drupal generated pages with gatsby generated pages quickly and
+seamlessly.
+
+This change will allow developers, openedx operators and resellers to deploy
+new openedx sites quicker and easier than they currently can.  The router
+will provide a map for people to understand which services they need to deploy.
+
+
+Specification
```
tuchfarber: This might be implied because it's in an OEP, but...

Since this solution is replacing a vendor specific solution we have today, it would be good to note that this would be a separate application that is part of the Open edX stack and therefore it can run independently both in production and locally in a devstack. 

Partially related to this: In the MFE world, will this only act as a router to a separate service, or will this also be able to host static files? Internally we have a solution that's DNS + static webserver for MFEs. Will this handle both of those use cases, or will MFE's static webserver be a separate OEP / service?

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I think a goal is to reduce hostname sprawl.  http/2 and jwt work better with a single hostname (I think) so unless there's a good reason we would discourage services moving to a custom subdomain.

Technically, there isn't a big problem with custom subdomains for, say, multi-tenant installations.

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I think the question of where static content is hosted depends on the tech choice - some layer 7 router implementations (eg. traefik, ambassador) don't support hosting static content and require you to run a static content server like nginx

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+* Path based routing, including path subsets and regular expression matching
+
+This layer 7 routing application will be deployed on a common hostname.
+
+Services will begin to move API and user endpoints behind this layer 7 router
+after declaring their top level routing path. eg.
+
+* the LMS would occupy /courses
+* prospectus would occupy /course, /school, /micromasters
+* ecommerce would occupy /basket
```
robrap: Possibly update `/basket` to `/orders and /payments` to be closer to truth.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+* studio would occupy /studio 
+
+links between services would begin to point to the new endpoints, and old endpoints
+serve 301 redirects.
+
+Possible additional functionality that may be implemented at the l7 router:
```
robrap: I've sometimes add a `Future` or `Next Steps` section if we want to note things that we don't yet want to commit to.  Also, do we want to discuss/mention the relationship between this and the potential for using an API Gateway?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+using a layer 7 routing framework. 
+
+Motivation
+==========
+
+Microservices are currently implemented using layer 4 (DNS) based routing. Each
+microservice occupies a unique hostname, with dedicated ELB or CDN 
+infrastructure to support routing user traffic to the service. This pattern
+is simple to monitor, deploy, and manage, but slow and expensive to build,
+and regularly causes problems with cross service user information sharing. In
```
robrap: Also, maybe add some details to the cross-service issues?  CORS, inability to share localStorage, other?

On the flip side, I don't know if this affects cookies in the following ways:
* Did we have large subdomain specific cookies, that now all will be sent to every request across the entire site?  Also, could we more easily hit our cookie size limit of 4kb?
* Could we have any cookie name conflicts that were previously namespaced to the subdomain?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+Addition, modifying or splitting a monolithic service into microservices resuls
```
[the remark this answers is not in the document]
albemarle: Addition => addition

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,99 @@
...
+in confusing URL changes and costly migrations.
+
+A layer 7 routing layer would allow us to separate the user facing URL or API
+endpoint from the backend implementation. It would allow microservices to exist
+that only manage a small number of endpoints, without requring developers to
```
albemarle: requiring

oeps/oep-0036-microservice-routing-layer.rst
felipemontoya: I agree with the discourage to move MFE to new domains. Does this apply to more than just MFE, for example other IDAs?

Also, multi-tenant installations very often have a completely different domain. This sometimes is problematic for SSL termination, but most applications that deal with SSL termination support SNI natively.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+* Rate limiting
+* Web Application Firewall
+
+
+Backwards Compatibility
+=======================
+
+Services will begin to move API and user endpoints behind this layer 7 router after declaring their top level routing path. eg.
+
+* the LMS would occupy /courses
```
felipemontoya: I see now that it definitely includes IDAs.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+* prospectus would occupy /course, /school, /micromasters
+* ecommerce would occupy /orders, /payments
```
felipemontoya: If all the IDAs are located on the same domain, does this create a clash with the "sessionid" cookies?
Will the individual applications require that their cookies are prefixed as they are now? E.g "insights_sessionid". Or will all the authentication be solved by JWT?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+* SSL termination on the router.
+* Flexible configuration backends including container service discovery, flat files, and filesystem routing.
+* Path based routing, including path subsets and regular expression matching
+
+This layer 7 routing application will be deployed on a common hostname.
+
+Rationale
+=========
+
+A layer 7 routing layer would allow us to separate the user facing URL or API endpoint from the backend implementation. It would allow microservices to exist that only manage a small number of endpoints, without requring developers to build out load balancing or CDN infrastructure.
```
ormsbee: Nit: "requring" -> "requiring"

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+Backwards Compatibility
+=======================
+
+Services will begin to move API and user endpoints behind this layer 7 router after declaring their top level routing path. eg.
+
+* the LMS would occupy /courses
+* prospectus would occupy /course, /school, /micromasters
+* ecommerce would occupy /orders, /payments
+* studio would occupy /studio 
+
```
ormsbee: If we're moving around a bunch of URLs with 301 redirects, is it too much to force each service to only have its own top level namespace? Allowing each IDA to have multiple, arbitrary top level URLs like /course vs. /courses sounds like it will get confusing over time (as we get dozens of these), and potentially lead to name collisions and potentially configuration issues down the road. Whereas if there's a 1:1 mapping of IDA to top level, then once an IDA is established, they control everything below that namespace and they don't have to touch the centralized routing layer config to add a new directory somewhere.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+links between services would begin to point to the new endpoints, and old endpoints serve 301 redirects.
+
+Old URLs will be migrated using 301 redirects, which will automatically migrate users browsers to the new endpoints. For an overlapping time period, both URLs will be maintained and function, so that services can transition to the new schema.
```
ormsbee: Are we talking a big-bang migration of everything, or will the new routing layer slowly see services ported over to it?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+    - Draft
+  * - Type
+    - Architecture
+  * - Created
+    - 2019-04-25
+
+Abstract
+========
+
+Proposes a key openedx component that will connect users to applications using a layer 7 (application layer/URL Based) routing framework.
```
nedbat: "Open edX"

oeps/oep-0036-microservice-routing-layer.rst
ormsbee: I guess the larger version of this is that I would like to better understand what the restrictions on usage are, and how we're going to reduce the chances of teams accidentally stepping on each other. I strongly support having some routing layer as proposed in this document, but I'm concerned that it will gradually grow in complexity into a hodgepodge of global and local rules (e.g. "hey, did my new redirect pattern affect images for XYZ other service?") that people are afraid to touch. I'm concerned that we're going to contort it to accommodate use cases that add complexity rather than forcing URL restrictions or requiring downstream changes. I'm concerned that someone else is going to want `/courses/`, and so we'll write a regex matcher that says `/courses/{uuid-pattern}` goes to Publisher while `/courses/{course-key}` goes to the LMS.

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+We are currently using a layer 7 routing framework for the edx.org marketing site, allowing them to replace drupal generated pages with gatsby generated pages quickly and seamlessly.
+
+This change will allow developers, openedx operators and resellers to deploy new openedx sites quicker and easier than they currently can.  The router will provide a map for people to understand which services they need to deploy.
+
+
+
+Implementation
+==============
+
+* to be determined
```
regisb: It seems that this OEP was written with [Traefik](https://docs.traefik.io/configuration/backends/file/), [Ambassador](https://www.getambassador.io/) and Nginx in mind. Why not mention these options here?

oeps/oep-0036-microservice-routing-layer.rst
ormsbee: Okay. Just to check then -- you think it's reasonable to have a 1:1 mapping of URL prefix to service if we're not using technical words (e.g. `/learning` instead of `/lms`)? Or do you think that we need the flexibility to give services multiple top-level directories and decouple URLs from their backing services entirely?

oeps/oep-0036-microservice-routing-layer.rst
```diff
@@ -0,0 +1,89 @@
...
+* Web Application Firewall
+
+
+Backwards Compatibility
+=======================
+
+Services will begin to move API and user endpoints behind this layer 7 router after declaring their top level routing path. eg.
+
+* the LMS would occupy /courses
+* prospectus would occupy /course, /school, /micromasters
```
ormsbee: Since we'll be starting out the gate with stuff that's not applicable to folks in the community, it would be good to map out how we'll cleanly separate out the pieces of it that are generically Open edX vs. the pieces that are site-specific customizations. Things like:

* Can a site cleanly upgrade to Koa/Lilac without having to change any site-specific config for their own marketing pieces?
* What conventions should we have so that we don't introduce namespace conflicts later on? Should site-specific stuff be put on its own domain? With a known prefix? With an org-specific prefix (e.g. `/edx`)?

oeps/oep-0036-microservice-routing-layer.rst
regisb: > you think it's reasonable to have a 1:1 mapping of URL prefix to service if we're not using technical words (e.g. /learning instead of /lms)?

Yes.

davidjoy:
Apologies for the delay - here are my notes from our in-person discussion about the OEP.  Please feel free to comment with any corrections/additional notes.  

https://docs.google.com/document/d/1D3asBdpZ-XnA-ieZJ5YkqcogUjgfLNndA5XgK7qLIPY/edit?usp=sharing

felipemontoya:
Thanks a lot for sharing those notes @davidjoy. They have a bit of a story so I did not want to share my thoughts there. I will post them here instead. Some might not be part of the OEP but might contribute to the discussion.

- Some Open edX sites might still have reasons after this is implemented to keep separate domains. The most familiar that comes to mind is the university that hires service providers to stand up an instance. They will be very reluctant about passing all the traffic from all other university services through the routing application. The Open edX installation will surely go through the routing app, but there is still a very useful case for something like the marketing site integration. 

- Multiple domains and multiple SSL termination is necessary for multi-tenancy (whitelabel), but as a large multi-tenant instance we are willing to work on our own, or fork and add this functionality. If there is something we can do to help in this regard ahead of time, we will be very happy to collaborate.

- Is this going to be something to be worth waiting for in regards to Juniper?
The first MFE that I know is the profile page, which comes after ironwood. If Juniper is cut before this goes out, it means that anyone following open releases will have to go to subdomains for juniper and then go to the routing layer solution on Koa. Since this change is so fundamental I would rather skip the subdomains and wait for the routing layer. Now, I am not sure if something was decided against waiting for features when open releases are on the line. I say this only from the perspective of minimizing breaking infrastructure changes in many installations.

- Not a part of the OEP, but as we are moving rapidly towards more MFE the theming of those will be crucial for the community

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I'm expecting a gradual migration.  First candidates will be new services.

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: Your suggested urls.json file looks a lot like I imagined this would work.  My expectation is that this will jinja2 straight into an nginx config, without ansible or any other complex tooling.  The expectation is that this will never be implemented with ansible.

I agree that using meaningful URLs is a good idea.

at some point, it will include an implementation, but I'm still gathering requirements at this stage.

oeps/oep-0036-microservice-routing-layer.rst
fredsmith: I would like this to be entirely applicable to the community out of the gate.  Perhaps we should segment edx.org things into their own bucket, but since the plan is for us to take over the marketing site, I would like to keep their URLS functional.

davidjoy:
I'm going to close this PR, as there hasn't been any activity in over two years and I don't think we have the appetite to continue this.  Feel free to chime in if you feel otherwise and would like to pick up this work.

tuchfarber:
Fare thee well OEP-36 :vulcan_salute:

fredsmith:
😢

PR #115, opened 2019-05-24
Don't prefix with zeros in the titles for consistency.

feanil:


oeps/oep-0003-arch-async-tasks.rst
```diff
@@ -3,7 +3,7 @@ OEP-3: Asynchronous Task Management
 ===================================
 
 +---------------+-------------------------------------------+
-| OEP           | :doc:`oep-0003-arch-async-tasks`          |
+| OEP           | :doc:`oep-3-arch-async-tasks`             |
```
nasthagiri: Does this link to the Doc file and should still contain the preceding 000s?

feanil: It turns out this is just a link is a self reference so it goes nowhere.  It didn't break anything.

PR #117, opened 2019-06-05
Update openedx.yaml when archiving

nedbat:
@feanil I forget if there was another detail we wanted to add to this.

### DIFF
--- a/oeps/oep-0014-proc-archive-repos.rst
+++ b/oeps/oep-0014-proc-archive-repos.rst
@@ -123,15 +123,18 @@ For Repos that are Forks
 ------------------------
 
 If the repository is a fork of an upstream repository that is not within the
-edX organization, and will no longer be maintained, it can be deleted with the
-following steps.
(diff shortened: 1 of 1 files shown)


oeps/oep-0014-proc-archive-repos.rst
```diff
@@ -82,36 +85,49 @@ steps to see if anyone would like to take ownership of it:
...
     this repository will also not be merged.
 
+
 For Repos that are Forks
 ------------------------
-If the repository is a fork of an upstream repository that is not within the edX organization, and will no longer be maintained, it can be deleted with the following steps.
 
-1. Review any differences between the upstream parent and the forked repo. Capture anything
-   that should be kept for historical purposes.
+If the repository is a fork of an upstream repository that is not within the
```
feanil: Did you want to add that steps for moving the fork to the edx-unsupported org?

nedbat: Added. Ready for review.

PR #118, opened 2019-06-07
OEP-37: Dev Data

jmbowman:
This is a first draft of an approach to consistently managing data sets that can be shared across unit tests and development environments to facilitate the development and testing of Open edX functionality.  I suspect we'll want better nomenclature for some of the ideas outlined here, but first I'd like to get some feedback on whether or not this is even the direction we want to go.  Feedback of any and all kinds are welcome!

### DIFF
--- a/oeps/oep-0037-bp-test-data.rst
+++ b/oeps/oep-0037-bp-test-data.rst
@@ -171,6 +171,19 @@ Rejected Alternatives
 .. _django-sanitized-dump: https://github.com/andersinno/django-sanitized-dump
 .. _model-mommy: https://model-mommy.readthedocs.io/en/latest/
 
+
+Related Decisions
+=================
+
+The following related decisions modify or enhance this OEP, but have not yet been fully incorporated as updates to this OEP:
+
+.. toctree::
+   :caption: OEP-37 Decisions
+   :maxdepth: 1
+   :glob:
+
+   oep-0037/decisions/*
+
 Change History
 ==============
 
--- a/oeps/oep-0037-bp-test-data.rst
+++ b/oeps/oep-0037-bp-test-data.rst
@@ -1,0 +1,177 @@
+==================
+OEP-37: Dev Data
+==================
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :doc:`OEP-37 </oeps/oep-0037-bp-test-data>`
+   * - Title
+     - Devstack Dev Data
+   * - Last Modified
+     - 2019-11-25
+   * - Authors
+     - Jeremy Bowman <jbowman@edx.org>
+   * - Arbiter
+     - Bill DeRusha <bill@edx.org>, Feanil Patel <feanil@edx.org>
+   * - Status
+     - Under Review
+   * - Type
+     - Best Practice
+   * - Created
+     - 2019-06-07
+   * - Review Period
+     - 2019-06-19 to 2019-11-25
+   * - Resolution
+     - Provisional
+
+Abstract
+========
+
+Proposes best practices in creating and maintaining sample dev data for use in Open edX test cases and development environments.
+
+Motivation
+==========
+
+When testing or developing an Open edX service, it is usually necessary to have some dev data already loaded and configured: users, courses, grades, etc.  This is true for unit tests, browser automation tests, load tests, development work in devstack, and more.  There are many different ways of creating and loading such data, and in the absence of concrete guidance on best practices, many of them have been (inconsistently) used in Open edX.  This makes it difficult to decide how to prepare such dev data, to learn how to load it when needed, and also to keep it working correctly as the code changes over time.
+
+This OEP attempts to provide the guidance needed to manage this data logically and consistently in order to streamline development, refactoring, and automated testing.
+
+Specification
+=============
+
+There are a few different aspects of dev data management, each of which has different tools that can make it more painless:
+
+Reusing Data
+------------
+
+Once a development environment has been populated with a dev data set comprehensive enough to exercise certain functionality in Open edX, it can be useful to "save it" for later use.  The first step is to make sure that any `PII`_ has been stripped from the data.  Even development data may include a developer's email address and other personal information.  Tools such as `dj_anonymizer`_ can be useful in making a first pass at stripping such PII, but it should be followed by manual verification against the `OEP-30`_ annotations.
+
+Once PII has been removed, an attempt can be made to export just the needed data.  This may be a combination of course export to OLX and generation of fixtures from selected relational data using a utility such as `django-fixture-magic`_.  These fixtures should not be saved verbatim, but rather used as a guide for producing Python code that generates roughly equivalent data using ``factory_boy`` as described below.
+
+Note that if the following recommendations on creating data are followed, test data created for a unit test can be fairly easily re-purposed for use as development environment data sets.  This can be useful because the starting point is already free of PII and has been verified by tests to work as intended.
+
+.. _PII: https://open-edx-proposals.readthedocs.io/en/latest/oep-0030-arch-pii-markup-and-auditing.html
+.. _dj_anonymizer: https://github.com/preply/dj_anonymizer
+.. _OEP-30: https://open-edx-proposals.readthedocs.io/en/latest/oep-0030-arch-pii-markup-and-auditing.html
+.. _django-fixture-magic: https://github.com/davedash/django-fixture-magic
+
+Creating Data
+-------------
+
+Whether defining it from scratch or basing it on a dump of data from an existing environment, at some point code needs to be written which will populate the data stores of each relevant service in order to perform the desired test or development task.  For relational data in simple unit tests, this is often done in Python via the `factory_boy`_ package (and custom model factories using it) directly in ``setUpTestData`` or ``setUp``.  For development data (or more complex test setup which could be useful to reproduce in a development environment), the setup code should instead reside in its own function or static method.  For example:
+
+.. code-block:: python
+
+    def account_retirement_example(data):
+        data.test_user = UserFactory.create()
+        data.test_superuser = SuperuserFactory.create()
+        data.original_email = data.test_user.email
+
+        # Misc. setup
+        PendingEmailChangeFactory.create(user=data.test_user)
+        UserOrgTagFactory.create(user=data.test_user, key='foo', value='bar')
+        UserOrgTagFactory.create(user=data.test_user, key='cat', value='dog')
+
+        CourseEnrollmentAllowedFactory.create(email=data.original_email)
+
+(JavaScript tests seeking to stub out data returned from a server can use similar functions utilizing a package like `rosie`_ or `factory-girl`_.)
+
+Setting attributes of the provided argument like this allows the test object or class to be provided when used in a unit test (so assertions can be written naturally), and something like a `SimpleNamespace`_ object can be provided when run from a developer console so the generated data can be quickly located and inspected.  When needed, course data should be loaded from an `OLX`_ archive via ``call_command()`` using the existing management commands for course import.
+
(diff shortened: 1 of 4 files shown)


oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,254 @@
...
+
+.. _PII: https://open-edx-proposals.readthedocs.io/en/latest/oep-0030-arch-pii-markup-and-auditing.html
+.. _dj_anonymizer: https://github.com/preply/dj_anonymizer
+.. _OEP-30: https://open-edx-proposals.readthedocs.io/en/latest/oep-0030-arch-pii-markup-and-auditing.html
+.. _django-fixture-magic: https://github.com/davedash/django-fixture-magic
+
+Creating Data
+-------------
+
+Whether defining it from scratch or basing it on a dump of data from an
```
ahmedaljazzar: @jmbowman Why don't we load those data seamlessly without having to worry about writing the commands and integrating it in all backends? I was suggesting here having some sort of test case that loads the data for every test case it hits in certain path dynamically. So for example if I'm running:

```
./manage.py test ecommerce.courses.tests.test_utils
```
The runner loads data from `ecommerce.courses.tests.data.*` seamlessly. We still can use the `load_test_data` command to perform this operation, but what do you think?

jmbowman: Currently, the test data required for each test case can vary widely even between different methods in the same `TestCase` subclass.  Instead of guessing at what data is needed for a given test case based on where it is, I'm thinking we'd be better off splitting the data setup out into separate functions which can be shared by multiple test cases (even in different areas of the code) which call them during `setUpTestData`, `setUp`, or the test method itself as appropriate.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,254 @@
...
+test or development task.  For relational data in simple unit tests, this is
+often done via the `factory_boy`_ package (and custom model factories using
+it) directly in ``setUpTestData`` or ``setUp``.  For development data (or
+more complex test setup which could be useful to reproduce in a development
+environment), the setup code should instead reside in its own Python function
+or static method.  For example:
+
+.. code-block:: python
+
+    def account_retirement_example(data):
```
cpennington: Should these functions be allowed to take parameters to allow them to be somewhat configurable (for example, how many courses to create?). If so, how could we expose those in the `load_test_data` management command?

bderusha:
This OEP addresses many of the pain points around test data.  However, one of my concerns that is not addressed is around drift/maintenance burden.  One of the biggest pain points around data is the cross-service use case.  Workflows which need the same course to exist in the mktg site, lms, ecommerce, and discovery.  I would love it if we had an answer to that that was more than make the change in 10 repos to maintain a functional dataset.

If we had a way of declaring first class test objects which could be loaded and used by all services that would be a huge win.

oeps/oep-0037-bp-test-data.rst
jmbowman: Yes, the functions could take additional parameters.  I was originally thinking that we'd need custom management commands for such cases, but I guess they could be given as positional arguments instead.  What I originally wrote:

> Most data sets should attempt to use a standard function path in each IDA to
allow use of this generic data loading command, but particularly complex cases
may use custom management commands and/or devstack ``make`` targets (perhaps
even taking arguments to specify parameters of the generated data set).

The devstack make target could probably also take parameters, something like:

```make load_test_data DATA_FUNCTION=test_data.retirement.account_retirement_example DATA_FUNCTION_ARGS=42,second_argument```

**Comment by jmbowman** [comment]
@bderusha I added a section on using minimal YAML files to define and link test entities which can then be parsed and interpreted by an appropriate data loader in each service; let me know if that addresses some of your concern.

File: oeps/oep-0037-bp-test-data.rst [hunk 3]
```diff
@@ -0,0 +1,299 @@
...
+Abstract
+========
+
+Proposes best practices in creating and maintaining sample data for use in
+Open edX test cases and development environments.
+
+Motivation
+==========
+
+When testing or developing an Open edX service, it us usually necessary
```
iloveagent57: s/us/is/

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+.. _dj_anonymizer: https://github.com/preply/dj_anonymizer
+.. _OEP-30: https://open-edx-proposals.readthedocs.io/en/latest/oep-0030-arch-pii-markup-and-auditing.html
+.. _django-fixture-magic: https://github.com/davedash/django-fixture-magic
+
+Creating Data
+-------------
+
+Whether defining it from scratch or basing it on a dump of data from an
+existing environment, at some point code needs to be written which will
+populate the data stores of each relevant IDA in order to perform the desired
```
iloveagent57: s/IDA/service/ (or give the definition of IDA).

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+.. _rosie: https://www.npmjs.com/package/rosie
+.. _factory-girl: https://www.npmjs.com/package/factory-girl
+.. _SimpleNamespace: https://docs.python.org/3/library/types.html#types.SimpleNamespace
+.. _OLX: https://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/index.html
+.. _Faker: https://faker.readthedocs.io/en/stable/
+
+Loading Data
+------------
+
+Each IDA should provide a ``load_test_data`` management which takes dotted
```
iloveagent57: insert "command" after "management"

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+management commands and ``make`` targets described below.
+
+.. _factory_boy: https://factoryboy.readthedocs.io/en/latest/index.html
+.. _rosie: https://www.npmjs.com/package/rosie
+.. _factory-girl: https://www.npmjs.com/package/factory-girl
+.. _SimpleNamespace: https://docs.python.org/3/library/types.html#types.SimpleNamespace
+.. _OLX: https://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/index.html
+.. _Faker: https://faker.readthedocs.io/en/stable/
+
+Loading Data
```
bmedx: This OEP is great! It would be nice if we explicitly dealt with the case where test data is loaded multiple times. Ex. into a dev environment, modified by hand, and then reloaded. I would expect that in that case the data loading function would stomp on the manual changes, but it would be good to have at least some guidance so people don't need to know the details of every command to know what will happen when it gets re-run.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
++-----------------+--------------------------------------------------------+
+| Title           | Test Data Management                                   |
++-----------------+--------------------------------------------------------+
+| Last Modified   | 2019-06-19                                             |
++-----------------+--------------------------------------------------------+
+| Authors         | Jeremy Bowman <jbowman@edx.org>                        |
++-----------------+--------------------------------------------------------+
+| Arbiter         | Bill DeRusha <bill@edx.org>                            |
++-----------------+--------------------------------------------------------+
+| Status          | Under Review                                           |
```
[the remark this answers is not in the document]
jmbowman: We plan to create such a reference implementation shortly.  The OEP review process helped make sure that the reference implementation will be as useful as possible.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+There are a few different aspects of test data management, each of which
+has different tools that can make it more painless:
+
+Reusing Data
+------------
+
+Once a development or production environment has been populated with a data
+set comprehensive enough to exercise certain functionality in Open edX, it can
+be useful to "save it" for later use.  The first step is to make sure that any
+`PII`_ has been stripped from the data.  This is a particular concern for
```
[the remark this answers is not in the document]
jmbowman: I was anticipating reuse of just small sets of database records, not big chunks of a course.  Might be best to completely avoid suggesting reuse of production data, though.  As mentioned below, the PII stripping is relevant even for dev environment data.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+production data that may include customer information, but even development
+data may include a developer's email address and other personal information.
+Tools such as `dj_anonymizer`_ can be useful in making a first pass at
+stripping such PII, but it should be followed by manual verification against
+the `OEP-30`_ annotations.
+
+Once PII has been removed, an attempt can be made to export just the needed
+data.  This may be a combination of course export to OLX and generation of
```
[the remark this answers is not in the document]
jmbowman: Removing discussion of prod data reuse as noted above.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+fixtures from selected relational data using a utility such as
+`django-fixture-magic`_.  These fixtures should not be saved verbatim, but
+rather used as a guide for producing Python code that generates roughly
+equivalent data using ``factory_boy`` as described below.
+
+Note that if the following recommendations on creating data are followed, test
+data created for a unit test can be fairly easily re-purposed for use as
```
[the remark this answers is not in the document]
jmbowman: No, just specific records needed to exercise particular functionality.  And as noted above, the primary intention is to either create completely synthetic data or reuse data created in a dev environment (especially when creating the data correctly is much easier via the UI than by hand coding).

oeps/oep-0037-bp-test-data.rst
jmbowman: Because we're trying to avoid specifying primary keys, the most likely behavior is that reloading a fixture will add a new set of comparable records with different IDs.  Unless that poses a problem, I'll just document that behavior here.

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,299 @@
...
+Reference Implementation
+========================
+
+For test suite data, see the usage of ``factory_boy`` and OLX in edx-platform
+and other repositories.  For development data, the ``create-test-course`` and
+``build-courses`` devstack ``make`` targets (as well as
+``programs/provision.sh``) should be reviewed and refactored to comply with
+the guidelines outlined here.
+
+Rejected Alternatives
```
albemarle: This is one of the best Rejected Alternatives sections I've seen.

jmbowman:
@nasthagiri @bmedx Trying to restore momentum on this; I've attempted to address your last comments, any other suggestions?

oeps/oep-0037-bp-test-data.rst
```diff
@@ -0,0 +1,180 @@
...
+.. _rosie: https://www.npmjs.com/package/rosie
+.. _factory-girl: https://www.npmjs.com/package/factory-girl
+.. _SimpleNamespace: https://docs.python.org/3/library/types.html#types.SimpleNamespace
+.. _OLX: https://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/index.html
+.. _Faker: https://faker.readthedocs.io/en/stable/
+
+Loading Data
+------------
+
+Each IDA should provide a ``load_test_data`` management command which takes dotted paths to one or more test data generation functions as arguments.  For example, the command to load the example data above into the LMS might look something like this:
```
adzuci, line 97: Reading this OEP a bit late.  This may be a silly question, but if we do decide to use a dump file to speed up local development environments (such as for edx-platform), should `load_test_data` be primarily used to generate the data that is then dumped to a file and would there be another make target or place for this dump to be consumed?  In the current `enterprise-catalog` provisioning script for LMS we run `docker-compose exec -T mysql mysql edxapp < dump.sql` directly rather than using a make target, but I could see us not wanting to do that.
https://github.com/edx/enterprise-catalog/blob/master/decentralized_devstack/provision-lms.sh#L15

PR #119, opened 2019-06-12
OEP-38 Data Modeling

macdiesel:
A guide to data modeling at edX

### DIFF
--- a/oeps/oep-0038-Data-Modeling.rst
+++ b/oeps/oep-0038-Data-Modeling.rst
@@ -1,0 +1,411 @@
+======================================
+OEP-38:  Data Modeling Best Practices
+======================================
+
++---------------+------------------------------------------------------------+
+| OEP           | :doc: `OEP-0038 <oep-0038-data-modeling>`                  |
++---------------+------------------------------------------------------------+
+| Title         | Data Modeling Best Practices                               |
++---------------+------------------------------------------------------------+
+| Last Modified | 2019-09-19                                                 |
++---------------+------------------------------------------------------------+
+| Author        | Brian Beggs <bbeggs@edx.org>                               |
++---------------+------------------------------------------------------------+
+| Arbiter       | Albemarle                                                  |
++---------------+------------------------------------------------------------+
+| Status        | Accepted                                                   |
++---------------+------------------------------------------------------------+
+| Type          | Best Practices                                             |
++---------------+------------------------------------------------------------+
+| Created       | 2019-05-01                                                 |
++---------------+------------------------------------------------------------+
+| Review Period | 2019-08-01 - 2019/10/01                                    |
++---------------+------------------------------------------------------------+
+| Resolution    | 'Original pull request'                                    |       
++---------------+------------------------------------------------------------+
+
+Abstract
+========
+
+In order to obtain the highest possible value from data collected in the
+Open edX ecosystem, this document attempts to provide guiding thoughts and
+principles on data modeling.
+
+.. contents::
+
+What this OEP explicitly does not cover
+=======================================
+
+-  Data warehousing
+
+-  Data pipelines
+
+-  Event data
+
+
+Definitions
+===========
+
+**BI** - Business intelligence. Technologies, applications and practices for
+the collection, integration, analysis, and presentation of business
+information. The purpose of Business Intelligence is to support better
+business decision making.
+
+**Composite Key** - In the context of relational databases, a composite key
+is a combination of two or more columns in a table that can be used to
+uniquely identify each row in the table. Uniqueness is only guaranteed
+when the columns are combined; when taken individually, the columns do
+not guarantee uniqueness.
+
+**CUD** - Create, Update, Delete. These are some of the actions that may
+happen to data in a database system.
+
+**CRUD** - Create, Read, Update, Delete. Same as CUD but with the *Read* operation added.
+
+**Data Best Practices** - These practices are designed to help teams create
+rich and efficient data models within the Open edX ecosystem. They are not
+standards but guidelines to help teams think about how to store data.
+
+**Data Dimension** - A Data Dimension is a set of data attributes pertaining
+to something of interest to a business. Dimensions are things like
+"customers", "products", "stores" and "time". For users of Data
+Warehouses, data dimensions are entry points to numeric facts (e.g.
+sale, profit, revenue) that a business wishes to monitor.
+
+**Data Modeling** - A process used to define and analyze data requirements
+needed to support the business processes within the scope of
+corresponding information systems in organizations.
+
+**Data Modeling Standards** - The most basic and standard design principles
+for data modeling. These standards must be adhered to when creating new
+models or updating existing models in the Open edX ecosystem.
+
+**IDA** - Independently Deployed Application. Similar to a microservice or
+stand alone application.
+
+
+Open edX Ecosystem
+------------------
+
+Used in this document, the phrase "Open edX Ecosystem" includes the
+services that comprise and support the running of an Open edX
+installation (e.g.,. edx.org). This includes:
+
+- The LMS and Studio
+
+- IDAs such as Notes, Ecommerce, and Forums
+
+- Backend systems such as data analytics and operational logging
+
+- Third party services used to extend functionality of applications, such as Optimizely, Google Analytics, and Sailthru
+
+
+Specification
+=============
+
+Guiding Principles
+------------------
+This section is intended to give the reader a framework to think about data modeling at edX.
+
+Data is an asset
+~~~~~~~~~~~~~~~~
+
+edX uses data to decide which people should receive marketing emails,
+who passes or fails a course, or how much to pay our partners. Our
+partners use our data to target users and to
+refine and improve their courses. The data we collect today is being
+used to advance academic research about online learning and pedagogy.
+Decision making at edX should be data-driven and based on this collected data.
+
+This data is one of our most valuable assets and it
+should be a first-order concern. Save everything (disk is cheap). The more data we are able to collect about users,
+behavior, and system state now, the more opportunities we’ll have to
+improve our decision-making in the future.
+
+Use industry best practices
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Data stored in the OpenEdX ecosystem should adhere to industry best
+practices. For example, since an industry best practice uses numerical
+identifiers to identify rows of data, the Open edX ecosystem should also use
+numerical identifiers. Adhering to industry best practices and the
+practices outlined in this document ensures our data is approachable for
+experienced engineers and new hires alike and allows us to leverage 3rd
+party tools more easily to assist in the analysis phase.
+
+Don’t make me think
+~~~~~~~~~~~~~~~~~~~
+
+Steve Krug, the author of “Don’t Make Me Think”, says: “Your objective
+should always be to eliminate instructions entirely by making everything
+self-explanatory, or as close to it as possible.” This principle should
+also be applied to data modeling. Data models should have descriptive
+names clearly identifying the data that the model holds. The field names
+should make sense to most people familiar with the domain, without
+needing much clarifying documentation.
+
+Consider the analytics user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For every hour of engineer-time spent creating a data model, many
+more are spent using the model and analyzing its data. Taking
+the time to consider how the data will be used, and thoughtfully design
+a data representation, can potentially save many hours during the
+analysis phase.
+
+This is especially true since changing data models, once they are in
+use, can be time-consuming. A data model change in a core application
+can affect many other systems. Small changes to a data model could
+possibly cause hours or days of work for different teams throughout the
+organization. An example of the types of work that may occur while changing a model:
+
+-  Data engineering and Analytics to update their workflows
+
+-  BI teams to update reports that utilize this data,
+
+-  Devops to deploy migrations to large tables, with the risk of
+      potential downtime.
+
+Due to the potentially high cost of changing a model, it pays to get it
+as correct as possible the first time.
+
+Think about the person who will analyze this data later and the person
+who manages the system day to day. (That person could possibly be future
+you!) Try to do as much careful design up front to make your quality of
+life better later.
+
+History is best (but events are a close second)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Historical data accuracy is best when history about a change is captured at the database level. Changes in data relating to financial systems (e.g. enrollments, payments, course price changes) should be stored historically where the change is made, preferably in the same system as the system of record. This allows us to reconstruct the data at different points with much greater reliability.
+
+If, for some reason, it is not possible to create history at the time of
+entry (for example, if we expect the table to become too large or the
+writes to be too frequent), it is recommended that an event be issued in its
+place.
+
+
+Data Modeling Standards
+-----------------------
+
+The standards below are designed to ensure edX can gain the highest value and insights from the data.
+The application of these standards is the most basic level of support to which all applications in the Open edX ecosystem
+should adhere. When creating new applications or models please ensure the models being created conform to the following.
+
+Use a numeric primary key
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  It is recommended to use :code:`BigAutoField`_.
+.. _BigAutoField: https://docs.djangoproject.com/en/2.2/ref/models/fields/#bigautofield
+
+-  Do not use composite based primary keys. Use a primary key column.
+
+Have updated and created timestamps
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  The preferred method for doing this in Open edX Django applications is to inherit the :code:`TimeStampedModel`_. class.
+.. _TimeStampedModel: https://django-model-utils.readthedocs.io/en/latest/models.html#timestampedmodel
+
+- Time should be stored in UTC time by setting :code:`USE_TZ=True`_. in your python config.
+.. _USE_TZ=True: https://docs.djangoproject.com/en/2.2/topics/i18n/timezones/#overview
+
+-  If for some reason you can not inherit from :code:`TimeStampedModel`_. use the following naming conventions:
+
+   -  Created date should be named: “created”
+
+   -  Updated date should be named: “modified”
+
+Foreign keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Data should be joined using primary keys wherever possible
+
+-  Foreign keys should use a naming convention of :code:`<object_name>_id` where object name is the name of the table of the foreign key relationship.
+
+-  Do not join on things such as username, email address, or other dimensions of data that may change over time
+
+-  Do not join on PII
+
+- Joining between IDAs should be done by using a universally unique identifier (UUID)
+
+- In Django use `Attributes for fields with relations`_. to identify and link models with relationships.
+.. __Attributes for fields with relations: https://docs.djangoproject.com/en/2.2/ref/models/fields/#module-django.db.models.fields.related
+
+History for models involved with enrollments, courses and course metadata, or data involving financial payments and transactions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  History using django-simple-history_. .. _django-simple-history: https://django-simple-history.readthedocs.io/en/latest/
+
+-  Remember to `backfill history`_. for existing models.
+
+.. _backfill history: https://django-simple-history.readthedocs.io/en/latest/quick_start.html#existing-projects
+
+-  Where Django simple history is not an option, the following data should be captured:
+
+   -  Fields that were changed
+
+   -  Date & time of the change
+
+   -  The foreign key of the user who initiated the change
+
+Use the correct data type for a column
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Don’t use a :code:`IntegerField` when a :code:`BooleanField` would do.
+
+-  Use :code:`BigIntegerField`_. for your foreign keys
+
+-  Don’t store an Integer field as :code:`CharField`_..
+.. _CharField: https://docs.djangoproject.com/en/2.2/ref/models/fields/#charfield
+
+- Store UUID's as :code:`UUIDField`_. with a max length that matches the max length of the UUID.
+.. _UUIDField: https://docs.djangoproject.com/en/2.2/ref/models/fields/#uuidfield
+
+
+Each column in a table should only store a single fact or dimension
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- If a column could be a mix of integer data and character data it is best to store these items as 2 different fields in the database
+
+Models should be Annotated for PII
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  All models in the Open edX ecosystem should be tagged for PII using `code annotations`_. by following OEP-30_.
+.. _code annotations: https://github.com/edx/code-annotations>
+.. _OEP-30: https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst
+
+
+Have sane default values
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- A model should have default values that make sense for the application
+
+- :code:`CharField`_. Should be defined with the :code:`null=True`_. option.
+.. _null=True: https://docs.djangoproject.com/en/2.2/ref/models/fields/#null
+   - This is to improve post processing. Defaulting :code:`CharField`_. to null enables us to better derive the intent of the user. If the field is null no intent was made to enter that field. If the field is blank a string was entered and was modified later by the user.
+
+- For example if you are adding a boolean to flag that a learner has not yet activated their account, the default value should be set to False, not None.
+
+The database layer should preserve uniqueness
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- If a model needs to preserve uniqueness between many fields use :code:`unique_together`_.
+.. _unique_together: https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together
+
+
+
+Best Practices
+--------------
+These practices are designed to help teams create rich and efficient data models within the OpenedX ecosystem.
+They are not standards but guidelines to help teams think about how they are storing data.
+
+Columns should store only exactly what the name describes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- The name of a column in a table should accurately describe the data in that table.
+
+- If a column is named course_id it should only store the course_id. Not the course_key, not a numeric value, not a timestamp. Just the course_id. Conversely if a column is named course_run_key it should store the course run key, not the course_id.
+
+
+Deleting data
+~~~~~~~~~~~~~
+
+-  It is better to have a column to mark the record as inactive than to
+      remove the data from the system using the SQL delete keyword. These models should use Django's
+      :code:`SoftDeletableModel`_.
+.. _SoftDeletableModel: https://django-model-utils.readthedocs.io/en/latest/models.html#softdeletablemodel
+
+
+-  Please note that GDPR may require that data be deleted. If a field
+         is determined to contain PII and falls under the realm of GDPR,
+         That data should be deleted from the system, or obfuscated from the system. `For more information about GDPR and how to delete user data from edx please refer to this documentation`_.
+.. _For more information about GDPR and how to delete user data from edx please refer to this documentation: https://openedx.atlassian.net/wiki/spaces/PLAT/pages/930021733/User+Retirement+Tutorial+for+Developers
+
+
+Don’t trap the data
+~~~~~~~~~~~~~~~~~~~
+
+-  Each piece of information should have its own column. Avoid storing
+      data in blob fields or as JSON in the database.
+
+- Another example is a concatenated string with a separator. It is best to treat these data items in 2 distinct fields.
+
+-  Don’t store encoded (pickle, json, other) objects in the database! If
+      you need to run the python environment to decode the data, analysts who use SQL will have a difficult time querying and decoding this data.
+
+Store everything
+~~~~~~~~~~~~~~~~
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database or add history the answer is almost always yes. Store the data and add history. It can be removed later if it is found unnecessary.
+
+-  Still not sure? The default answer is yes.
+
+Use methods on models to access data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- CRUD operations should access models via methods on models (where they exist), instead of querying managers directly.
+
+-  For example, prefer creating something like
+      CourseEnrollment.is_enrolled(...) rather than having views check
+      CourseEnrollment.objects.filter(...).exists().
+
+-  This allows us to more easily make internal representation changes
+      like switching to using a “deleted” flag instead of deleting the
+      row.
+
+-  This also reduces the likelihood that people will query models in a
+      non-performant way (e.g. sorting by an unindexed field).
+
+Enforce logical constraints at the database layer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Don’t allow impossible states to be represented in the database.
+
+-  If your code expects a 1:1 relationship, use Django's :code:`Unique`_. instead of trying to enforce the constraint in
+      Python.
+.. _Unique: https://docs.djangoproject.com/en/2.2/ref/models/fields/#unique
+
+-  Python will not save you from race conditions. Database constraints will.
+
+-  For example, an enrollment should have a unique constraint on
+      (course_id, user_id), since a given user should only have one
+      enrollment per course. In this case you should use Django's :code:`unique_together`_
+
+Keep indexes to a minimum
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Create indexes only on the fields necessary to make queries performant
+
+-  Keep in mind that indexes cost space and have their own set of performance concerns.
+
+-  Over-indexing data could actually make the database less performant (slower writes/updates)
+
+
+
+Developer Responsibility
+========================
+
+It is the responsibility of the developer to adhere to all of the
+standards in the Data Modeling Standards section of this document.
+
+Code Reviewer Responsibility
+============================
+
+The code reviewer is responsible for ensuring the standards set forth in
+the Data Modeling Standards section of this document are met.
+
+Responsibility for Third-party Service Integrations
+===================================================
+
+Adhere to the same standards.
+
+Backward Compatibility
+======================
+
+Data models that are not within the standards of this document do not need to be updated to adhere to OEP-38 standards.
+
+
+
+                 


oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+   -  Do not use composite based primary keys. Use a primary key column
+
+-  Updated and created timestamps
+
+   -  Use UTC time.
+
+   -  Created date should be named: “created_date”
+
+   -  Updated date should be named: “modified_date”
```
ormsbee: A lot of our models use `TimeStampedModel`, which has these fields as `created` and `modified`. Is it fine to stick with that convention?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+information. The purpose of Business Intelligence is to support better
+business decision making.
+
+Composite Key - in the context of relational databases, a composite key
+is a combination of two or more columns in a table that can be used to
+uniquely identify each row in the table. Uniqueness is only guaranteed
+when the columns are combined; when taken individually, the columns do
+not guarantee uniqueness.
+
+CUD - Create, Update, Delete. These are some of the actions that may
```
ormsbee: Is this intentionally omitting the Read in CRUD?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+corresponding information systems in organizations.
+
+Data Modeling Standards - The most basic and standard design principles
+for data modeling. These standards must be adhered to when creating new
+models or updating existing models in the edX ecosystem.
+
+IDA - “Independently Deployed Application”. Similar to a microservice or
+stand alone application.
+
+OpenedX Ecosystem
```
ormsbee: Nit: Open edX

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+practices outlined in this document ensures our data is approachable for
+experienced engineers and new hires alike and allows us to leverage 3rd
+party tools more easily to assist in the analysis phase.
+
+Don’t make me think
+~~~~~~~~~~~~~~~~~~~
+
+Steve Krug, the author of “Don’t Make Me Think”, says: “Your objective
+should always be to eliminate instructions entirely by making everything
+self-explanatory, or as close to it as possible.” This principal should
```
ormsbee: Nit: "principal" -> "principle"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+-  Each piece of information should have its own column. Avoid storing
+      data in blob fields or as JSON in the database.
+
+-  Don’t store encoded (pickle, json, other) objects in the database! If
+      you need to run the python environment to decode the data, the
+      data is worthless.
+
+Store everything
+
+-  Storage is cheap!
```
ormsbee: Maybe we could provide some ballpark guidelines for how much is definitely okay to store in a table, both in terms of number of rows and overall size in bytes? Storage is cheap, but if back-of-the-envelope math shows that a new database table will grow to 30 terabytes in the first year, there needs to be a followup conversation. (I have definitely encountered at least one proposal in the last year that would have done that.)

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+-  The correct data type for a column.
+
+   -  Don’t use a long when an Int or bool would do.
+
+   -  Use Integers for your keys
+
+   -  Don’t store an int as varchar
+
+-  Don’t mix types in a column
```
ormsbee: What does this mean?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+-  Annotations
+
+   -  All models in the OpenedX ecosystem should be tagged for PII
```
ormsbee: Nit: "Open edX"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+edX uses data to decide which people should receive marketing emails,
+who passes or fails a course, or how much to pay our partners. Our
+partners use our data to target users who fit certain criteria and to
+refine and improve their courses. The data we collect today may be used
+to gain new educational insights we have yet to realize. Decision making
+at edX should be data-driven and based on this collected data.
+
+This data is one of our most valuable assets and it should be and it
+should be a first-order concern. Save everything (disk is cheap). Fire
+lots of events. The more data we are able to collect about users,
```
kdmccormick: > Fire lots of events.

Is there a such thing as over-eventing? For example, if a single API view fires three redundant events, could that actually hinder data analysis?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+self-explanatory, or as close to it as possible.” This principal should
+also be applied to data modeling. Data models should have descriptive
+names clearly identifying the data that the model holds. The field names
+should make sense to most people familiar with the domain, without
+needing much clarifying documentation.
+
+Prioritize ease of analysis
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For every hour of engineer-time spent creating a data model, several
```
kdmccormick: "several thousand" strikes me as hyperbolistic; how about "many more"?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+Think about the person who will analyze this data later and the person
+who manages the system day to day. (That person could possibly be future
+you!) Try to do as much careful design up front to make your quality of
+life better later.
+
+History is best (but events are a close second)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Accuracy is best when history about a change is captured at the database
```
kdmccormick: > Accuracy is best when history about a change is captured at the database level when the change happens.

Consider rewording.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+continuity should be stored historically where the change is made,
+preferably in the same system as the system of record. This allows us to
+reconstruct the data at different points with much greater reliability.
+
+If, for some reason, it is not possible to create history at the time of
+entry (for example, if we expect the table to become too large or the
+writes to be too frequent), We recommend that an event be issued in its
+place.
+
+Fire events for CUD operations
```
kdmccormick: Are there any recommendations for eventing best-practices at edX?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+All models in the edX ecosystem should have:
+
+-  A Primary Key
+
+   -  Should be named: “id”
+
+   -  Use numeric based identifiers
+
+   -  A key size large enough so the keys will not run out for a very
```
kdmccormick: For those of us not experienced with defining custom `id` fields, is the default Django `id` generally sufficient?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
+   -  Created date should be named: “created_date”
+
+   -  Updated date should be named: “modified_date”
+
+-  Use a numeric foreign key(s)
+
+   -  Data should be joined on other foreign keys
+
+   -  Do not join on things such as username, email address, or other
```
kdmccormick: What about entities that exist across multiple IDAs (e.g. course runs, programs, etc.)? Should we key on human-readable string identifiers, UUIDs, or something else?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+      CourseEnrollment.objects.filter(...).exists().
+
+-  This allows us to more easily make internal representation changes
+      like switching to using a “deleted” flag instead of deleting the
+      row.
+
+-  This also reduces the likelihood that people will query models in a
+      non-performant way (e.g. sorting by an unindexed field).
+
+Enforce logical constraints at the database layer.
```
kdmccormick: In Django, specifying `unique=True` also creates an index. Are enforcing logical constraints and having the minimum number of indexes at odds?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+Data models that are not within the standards of this document should be
+updated the next time a team needs to make changes to that model
+
+Some things are exempt from this process:
+
+-  Do not update primary key types without business/technical
+      justification, as there may be many downstream changes that will
+      need to change as well.
+
+Reference Implementation
```
kdmccormick: I think illustrative examples of good and bad models (either by linking or inlining some code) would make the OEP more impactful.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+Store everything
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database,
+      add history, or fire an event, the answer is almost always yes
+
+-  Still not sure? The default answer is yes.
+
+Views should access models via methods on those models, instead of
```
kdmccormick: Consider mentioning the edx-platform pattern of putting all business logic in `api.py` and making `models.py` an internal-only file, which is formalized in [this ADR](https://github.com/edx/edx-platform/blob/master/docs/decisions/0002-inter-app-apis.rst)

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+-  Where Django simple history is not an option, the following data
+      should be captured:
+
+   -  Fields that were changed
+
+   -  Date & time of the change
+
+   -  ID of the user who initiated the change
+
+-  The correct data type for a column.
```
kdmccormick: Since this OEP is pretty Django-specific, consider using the Django ORM terms here: `BigIntegerField`, `IntegerField`, `BooleanField`, `ForeignKey`, `CharField`, `TextField`, etc.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+lots of events. The more data we are able to collect about users,
+behavior, and system state now, the more opportunities we’ll have to
+improve our decision-making in the future.
+
+Use industry best practices
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Data stored in the OpenEdX ecosystem should adhere to industry best
+practices. For example, since an industry best practice uses numerical
+identifiers to identify rows of data, the edX ecosystem should also use
```
kdmccormick: > should also do so.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+-  Don’t mix types in a column
+
+-  Annotations
+
+   -  All models in the OpenedX ecosystem should be tagged for PII
+      using `code
+      annotations <https://github.com/edx/code-annotations>`__\  following \ `OEP-30 <https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst>`__
+
+Best Practices 
+~~~~~~~~~~~~~~~
```
kdmccormick: I remember you talking about some of the merits of denormalization in terms of ability to analyze data. Are there any best practices regarding (de)normalization that might go here?

oeps/oep-38-Data-Modeling.rst
ormsbee: +1 -- we should make sure it's explicitly called out. The default `id` gives you a signed 4-byte int, so the max is (2^31 - 1). Sufficient for most of our tables, but definitely not all.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
...
+
++---------------+------------------------------------------------------------+
+
+| Type | Architecture |
+
++---------------+------------------------------------------------------------+
+
+| Created | 2018-04-19 |
+
++---------------+------------------------------------------------------------+
```
nasthagiri: @macdiesel Thanks for authoring and sending this out.  Please specify a beginning and end date for the review period, as soon as possible.  Also, you or @albemarle will want to publish the review of this OEP to the appropriate #open-edx slack channels (as documented in OEP-1).  Thanks.

oeps/oep-38-Data-Modeling.rst
ormsbee: I guess it could be seen to be at odds, though if there's a unique constraint, it's also extremely likely that you'd want to search by it as well. In either case, I think we want to default to "add the constraints necessary to ensure data integrity" and make compromises on that only under unusual circumstances.

oeps/oep-38-Data-Modeling.rst
brianhw: CUD only applies to CoW databases...

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+happen to data in a database system.
+
+Data Best Practices - These practices are designed to help teams create
+rich and efficient data models within the edX ecosystem. They are not
+standards but guidelines to help teams think about how to store data.
+
+Data Dimension - A Data Dimension is a set of data attributes pertaining
+to something of interest to a business. Dimensions are things like
+"customers", "products", "stores" and "time". For users of Data
+Warehouses, data dimensions are entry points to numeric facts (e.g.
```
brianhw: nit: I didn't really get the "entry points to numeric facts" reference.   Maybe enough to replace "numeric facts" with "'Data Facts'", to draw the contrast with the "Data Dimension".

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+------------------
+
+Used in this document, the phrase "OpenedX Ecosystem" includes the
+services that comprise and support the running of an OpenedX
+installation (e.g.,. edx.org). This includes:
+
+- The LMS and Studio
+
+- IDAs such as Notes, Ecommerce, and Forums
```
brianhw: nit: do we still have Notes?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+~~~~~~~~~~~~~~~~
+
+edX uses data to decide which people should receive marketing emails,
+who passes or fails a course, or how much to pay our partners. Our
+partners use our data to target users who fit certain criteria and to
+refine and improve their courses. The data we collect today may be used
+to gain new educational insights we have yet to realize. Decision making
+at edX should be data-driven and based on this collected data.
+
+This data is one of our most valuable assets and it should be and it
```
brianhw: nit: "and it should be" appears twice

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+.. \_OEP-002:
+https://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html
+
+Abstract
+
+========
+
+In order to obtain the highest possible value from data collected in the
+edX ecosystem, this document attempts to provide guiding thoughts and
```
brianhw: There's a lot of cases of using "edX ecosystem" and "OpenedX ecosystem".  Are these used intentionally to define different things, or are they intended to be interchangeable?   If the latter, it would be good to choose one and use it consistently.  If the former, then it probably needs a separate definition in the glossary.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+who manages the system day to day. (That person could possibly be future
+you!) Try to do as much careful design up front to make your quality of
+life better later.
+
+History is best (but events are a close second)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Accuracy is best when history about a change is captured at the database
+level when the change happens. Changes in data relating to business
+continuity should be stored historically where the change is made,
```
brianhw: What does "relating to business continuity" mean?   Things that pertain to edX not going out of business?   It might be helpful to provide examples of tables that would require history, and some that would not require history.  For example, important "dimensions" like courses are good to have history, as are things like "current enrollment state".   Tables that are already append-only tables representing some kind of event would not benefit from history.   But do we keep history on anything that supports "update"?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+preferably in the same system as the system of record. This allows us to
+reconstruct the data at different points with much greater reliability.
+
+If, for some reason, it is not possible to create history at the time of
+entry (for example, if we expect the table to become too large or the
+writes to be too frequent), We recommend that an event be issued in its
```
brianhw: royal "We"?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+   -  You should use as many indexes as you need to make the application
+         performant and no more.
+
+-  History for models involved with enrollments, or anything that may
+      have an effect on financial payments or transactions.
+
+-  History using django-simple-history
+
+   -  Should capture the ID of the user who made the change
```
brianhw: Should there be a recommendation about how to deal with ManyToManyField table values?

oeps/oep-38-Data-Modeling.rst
brianhw: I assume it means something like storing numbers and strings in the same column (presumably as strings).

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+~~~~~~~~~~~~~~~
+
+Deleting data:
+
+-  It is better to have a column to mark the record as inactive than to
+      remove the data from the system using the SQL delete keyword.
+
+   -  Please note that GDPR may require the data be deleted. If a field
+         is determined to contain PII and falls under the realm of GDPR,
+         That data should be deleted from the system.
```
brianhw: nit: "that"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+      using `code
+      annotations <https://github.com/edx/code-annotations>`__\  following \ `OEP-30 <https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst>`__
+
+Best Practices 
+~~~~~~~~~~~~~~~
+
+Deleting data:
+
+-  It is better to have a column to mark the record as inactive than to
+      remove the data from the system using the SQL delete keyword.
```
brianhw: Also note the impact of deletions on data in other models.   Models that use ForeignKey with on_delete=models.CASCADE will lose their values when the object the foreign key points to is deleted.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+-  Guiding Principles
+
+-  Data Modeling Standards
+
+-  Best Practices
+
+Guiding Principles - This section is intended to give the reader a
+framework to think about data modeling at edX. While no specific
+instructions are given in this section, it should act as a framework to
+help the reader think about data modeling on the whole at edX.
```
bmedx: This seems to repeat itself and "no specific instructions" seems to contradict what we're doing with "Data modeling standards".

oeps/oep-38-Data-Modeling.rst
bmedx: @brianhw We do!

oeps/oep-38-Data-Modeling.rst
bmedx: @kdmccormick I would say that there is so a thing as over-eventing, but that it is a better problem to have than not having the necessary data to make a decision. People should make reasonable attempts to not duplicate things, but if it's unclear that an event would be a duplicate I think we would rather lean towards having it.

oeps/oep-38-Data-Modeling.rst
bmedx: Might want to explicitly call out `UnsignedBigIntAutoField` for Django models here if that's what we want to be used.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+-  A Primary Key
+
+   -  Should be named: “id”
+
+   -  Use numeric based identifiers
+
+   -  A key size large enough so the keys will not run out for a very
+         long time.
+
+   -  Do not use composite based primary keys. Use a primary key column
```
bmedx: In this case do we want to recommend, ex, `unique_together` to preserve integrity?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+-  The minimum number of indexes possible to make the table/queries
+      performant
+
+   -  Indexes cost space and have their own set of performance concerns.
+
+   -  Over-indexing data could actually make the database less
+         performant (slower writes/updates)
+
+   -  You should use as many indexes as you need to make the application
```
bmedx: This reads as a duplicate of "The minimum number of indexes possible to make the table/queries performant"

oeps/oep-38-Data-Modeling.rst
bmedx: If we do, should we call out other limitations we've run into, like AutoSlugField?

oeps/oep-38-Data-Modeling.rst
bmedx: Might want to mention that if data can't be deleted it should be obfuscated, maybe a description of when that data should be removed (when no longer used, on retirement?)

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+      data is worthless.
+
+Store everything
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database,
+      add history, or fire an event, the answer is almost always yes
+
+-  Still not sure? The default answer is yes.
```
bmedx: Maybe "Ask, but the answer will probably be yes"?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+-  Event data
+
+Motivation
+
+==========
+
+The goal of this OEP is to provide a framework about how to think about
+data modeling at edX and provide guidelines on what should and should
+not be done when creating data models.
```
pwnage101: It seems like the purpose of the motivation section might be to provide some context and background that clarifies the events that culminated to the authoring of this document, or a description of scenarios that it is indented to prevent. Otherwise, it just feels like a reiteration of the abstract.

oeps/oep-38-Data-Modeling.rst
pwnage101: I think we would usually say "dimensions" are applied to "metrics".  I would perceive facts as perhaps an immutable subset of metrics, but "metrics" is more broadly applicable.

oeps/oep-38-Data-Modeling.rst
bderusha: Can we also have a recommendation for how to access only active records by default.  Whether that is django object managers that default filter the active field or managing database views that select only the active fields or something.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+Views should access models via methods on those models, instead of
+querying managers directly.
+
+-  For example, prefer creating something like
+      CourseEnrollment.is_enrolled(...) rather than having views check
+      CourseEnrollment.objects.filter(...).exists().
+
+-  This allows us to more easily make internal representation changes
+      like switching to using a “deleted” flag instead of deleting the
```
bderusha: Flag is an ambiguous term.  Can we have a stance on flags being booleans with default values, nullable booleans, or timestamps (EG delete_at. Null for active records, timestamped deletes otherwise)

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database,
+      add history, or fire an event, the answer is almost always yes
+
+-  Still not sure? The default answer is yes.
+
+Views should access models via methods on those models, instead of
+querying managers directly.
```
bderusha: Is this the recommendation for "simple gets" as well, like `User.objects.get(id=user_id)`?

oeps/oep-38-Data-Modeling.rst
feanil: If we go with this change adding a code block with a sample model will make it so that more people can easily do the right thnig.

ormsbee: I'm not strongly advocating either way, but keep in mind that for InnoDB, the primary key is copied into every index on the table.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+writes to be too frequent), We recommend that an event be issued in its
+place.
+
+Fire events for CUD operations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We highly recommend firing events for any create, update or delete (CUD)
+operations that happen within the OpenedX ecosystem. The cost of firing
+and storing these events is low, and will allow teams in the future to
+reconstruct the state of the application at various times.
```
pwnage101: In the previous section, we recommend firing events as a solution to being unable to record history in the database, but in this section we appear contradict that by recommending firing events regardless of ability to record history in the database.  I feel like it's worth clarifying when developers should be expected to fire events.

Perhaps in this section you're referring to front-end events that don't necessarily pertain to back-end database table updates, or backend subroutines that deserve an singular event but would otherwise cause multiple table CUDs.

oeps/oep-38-Data-Modeling.rst
pwnage101: FWIW, I'd argue that `created` and `modified` is better than `created_date` and `modified_date` when they are timestamps (as opposed to dates).  I think most of the time we are talking about timestamps.  Of course, this is already decided if we just use TimeStampedModel.

oeps/oep-38-Data-Modeling.rst
pwnage101: I think the major thing to avoid is joining on PII.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+Data Modeling Standards
+-----------------------
+
+All models in the edX ecosystem should have:
```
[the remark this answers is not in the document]
ormsbee: Also recommendations on field sizes -- e.g. if you use utf8mb4, you're limited to varchar 181 on an indexed field, etc.

oeps/oep-38-Data-Modeling.rst
pwnage101: Also, there is an effort to standardize on the LMS user ID and make that value exist in every IDA that has an auth_user table of learners.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+   -  Date & time of the change
+
+   -  ID of the user who initiated the change
+
+-  The correct data type for a column.
+
+   -  Don’t use a long when an Int or bool would do.
+
+   -  Use Integers for your keys
```
pwnage101: Might be worth reminding the reader of `UnsignedBigIntAutoField` here, and its merits.

oeps/oep-38-Data-Modeling.rst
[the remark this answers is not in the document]
ormsbee: I don't know if this is a pattern we want to follow, but to allow Blockstore to live in a shared MySQL env that may or may not be configured with the appropriate defaults, we include this migration to be run first: https://github.com/open-craft/blockstore/blob/master/blockstore/apps/mysql_unicode/migrations/0001_initial.py

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+   -  Use numeric based identifiers
+
+   -  A key size large enough so the keys will not run out for a very
+         long time.
+
+   -  Do not use composite based primary keys. Use a primary key column
+
+-  Updated and created timestamps
+
+   -  Use UTC time.
```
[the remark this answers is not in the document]
ormsbee: I don't think MySQL DATETIME actually supports timezone information at all, so we should be very opinionated on storing in UTC and making sure the `USE_TZ` setting is true.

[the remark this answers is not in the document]
ormsbee: It would be easy for someone unfamiliar with Django to not realize they had to tweak the setting and end up storing DATETIMEs in the current timezone (which could be configured separately as UTC, but then the code behavior would be wrong if anyone changed the timezone later as part of their configuration).

oeps/oep-38-Data-Modeling.rst
mulby: This is the only doc I know of: https://openedx.atlassian.net/wiki/spaces/AN/pages/13205895/Event+Design+and+Review+Process

albemarle: Review decision: We will do a separate OEP for eventing.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,438 @@
+.. vim: syntax=rst
+
+===============================
```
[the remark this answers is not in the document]
robrap: Some tips for the RST format that you can pull from [OEP-19 (raw format)](See this https://raw.githubusercontent.com/edx/open-edx-proposals/master/oeps/oep-0019-bp-developer-documentation.rst): 
* See the simpler alternate format for the initial table of metadata. 
* Reminder to not use line breaks for long lines.

oeps/oep-38-Data-Modeling.rst
davidjoy: Is it worth recommending having a `deleted_date` for soft deletes in addition to the flag?

oeps/oep-38-Data-Modeling.rst
macdiesel: It was my intent to use openedx ecosystem. I have replaced all the references to edX ecosystem with OpenedX ecosystem..

oeps/oep-38-Data-Modeling.rst
macdiesel: I have removed this section.

oeps/oep-38-Data-Modeling.rst
macdiesel: Removed the line about firing events.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+Don’t make me think
+~~~~~~~~~~~~~~~~~~~
+
+Steve Krug, the author of “Don’t Make Me Think”, says: “Your objective
+should always be to eliminate instructions entirely by making everything
+self-explanatory, or as close to it as possible.” This principal should
+also be applied to data modeling. Data models should have descriptive
+names clearly identifying the data that the model holds. The field names
+should make sense to most people familiar with the domain, without
+needing much clarifying documentation.
```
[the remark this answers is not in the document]
macdiesel: I do agree that it can be hard to capture all of the nuance and complexity of some items. That said I do think that this is still something that we should strive for where possible. 

Also let me point out this part: "The field names should make sense to most people **familiar with the domain**, without needing **much** clarifying documentation." (emphasis mine)

I am going to leave this in here as it is unless we receive more feedback that people do not think this is useful/usable.

oeps/oep-38-Data-Modeling.rst
macdiesel: There was a lot of back and forth on this line before I published the OEP. I agree that it turned out a bit hyperbolic. While I am unable to back up this fact with exact numbers I'm not sure that we should try. I like your suggestion @kdmccormick.

@mulby I do not think this section suggests that models should be optimized for analysis. I do however believe that many more hours are spent analyzing the data. A little more thought about the analysis up front could save us a lot of time later.

macdiesel: Ok I take that back I do see I said Prioritize. How about consider?

oeps/oep-38-Data-Modeling.rst
macdiesel: This section has been removed from the OEP.

oeps/oep-38-Data-Modeling.rst
macdiesel: I've reworded this section a bit to talk about the areas which are the highest priority for history.

oeps/oep-38-Data-Modeling.rst
macdiesel: Sure space is cheap, but memory always isn't. These rows will eventually be loaded in to memory for analysis and joined on. 


I don't think that blanket usage of UnsignedBigIntAutoField is the right thing here. I think that those designing the system need to think about what the correct choice is for the application they are designing.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+   -  A key size large enough so the keys will not run out for a very
+         long time.
+
+   -  Do not use composite based primary keys. Use a primary key column
+
+-  Updated and created timestamps
+
+   -  Use UTC time.
+
+   -  Created date should be named: “created_date”
```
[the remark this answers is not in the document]
macdiesel: I have updated the OEP to direct the user to use TimeStampedModel.

oeps/oep-38-Data-Modeling.rst
macdiesel: I have updated the OEP to direct the user to use TimeStampedModel.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+   -  Updated date should be named: “modified_date”
+
+-  Use a numeric foreign key(s)
+
+   -  Data should be joined on other foreign keys
```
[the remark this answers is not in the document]
macdiesel: Thanks for pointing this out I've cleaned up the wording here. The idea is that you should join on other Primary Keys instead of joining on things such as username.

oeps/oep-38-Data-Modeling.rst
macdiesel: @trevorbass Do you have any thoughts about ID's shared between different systems. I was thinking you recommendation here would be to prefer numeric based identifiers over strings or UUIDs want wanted to check in with you first to see if we are aligned.

oeps/oep-38-Data-Modeling.rst
macdiesel: Line has been struck.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+   -  Indexes cost space and have their own set of performance concerns.
+
+   -  Over-indexing data could actually make the database less
+         performant (slower writes/updates)
+
+   -  You should use as many indexes as you need to make the application
+         performant and no more.
+
+-  History for models involved with enrollments, or anything that may
+      have an effect on financial payments or transactions.
```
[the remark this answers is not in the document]
macdiesel: I've added a bit here and re-worded the line.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+-  History using django-simple-history
+
+   -  Should capture the ID of the user who made the change
+
+-  Where Django simple history is not an option, the following data
```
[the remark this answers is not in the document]
macdiesel: After our conversation during the Arch meeting I believe that this should be covered under the eventing OEP. 

That said it certainly is an option. The upside is that we would capture nearly everything that happens. The down side is that it becomes a bit harder to analyze the data. I'll bring this up as a conversation with the Data team and see what they think.

oeps/oep-38-Data-Modeling.rst
macdiesel: Still not sold on everything being an UnsignedBigIntAutoField

oeps/oep-38-Data-Modeling.rst
macdiesel: Some options here for rewording this:
- Do not store multiple data dimensions or facts in a single column
- Each column in a table should only store a single fact or dimension

oeps/oep-38-Data-Modeling.rst
macdiesel: Yeah I should have expanded this section further. I will rework this.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+-  Don’t store encoded (pickle, json, other) objects in the database! If
+      you need to run the python environment to decode the data, the
+      data is worthless.
+
+Store everything
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database,
+      add history, or fire an event, the answer is almost always yes
```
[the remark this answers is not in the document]
macdiesel: I've updated this text to be more clear.

oeps/oep-38-Data-Modeling.rst
macdiesel: I have reworded this line.

oeps/oep-38-Data-Modeling.rst
macdiesel: I feel this is an architecture concern and not a data modeling concern.

oeps/oep-38-Data-Modeling.rst
macdiesel: Dave is your suggestion here to change "Enforce logical constraints at the database layer." to "add the constraints necessary to ensure data integrity"?

albemarle:
# Notes on 8/5 meeting at edX

## Action Items

- who will write the eventing OEP?

## Notes

this is a best practice OEP, can evolve

flesh out how to fire an event
- **decision:** separate eventing OEP, remove eventing from this OEP

more code samples / good examples will help adoption
- concern: will it get outdated?
- **decision:** yes, add more code examples
- do not use json blobs

add: avoid duplicating data between applications - single source of truth

join on foreign keys
- distributed foreign key relationship?

Q: is this for IDA databases as well? (A: yes)

how to verify compliance?
- add a section of “work that would help compliance”?
- rollout plan already exists to help reinforce this

add “don’t delete columns” ?
- we want data to be able to evolve
- can we remove the column from the model but not from the backing table?
- be more opinionated on soft deletes of rows
- consumers of columns: application but also analytics
- use case for a DB view? how does that work with Django?

how long to keep history?
- even if we don’t have a plan to solve it, call out that the issue exists
- Courseware Student Module History
- not just financial stuff, define areas where we need to store history better
- store at application level instead?
- CSMH table never used? just once to debug an issue
- big tables should live in warehouse
- historically we relied on event log (what the academic research community uses)
- slow-changing, important data
- can we hard delete rows that are application only?
- use an annotation to identify the different types of data?

**decision:** let’s meet one more time

oeps/oep-38-Data-Modeling.rst
macdiesel: It is, There is a section below that deals with when to write history. Changes in a tables state only happens during Create, update and delete actions. Hence the omission of Read.

oeps/oep-38-Data-Modeling.rst
macdiesel: Reworded this.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+------------------
+
+Data is an asset
+~~~~~~~~~~~~~~~~
+
+edX uses data to decide which people should receive marketing emails,
+who passes or fails a course, or how much to pay our partners. Our
+partners use our data to target users who fit certain criteria and to
+refine and improve their courses. The data we collect today may be used
+to gain new educational insights we have yet to realize. Decision making
```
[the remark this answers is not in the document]
macdiesel: I reworked some language here and took your suggestion. Thanks!

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,442 @@
...
+
+-  BI teams to update reports that utilize this data,
+
+-  Devops to deploy migrations to large tables, with the risk of
+      potential downtime.
+
+Due to the potentially high cost of changing a model, it pays to get it
+as correct as possible the first time.
+
+Think about the person who will analyze this data later and the person
```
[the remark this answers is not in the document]
macdiesel: I do agree that this is beyond the scope of what we would like to do in this OEP. That said I think that more show and tell around data analytics is a good idea.

oeps/oep-38-Data-Modeling.rst
macdiesel: Ended up going with BigAutoField as the recommended field type.

oeps/oep-38-Data-Modeling.rst
macdiesel: Excellent idea.

oeps/oep-38-Data-Modeling.rst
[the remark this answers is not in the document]
macdiesel: I have added some text to the OEP informing the reader they should set USE_TZ=True in their django project.

oeps/oep-38-Data-Modeling.rst
macdiesel: Went with UUID.

oeps/oep-38-Data-Modeling.rst
macdiesel: Updated.

oeps/oep-38-Data-Modeling.rst
macdiesel: In the case where there is a method on the model to do the thing you want to do you should use that. The thought here being that that model's methods may encapsulate logic such as firing events or deleting other objects, etc... that could be missed by using the objects directly. I'll clean up the language a bit here.

oeps/oep-38-Data-Modeling.rst
macdiesel: I have updated the OEP to direct the reader to use django_utils [SoftDeletableModel](https://django-model-utils.readthedocs.io/en/latest/models.html#softdeletablemodel)

macdiesel: I added an example of adding a default value to a boolean and having sane defaults. That said I think there are cases where having None, True and False for values on a boolean are valid (None: I haven't viewed the thing, True: I agree, False: I disagree). I don't think we should take a hard stance here.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -237,20 +233,17 @@ All models in the OpenedX ecosystem should have:
...
-
-   -  Do not join on things such as username, email address, or other
-         dimensions of data that may change over time
+   -  Do not join on things such as username, email address, or other dimensions of data that may change over time
 
    -  Do not join on PII
 
--  The minimum number of indexes possible to make the table/queries
-      performant
+   - Joining between IDAs should be done by using an Universally unique identifier (UUID)
```
albemarle: I suggest "a universally unique identifier (UUID)"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -293,7 +286,12 @@ Deleting data:
 
    -  Please note that GDPR may require that data be deleted. If a field
          is determined to contain PII and falls under the realm of GDPR,
-         That data should be deleted from the system.
+         That data should be deleted of obfuscated from the system.
```
albemarle: Suggestion: "should be deleted from the system, or obfuscated"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -293,7 +286,12 @@ Deleting data:
...
+
+Preserving uniqueness:
+
+   - If a model needs to preserve uniquness between many fields use :code:`unique_together`_.
```
albemarle: Typo: uniqueness

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+.. _For more information about GDPR and how to delete user data from edx please refer to this documentation: https://openedx.atlassian.net/wiki/spaces/PLAT/pages/930021733/User+Retirement+Tutorial+for+Developers
+
+
+Don’t trap the data
+~~~~~~~~~~~~~~~~~~~
+
+-  Each piece of information should have its own column. Avoid storing
+      data in blob fields or as JSON in the database.
+
+- Another example is a concatenated string with a seperator. It is best to treat these data items in 2 distinct fields.
```
albemarle: Typo: separator

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+
+-  Don’t store encoded (pickle, json, other) objects in the database! If
+      you need to run the python environment to decode the data, analyists who use SQL will have a difficult time querying nd decoding this data.
```
albemarle: Typo: nd -> and

oeps/oep-38-Data-Modeling.rst
```diff
@@ -65,39 +46,39 @@ This OEP explicitly does not cover:
...
 needed to support the business processes within the scope of
 corresponding information systems in organizations.
 
-Data Modeling Standards - The most basic and standard design principles
+**Data Modeling Standards** - The most basic and standard design principles
 for data modeling. These standards must be adhered to when creating new
 models or updating existing models in the OpenedX ecosystem.
 
-IDA - “Independently Deployed Application”. Similar to a microservice or
+**IDA** - “Independently Deployed Application”. Similar to a microservice or
```
albemarle: CUD is defined without quotes around the phrase, so for consistency IDA should be as well.

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+entry (for example, if we expect the table to become too large or the
+writes to be too frequent), it is recommended that an event be issued in its
+place.
+
+
+Data Modeling Standards
+-----------------------
+
+The standards below are designed to ensure edX can gain the highest value and insights from the data.
+The application of these standards is the most basic level of support that all application in the OpenedX ecosystem
```
albemarle: "all applications", "should adhere to" (or "should provide")

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+should adhere. When creating new applications or models please ensure the models being created comform to the following.
```
albemarle: Typo: conform

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- If a model needs to preserve uniquness between many fields use :code:`unique_together`_.
+.. _unique_together: https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together
+
+
+
+Best Practices
+--------------
+These practices are designed to help teams create rich and efficient data models within the OpenedX ecosystem.
```
albemarle: Open edX?  (I think.)

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+They are notstandards but guidelines to help teams think about how they are storing data.
```
albemarle: "not standards"

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database or add history the answer is almost always yes. Store the data and add history. It can be removed later if it is found unnecessary.
+
+-  Still not sure? The default answer is yes.
+
+Use methods on models to access data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- CRUD operations should access models via methods on models (where they exist), instead of querying managers directly.
```
albemarle: Should we define CRUD as well as CUD in the glossary?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+   -  Date & time of the change
+
+   -  The foreign key of of the user who initiated the change
+
+Use the correct data type for a column
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Don’t use a :code:`IntegerField` when a :code:`BooleanField` would do.
+
+-  Use :code:`BigIntegerField`_. for your foreign keys
```
albemarle: Is the period after `BigIntegerField`_ part of the syntax or a typo?

oeps/oep-38-Data-Modeling.rst
```diff
@@ -0,0 +1,401 @@
...
+
+Each column in a table should only store a single fact or dimension
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- If a column could be a mix of integer data and character data it is best to store these items as 2 different fields in the database
+
+Models should be Annotated for PII
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  All models in the OpenedX ecosystem should be tagged for PII using `code annotations`_. by following OEP-30_.
```
albemarle: Same question here, sorry if it's actually correct RST!

oeps/oep-38-Data-Modeling.rst
macdiesel: It's part of the syntax to make that code block also a clickable link.

oeps/oep-38-Data-Modeling.rst
macdiesel: I think it's correct RST?

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+   -  Date & time of the change
+
+   -  The foreign key of of the user who initiated the change
+
+Use the correct data type for a column
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Don’t use a :code:`IntegerField` when a :code:`BooleanField` would do.
+
+-  Use :code:`BigIntegerField`_. for your foreign keys
```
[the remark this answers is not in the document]
macdiesel: Done.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+
+Have sane default values
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- A model should have default values that make sense for the application
+
+- For example if you are adding a boolean to flag that a learner has not yet activated their account, the default value should be set to False, not None.
+
+The database layer should preserving uniqueness
```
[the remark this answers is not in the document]
macdiesel: Done

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,414 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  Don’t use a :code:`IntegerField` when a :code:`BooleanField` would do.
+
+-  Use :code:`BigIntegerField`_. for your foreign keys
+
+-  Don’t store an Integer field as :code:`CharField`_..
+.. _CharField: https://docs.djangoproject.com/en/2.2/ref/models/fields/#charfield
+
+- Store UUID's as :code:`CharField`_. with a max lenght that matches the max length of the UUID.
```
robrap: 1. lenght => length
2. I see now that some code uses UUIDField: https://docs.djangoproject.com/en/2.2/ref/models/fields/#uuidfield . Should that be avoided?

macdiesel: One of the issues with UUID Field is that when a monotonically incrementing integer is not included in the column loading data without duplicates becomes more difficult. It's possible that that UUID may shift of change. Including an integer based primary key de-risks a number of data duplication scenarios.

macdiesel: I've also updated the OEP to point to using the UUID field from Django. Thanks for the tip!

kdmccormick: It makes sense why the PK should be an auto-incrementing int, but why not use `UUIDField` non-PK UUIDs? It's used all over Discovery, Registrar, and Credentials. AFAIU, on MySQL it's just var(32) under the hood, with some convenient validation and parsing on the Django level.

kdmccormick: Sorry, was posting on old version of this thread. Cool!

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,418 @@
...
+.. _code annotations: https://github.com/edx/code-annotations>
+.. _OEP-30: https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst
+
+
+Have sane default values
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- A model should have default values that make sense for the application
+
+- :code:`CharField`_. Should be defined with the :code:`null=True`_. option.
```
kdmccormick: @macdiesel Django says the opposite, can you include the rationale here?

oeps/oep-0038-Data-Modeling.rst
ormsbee: So I agree with the decision that primary keys should be integers, but I'm curious about this statement:

> One of the issues with UUID Field is that when a monotonically incrementing integer is not included in the column loading data without duplicates becomes more difficult. It's possible that that UUID may shift of change.

My understanding is that UUIDs should never change, since one of their primary uses is to uniquely identify data across IDAs. If folks are ever mutating the UUID values on a row, I'd really like to understand why, because it sounds like something that will almost certainly bite us down the road. Or does this have more to do with the difficulty in doing range queries...?

Thank you.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -282,6 +282,7 @@ Have sane default values
 
 - :code:`CharField`_. Should be defined with the :code:`null=True`_. option.
 .. _null=True: https://docs.djangoproject.com/en/2.2/ref/models/fields/#null
+   - This is to improve post processing. Defaulting :code:`CharField`_. to null enables use to better derive the intent of the user. If the field is null no intent was made to enter that field. If the field is an empty string eitehr an empty string was entered of the field was modified later by the user.
```
kdmccormick: Makes sense.
`eitehr` -> `either`
`entered of` -> `entered or`

I assume this doesn't apply to required CharFields?

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
+      you need to run the python environment to decode the data, analyists who use SQL will have a difficult time querying and decoding this data.
+
+Store everything
+~~~~~~~~~~~~~~~~
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database or add history the answer is almost always yes. Store the data and add history. It can be removed later if it is found unnecessary.
+
+-  Still not sure? The default answer is yes.
```
nasthagiri: Once you choose to persist certain data, it can be used by others whether or not the data or its usage is intentionally thought through, which then makes it harder to remove later.  We find this issue at the application layer, with other APIs, for example.  Although theoretically, one would think we should default to adding APIs, without deliberate consideration of the APIs, we will (1) be cornered into supporting them (2) remove them and break our stakeholders, or (3) spend valiant efforts to migrate them.

Given that this OEP takes the standpoint of supporting stakeholders with direct access to our data models, our data models become an API that we must support.  Moving forward, it's best for our APIs to be as intentional as possible.  Alternatively, we can annotate/designate which data models are supported APIs and which aren't (for example, in experimental/prototype phase).

oeps/oep-0038-Data-Modeling.rst
macdiesel: Sometimes Dave, people do goofy things. That said I'm happy to strike the sentence.

oeps/oep-0038-Data-Modeling.rst
macdiesel: The reasoning on this is that there is a difference between an empty string value and a null value. When we set the null=True option we know that no one has attempted to fill out that field vs someone had filled out and blanked the field. It's helpful for analysis.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
+rich and efficient data models within the Open edX ecosystem. They are not
+standards but guidelines to help teams think about how to store data.
+
+**Data Dimension** - A Data Dimension is a set of data attributes pertaining
+to something of interest to a business. Dimensions are things like
+"customers", "products", "stores" and "time". For users of Data
+Warehouses, data dimensions are entry points to numeric facts (e.g.
+sale, profit, revenue) that a business wishes to monitor.
+
+**Data Modeling** - A process used to define and analyze data requirements
```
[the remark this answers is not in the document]
macdiesel: I think I had intended the process of defining data requirements to be the creation of the model.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
+
+Due to the potentially high cost of changing a model, it pays to get it
+as correct as possible the first time.
+
+Think about the person who will analyze this data later and the person
+who manages the system day to day. (That person could possibly be future
+you!) Try to do as much careful design up front to make your quality of
+life better later.
+
+History is best (but events are a close second)
```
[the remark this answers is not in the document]
macdiesel: This is something that is on my radar and will address at some point. I would prefer to do that as an update though.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
+-----------------------
+
+The standards below are designed to ensure edX can gain the highest value and insights from the data.
+The application of these standards is the most basic level of support to which all applications in the Open edX ecosystem
+should adhere. When creating new applications or models please ensure the models being created comform to the following.
+
+Use a numeric primary key
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  It is recommended to use use :code:`BigAutoField`_.
```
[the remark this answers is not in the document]
macdiesel: Brain stutter.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
+
+Responsibility for Third-party Service Integrations
+===================================================
+
+Not sure… Adhere to the same standards.
+
+Backward Compatibility
+======================
+
+Data models that are not within the standards of this document should be
```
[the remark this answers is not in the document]
macdiesel: I'm going to change this to no action required. I'm not sure we want to open this can of worms and we would probably be better served by specifically targeting models with issues that will have a material impact on how we are analyzing data or data quality issues that are preventing analysis.

oeps/oep-0038-Data-Modeling.rst
macdiesel: I don't feel that we should be overly restrictive about data or our data models in a broad sense. While the removal or fields has happened and caused issues in the past once or twice I don't want us to overly index on this. The removed fields were not business critical. 

That said I think that we could identify a subset of models that are perhaps more important and should have an API put in place. That said I'm unsure that should fall under the scope of this OEP. At minimum I would prefer to get this merged and deal with this for the first revision.

albemarle: As arbiter, I agree.  Let's get this version of this OEP out into the real world, and see what happens when people actually use it.  :-)

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,419 @@
...
++---------------+------------------------------------------------------------+
+| Resolution    | 'Original pull request'                                    |       
++---------------+------------------------------------------------------------+
+
+Abstract
+========
+
+In order to obtain the highest possible value from data collected in the
+Open edX ecosystem, this document attempts to provide guiding thoughts and
+principles on data modeling.
```
inventhouse: This OEP appears to be "hard-wrapped", it would be better to use logical breaks and allow auto-wrapping; see [OEP-19 Best Practices](https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#best-practices)

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+-  Data should be joined using primary keys wherever possible
+
+-  Foreign keys should use a naming convention of :code:`<object_name>_id` where object name is the name of the table of the foreign key relationship.
+
+-  Do not join on things such as username, email address, or other dimensions of data that may change over time
+
+-  Do not join on PII
+
+- Joining between IDAs should be done by using a universally unique identifier (UUID)
```
doctoryes: Is this line in conflict with the first line in this section? If not, perhaps it should be clarified when joining with primary keys is the best practice vs. joining with UUIDs?

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+- If a model needs to preserve uniqueness between many fields use :code:`unique_together`_.
+.. _unique_together: https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together
+
+
+
+Best Practices
+--------------
+These practices are designed to help teams create rich and efficient data models within the OpenedX ecosystem.
+They are not standards but guidelines to help teams think about how they are storing data.
```
doctoryes: Should there be a best practice added about *not* using SQL/Snowflake keywords as column names in this section?

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+- The name of a column in a table should accurately describe the data in that table.
+
+- If a column is named course_id it should only store the course_id. Not the course_key, not a numeric value, not a timestamp. Just the course_id. Conversely if a column is named course_run_key it should store the course run key, not the course_id.
+
+
+Deleting data
+~~~~~~~~~~~~~
+
+-  It is better to have a column to mark the record as inactive than to
+      remove the data from the system using the SQL delete keyword. These models should use Django's
```
doctoryes: Nit: Capitalize `DELETE`?

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+Responsibility for Third-party Service Integrations
+===================================================
+
+Adhere to the same standards.
+
+Backward Compatibility
+======================
+
+Data models that are not within the standards of this document do not need to be updated to adhere to OEP-38 standards.
```
doctoryes: Nit: Wording change?:
```
*Existing* data models which do not follow the data practices in the OEP
do not need to be updated to follow OEP-38 standards. *New* data models
*do* need to adhere to these standards.
```

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+edX uses data to decide which people should receive marketing emails,
+who passes or fails a course, or how much to pay our partners. Our
+partners use our data to target users and to
+refine and improve their courses. The data we collect today is being
+used to advance academic research about online learning and pedagogy.
+Decision making at edX should be data-driven and based on this collected data.
+
+This data is one of our most valuable assets and it
+should be a first-order concern. Save everything (disk is cheap). The more data we are able to collect about users,
```
fredsmith: Citation needed.  What is cheap? Edx.org has multiple terrabytes of history data that is running on expensive "hot" databases, never archived, snapshotted daily, and that adds a significant management overhead to every database maintenance operation. This costs us tens of thousands of dollars per month.  Do we ever look at the data? Anecdotally I can say we don't.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+behavior, and system state now, the more opportunities we’ll have to
+improve our decision-making in the future.
+
+Use industry best practices
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Data stored in the OpenEdX ecosystem should adhere to industry best
+practices. For example, since an industry best practice uses numerical
+identifiers to identify rows of data, the Open edX ecosystem should also use
```
fredsmith: Consider the limits of default numerical identifier types, the rate data is expected to be written to the table, and the expected lifespan of the table.   History table data schema changes are very, very expensive - hundreds of thousands of dollars in engineering time, compute time, and storage costs.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+Think about the person who will analyze this data later and the person
+who manages the system day to day. (That person could possibly be future
+you!) Try to do as much careful design up front to make your quality of
+life better later.
+
+History is best (but events are a close second)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Historical data accuracy is best when history about a change is captured at the database level. Changes in data relating to financial systems (e.g. enrollments, payments, course price changes) should be stored historically where the change is made, preferably in the same system as the system of record. This allows us to reconstruct the data at different points with much greater reliability.
```
fredsmith: having a lower cost history storage mechanisim, instead of storing in the high cost system of record, seems more prudent.  Put the history in a log storage format where analysis and export are fast but random seeks are hard, instead of in a relational database, where analysis and export are hard and random seeks are fast.

oeps/oep-0038-Data-Modeling.rst
fredsmith: UUIDs can span databases, foreign keys cannot.

oeps/oep-0038-Data-Modeling.rst
```diff
@@ -0,0 +1,411 @@
...
+
+-  Don’t store encoded (pickle, json, other) objects in the database! If
+      you need to run the python environment to decode the data, analysts who use SQL will have a difficult time querying and decoding this data.
+
+Store everything
+~~~~~~~~~~~~~~~~
+
+-  Storage is cheap!
+
+-  If you are unsure whether you should store something in the database or add history the answer is almost always yes. Store the data and add history. It can be removed later if it is found unnecessary.
```
fredsmith: Can we implement a metric for when data can be labeled unnecessary and deleted? Can we implement a process or policy for moving historical data between "hot" and "warm" and "cold" storage?  eg. RDS, S3, Glacier.

albemarle:
@macdiesel Can you ping me when you are done addressing @fredsmith 's suggestions?  Then I will give my arbitorial blessing for merging this OEP.

macdiesel:
@fredsmith, I do not have time to address your comments now. If you would like to open this OEP again and modify please have at it. I welcome your suggestions and changes, please tag me as a reviewer.

PR #120, opened 2019-07-15
Config OEP draft

coryleeio:


oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+
+This change is not backwards compatible, but it need not be implemented in each application immediately.
+It is meant to serve as a set of best practices for moving forward, but
+
+In general the migration path for old applications would look like this:
+
+	For all applications:
+    rename production.py to settings.py
+    rename base.py or common.py to defaults.py
+    make production.py source defaults.py and read yaml into python settings module
```
bderusha: Referencing production.py here might be confusing.  Maybe `make the new settings.py source defaults.py`

oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+defaults.py 
+
+defaults.py should list every configuration key in the application, and provide a description of what it does.
+
+settings.py should do two things:
+1) It should first load defaults.py into the python settings module.
+2) Then should load a single yaml file into the default keyspace overwriting values in the settings module.
+
+The yaml file path loaded should be configurable via environment variable that is specific to the application
+eg: ECOMMERCE_CFG.
```
bderusha: Where and how should people expect set this ENV variable?

oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+In general the migration path for old applications would look like this:
+
+	For all applications:
+    rename production.py to settings.py
+    rename base.py or common.py to defaults.py
+    make production.py source defaults.py and read yaml into python settings module 
+    (it should not specify every key like it does in many apps, it should load keys from the yaml into keys in the settings module without knowledge of what keys are possible.)
+    
+    Make yaml files for overrides in any other environments and remove the additional python environments.
+    Remove any --settings= from places where we interface with the application such as management commands, and from documentation for the application.
```
bderusha: How should devs expect to set the right ENV variable when switching between dev, test, etc.  Will we capture this in a `Makefile` and set the right env for you?  Will they need to set it on their own?

oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The yaml file path loaded should be configurable via environment variable that is specific to the application
+eg: ECOMMERCE_CFG. 
+
+There should be no additional django configuration settings files like devstack.py, testing.py, docker.py, docker_devstack.py, or anything of the sort.
+
+Plugins should not modify the loaded settings python module. Since we load the keys from the yaml file, even the ones we do not know at the application layer.
+A plugin should not need to modify the configuration of the application.
+
+All of the configuration should be done in the yaml file
```
cpennington: There are some types of django configuration that are not just lists and dicts (and yaml representable data). I think that that's probably fine, but it would be worth mentioning that anything that is not lists-and-dicts should be computed from the lists-and-dicts after it that is loaded from the yaml file.

oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+
+configuration should never be appended to or removed from in python code after settings.py is loaded, only read.
+
+eg:
```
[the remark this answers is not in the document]
coryleeio: Currently in the edxapp config we append to lists of keys in the config after it is loaded to avoid having to overwrite the key and specify the defaults again, this is done in a few places

here is an example of one:
https://github.com/edx/edx-platform/blob/master/lms/envs/production.py#L143

Basically it is an extension of config should be read and not modified.

oeps/oep-0039-yaml-config.rst
coryleeio: Yeah interpreting the config to create django objects to configure libraries and whatnot is totally reasonable. I will clarify.

oeps/oep-0039-yaml-config.rst
```diff
@@ -0,0 +1,146 @@
...
+eg: ECOMMERCE_CFG. 
+
+There should be no additional django configuration settings files like devstack.py, testing.py, docker.py, docker_devstack.py, or anything of the sort.
+
+Plugins should not modify the loaded settings python module. Since we load the keys from the yaml file, even the ones we do not know at the application layer.
+A plugin should not need to modify the configuration of the application.
+
+All of the configuration should be done in the yaml file
+
+configuration should never be appended to or removed from in python code after settings.py is loaded, only read.
```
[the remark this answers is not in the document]
coryleeio: This seems like a good idea. I will add it.

oeps/oep-0039-yaml-config.rst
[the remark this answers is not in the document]
coryleeio: The intent of it was just to clarify that the yml config key
MY_KEY

should be accessible as settings.MY_KEY in the python code.

I will clarify the example

oeps/oep-0039-yaml-config.rst
coryleeio: I will clarify - my intent was that there should only be one python django configuration environment and that environment should load a yml, changing configuration would always be done in the yml, if you had multiple environments to configure you would likely have multiple ymls.

coryleeio: There are 14 of these environments in the edx platform lms presently. (similar number in the cms)

oeps/oep-0039-yaml-config.rst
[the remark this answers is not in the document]
coryleeio: Here is another good example(that is currently biting me):
`/home/coryleeio/Workspace/edx-platform/lms/envs/bok_choy.py:
  254  # Make sure we test with the extended history table
  255  FEATURES['ENABLE_CSMH_EXTENDED'] = True
  256: INSTALLED_APPS.append('coursewarehistoryextended')
  257  
  258  BADGING_BACKEND = 'lms.djangoapps.badges.backends.tests.dummy_backend.DummyBackend'

/home/coryleeio/Workspace/edx-platform/lms/envs/production.py:
  929  # The extended StudentModule history table
  930  if FEATURES.get('ENABLE_CSMH_EXTENDED'):
  931:     INSTALLED_APPS.append('coursewarehistoryextended')
  932  
  933  API_ACCESS_MANAGER_EMAIL = ENV_TOKENS.get('API_ACCESS_MANAGER_EMAIL')

/home/coryleeio/Workspace/edx-platform/lms/envs/test.py:
  207  # Make sure we test with the extended history table
  208  FEATURES['ENABLE_CSMH_EXTENDED'] = True
  209: INSTALLED_APPS.append('coursewarehistoryextended')`

sarina:
Closing as last action was over 2 years ago. Please open a new PR if you would like this to be reviewed.

edx-webhook:
Although this pull request is already closed, I've created [OSPR-6419](https://openedx.atlassian.net/browse/OSPR-6419) so that we can track it in Jira.

natabene:
> Closing as last action was over 2 years ago. Please open a new PR if you would like this to be reviewed.

@sarina This actually was internal, as Cory worked for us at the time. Closing this made bot check again and now that the author is not in the list, it was flagged as an OSPR. @nedbat I feel like we might want to fix this at some point.

PR #121, opened 2019-07-18
WIP: OEP-40 Git Best Practices

jdmulloy:


BessieSteinberg:
Are there plans to create a standard on what tense commit messages are written in? For example, saying "Fix Bug" or "Fixed Bug"

jdmulloy:
> Are there plans to create a standard on what tense commit messages are written in? For example, saying "Fix Bug" or "Fixed Bug"

Good question. I don't have particularly strong feelings, but we should add a decision one way or the other. The [official git docs](https://git.kernel.org/pub/scm/git/git.git/plain/Documentation/SubmittingPatches) recommends imperative-mood "https://git.kernel.org/pub/scm/git/git.git/plain/Documentation/SubmittingPatches", so "Fix bug" instead of "Fixed bug". I think this is what I tend to do but it could be up for discussion.

BessieSteinberg:
@jdmulloy I similarly do not have strong feelings but think we should pick one.  I think the imperative (ex. "Fix bug") is a good choice if it's git's standard.

oeps/oep-00XX-git-best-practices.rst
```diff
@@ -0,0 +1,107 @@
...
+---------------
+* The first line should be no longer than 50 characters
+* The second line should be no longer than 72 characters
+* The first line should concisely describe the reason for the change
+* The first line should include the JIRA Issue ID, if there is a JIRA Issue
+Branch names
+------------
+* Feature branch names should be prefixed with the creating user's github username and a slash
+* If there is an associated JIRA Issue the Issue ID should be next before another slash
+* Example: username/proj-1234/descriptive name
```
ormsbee: Does this really matter? Don't we immediately delete the branches after merging anyhow?

jdmulloy: Fair point. It's something I do and I find it helpful. It is a bit redundant once there's a commit on the branch with the JIRA ID in the message. This could be a suggestion vs a requirement. Do OEPs have the shall and may language that RFCs have?

bradenmacdonald: By default, GitHub puts the branch name as the first line of any merge commit ([example](https://github.com/edx/edx-platform/commit/af7ee42bdefc9646b19ed9df0ad52ed29503f085)), so a clear branch name makes it easier to scan the list of merge commits to see what each merge was. I wish they would change to put the PR title as the first line, which is much more useful.

I personally like to override the GitHub default and change the merge commit first line to
"Merge PR #NNNNN: Title of the PR"
vs.
"Merge pull request #NNNNN from username/partial-branch-nam..."

ormsbee:
For me, I think the most important part of the git message is explaining the "why", and to be able to do so without relying on the ticket for context. Including a reference to the ticket is great. But some tickets are not public for privacy/security reasons. Some tickets are not public because we still have trouble with all the permissions of JIRA. Some tickets don't work any more because they were part of our last JIRA system. Some tickets don't exist because they were from when we were using LightHouse. Who knows what we'll be forced to switch to in three years when Capital One buys Atlassian?

oeps/oep-00XX-git-best-practices.rst
```diff
@@ -0,0 +1,107 @@
...
+The specification describes the technical details of the Architecture, Best
+Practice or Process proposed by the OEP. If the proposal includes a new API,
+specify its syntax and semantics.
+
+Commit messages
+---------------
+* The first line should be no longer than 50 characters
+* The second line should be no longer than 72 characters
+* The first line should concisely describe the reason for the change
+* The first line should include the JIRA Issue ID, if there is a JIRA Issue
```
ormsbee: For me, I think the most important part of the git message is explaining the "why", and to be able to do so without relying on the ticket for context. Including a reference to the ticket is great. But some tickets are not public for privacy/security reasons. Some tickets are not public because we still have trouble with all the permissions of JIRA. Some tickets don't work any more because they were part of our last JIRA system. Some tickets don't exist because they were from when we were using LightHouse. Who knows what we'll be forced to switch to in three years when Capital One buys Atlassian?

Making guidelines about past vs. present tense or character width guidelines are good things. But I think we should emphasize the content of the message itself and give some examples. I'd much rather be looking at a badly formatted but informative commit message than a one liner reference to a ticket.

oeps/oep-00XX-git-best-practices.rst
albemarle: I think it's a useful convention, but doesn't belong in an OEP.

oeps/oep-00XX-git-best-practices.rst
```diff
@@ -0,0 +1,107 @@
+======================
+OEP-XXXX: OEP Template
+======================
+
+.. This OEP template is based on Python's PEP standard.
+
++-----------------+--------------------------------------------------------+
+| OEP             | :doc:`OEP-XXXX </oeps/oep-XXXX-bp-git-best-practixes`  |
++-----------------+--------------------------------------------------------+
+| Title           | Git best practices                                     |
```
albemarle: Maybe "Git Commit Messages" ?

oeps/oep-00XX-git-best-practices.rst
albemarle: Jira is not an acronym, just a regular name.

sarina:
This is about Git commit specification, and many of these recommendations have already or could be put into OEP-51 Conventional Commits https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html

Further, Jira number in the title is actively discouraged. This applies to edx.org team only, and often the ticket is inaccessible by the community. Many edx.org devs use Jira numbers as a way to avoid writing good commit messages or even PR descriptions.

If edx.org would like their commits to have the jira number, that can go in the commit body. Keep in mind that community members from other organizations may do the same thing with their ticket trackers.

PR #127, opened 2019-11-18
OEP-2: Updates to Ownership Assignment Metadata

nasthagiri:
This PR updates **OEP-2 Repository Metadata** to propose new practices for ownership bookkeeping and designation in our edX repositories.

#### Review Period
Nov 25, 2019 -> Dec 9, 2019

#### Arbiter
@cpennington 

#### Changes
At a high level, this PR proposes the following:

1. ownership is assigned to squads (i.e., teams with 5-6 people each) as opposed to individuals
2. owners are assigned at the granularity of higher-level, business-aligned functionality (user-facing services, user-facing apps, and system-wide frameworks)
3. lower-level repos are indirectly assigned owners through assignment to dependent higher-level repos

#### Motivation
See the rationale listed in the PR.  Essentially:

1. We'd like for healthy and longer-lasting teams to be steady and responsible technical stewards of the things they own.
2. We'd like to have more meaningful trade-off conversations with Product Owners by having ownership assigned and discussed at a higher level that directly aligns with business functionality.
3. We'd like to minimize the amount of maintenance of ownership assignments on lower-level repos.

#### Adoption of this PR
As posted in [this comment](https://github.com/edx/open-edx-proposals/pull/127/files#r349311753), the success of these changes rely on rapid adoption by product delivery/engineering teams and comprehensive update of all openedx.yaml files throughout the codebase. Additional tooling will also be necessary to easily consume and make these changes actionable and referenceable.

### DIFF
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -10,7 +10,7 @@ OEP-2: Repository Metadata
    * - Title
      - Repository Metadata
    * - Last Modified
-     - 2019-11-14
+     - 2019-12-11
    * - Authors
      - Calen Pennington, Feanil Patel, Nimisha Asthagiri
    * - Arbiter
@@ -37,19 +37,30 @@ Bear in mind that particular classes of Open edX repositories may have different
 Specification
 =============
 
+Keys
+----
+
 Each repo will include a file ``openedx.yaml``, with the following keys:
 
-``owner``: string (required unless ``archived`` is ``True``)
-    The GitHub user responsible for the repository.
+``owner``: dictionary (required)
+    This key contains information about the assigned owner of this repository.
 
-``supporting_teams``: list of strings (optional)
-    The GitHub teams supporting the owner for this repository.
+    ``type``: string (required)
+        The value of this key must be either ``team`` or ``repo``. It indicates which ownership model applies to this repository, and therefore which of these two keys should exist with a non-empty value.
+
+        **Note:** As much as possible, repos should be owned by higher-level repos, with only the highest-level repos owned by teams.
+
+    ``team``: string (optional)
+        The GitHub team responsible for this repository, prefixed with the Github organization (e.g., ``edx/arch-bom``). If this value is set, this repository contains high-level, business-aligned functionality that is suitable for direct ownership designation to an individual responsible team.
+
+    ``repo``: string (optional)
+        The GitHub repo responsible for this repository, prefixed with the Github organization (e.g., ``edx/XBlock``). If this value is set, this repository contains lower-level technical functionality that is affiliated with another primary owning higher-level repository.
 
 ``nick``: string (optional)
-    A short-name for the repository used by reporting tools created by the edX Open Source team. Should be 3-4 characters, because this value is used in charts with very limited space on their axes.
+    A short-name for this repository. This is used by reporting tools created by the edX Open Source team. Should be 3-4 characters, because this value is used in charts with very limited space on their axes.
 
 ``tags``: list of strings (optional)
-    Used to group repositories by in reporting tools. Sample tags: core, mobile analytics.
+    This key is used to annotate repositories, for example to categorize them for automated reporting tools. See Tags_.
 
 ``oeps``: dictionary (optional)
     A list of keys corresponding to Best Practice OEPs (in the format ``oep-n``). Each value would be one of ``True``, ``False``, or a reason dictionary. ``True`` or ``False`` indicate compliance (or lack thereof) with the specified OEP.
@@ -60,14 +71,20 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
     -  If the ``state`` is ``False`` a ``reason`` value must be included that explains why.
     -  If the ``state`` is ``True`` or ``applicable`` is ``False`` the ``reason`` can optionally provide more information. The ``reason`` value will be displayed in reporting tools. If an OEP isn't listed in the ``oeps`` dictionary, then it is assumed to be ``False``, unless the reporting tool can auto-detect accordance.
 
-``track-pulls``: obsolete
-    This key is obsolete, please remove if found.
-
-``openedx-release``: defined by :doc:`OEP-10 <oep-0010-proc-openedx-releases>`.
+``openedx-release``: defined by :doc:`OEP-10 <oep-0010-proc-openedx-releases>` (optional)
     Define this key if your repo is an application or IDA that is part of Open edX releases.  Omit this key if your repo is a library, or is not part of Open edX releases.  See :doc:`OEP-10 <oep-0010-proc-openedx-releases>` for details.
 
 ``archived``: boolean (optional)
     If ``True``, this specifies that this repository is archived and no longer maintained by edX.
+
+Obsolete Keys
+*************
+
+``track-pulls``:
+    This key is obsolete, please remove if found.
+
+``supporting_teams``: list of strings
+    This is an obsolete key that was used to specify Github handles for teams supporting the owner of the repository. This is now deprecated in favor of the updated ``owner`` dictionary. As a reasonable migration of this key, set ``owner.type`` to *team*, ``owner.value`` to the first value in ``supporting_teams``, and then remove ``supporting_teams``.
 
 Tags
 ----
@@ -89,7 +106,7 @@ The ``tags`` structure is very open ended but there are a few tags that are curr
     eg. frontend-app-profile, edx-app-ios
 
 ``framework``:
-    The repository contains shared functionality that form a common framework used by multiple services or apps in the platform.
+    The repository contains shared functionality that forms part of a common framework used by multiple services or apps in the platform.
 
     Code runs in production, behind the scenes, (intended to be) globally used throughout the platform.
 
@@ -123,6 +140,18 @@ The ``tags`` structure is very open ended but there are a few tags that are curr
 
     eg. xblock-review, staff_graded-xblock
 
+Ownership
+---------
+While the responsibilities of an owner are outside the scope of this OEP, a repository's metadata should provide sufficient information to identify and discover the repository's owner.
+
+Granularity of Code Ownership
+*****************************
+Rather than designating human owners to each and every repository, we designate owners to only high-level and business-aligned repositories as described in ``owner.team`` in Keys_. All other repositories are indirectly owned by owners of the high-level repositories. This indirect relationship is described in the ``owner.repo`` designation in Keys_. See `Ownership Rationale`_.
+
+Granularity of Human Owners
+***************************
+We choose to assign singular "healthy" teams as owners, rather than individuals or other organizational structures (such as working groups or edX "themes"). We consider a team size of 5-6 people as "healthy". See `Ownership Rationale`_.
+
 Example
 -------
 
@@ -135,12 +164,9 @@ For example, in the `edx-platform`_ repo, the file might look like:
     # openedx.yaml
 
     ---
-    owner: nasthagiri
-    supporting_teams:
-        - platform-core
-        - platform-core-extensions
-        - platform-authn
-        - platform-courseware
+    owner:
+        type: team
+        team: edx/arch-bom
     nick: edx
     tags:
         - core
@@ -164,7 +190,13 @@ For example, in the `edx-platform`_ repo, the file might look like:
 Rationale
 =========
 
+Keys Rationale
+--------------
+
 The keys in ``openedx.yaml`` were derived from existing repository metadata collected by edx.org.
+
+OEPs Rationale
+--------------
 
 The design of the ``oeps`` dictionary was guided by a couple of use cases:
 
@@ -172,9 +204,30 @@ The design of the ``oeps`` dictionary was guided by a couple of use cases:
 2. Repositories may have specific requirements that force them to not implement a best practice. The tools should be able to present that reasoning to anyone looking across repositories, and the reasons should be documented in the repositories themselves.
 3. As much as possible, Best Practices should be autodetected, but because they will often involve a judgement call, autodetection shouldn't be mandatory.
 
+Ownership Rationale
+-------------------
+
+Keeping the granularity of **code ownership to high-level repositories** affords us:
+
+#. Stronger collaboration on ownership with business functions since a high-level repo is directly aligned to a business need.
+#. Minimization of bookkeeping maintenance as code and owners evolve. However, we will need to ensure that all lower-level repos are still owned indirectly via high-level repos.
+
+Keeping the granularity of **human owners to individual teams** affords us:
+
+#. Using existing organizational structures that have scrum processes in-place and cross-functional communication channels to follow through on their ownership responsibilities.
+#. Robust and longer-lasting owner relationships that overcome the absences and attrition of individuals.
 
 Change History
 ==============
+
+2019-12-11
+----------
+
+* New practices for ownership bookkeeping and designation in our repositories:
+
+  * Ownership is assigned to squads (i.e., teams with 5-6 people each) as opposed to individuals.
+  * Owners are assigned at the granularity of higher-level, business-aligned functionality (user-facing services, user-facing apps, and system-wide frameworks).
+  * Lower-level repos are indirectly assigned owners through assignment to dependent higher-level repos.
 
 2019-10-29
 ----------
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -75,19 +75,53 @@ Tags
 The ``tags`` structure is very open ended but there are a few tags that are currently in wide use and are provided here so that they can be used consistently.
 
 ``webservice``:
-    Indicate that this repository is the root of a webservice.  eg. It can serve HTTP content.
+    The repository is the root of a webservice that serves HTTP content.
 
-``library``:
-    The repository is an installable package that is reusable but doesn't run standalone. eg. a pip installable pyton package
+    Code runs in production web servers and has publicly accessible views/APIs for end-users.
+
+    eg. credentials
+
+``frontend-app``:
+    The repository primarily contains code for a frontend application.
+
+    Code runs in production devices (such as web browsers and mobile devices) and is publicly accessible by end-users.
+
+    eg. frontend-app-profile, edx-app-ios
+
+``framework``:
+    The repository contains shared functionality that form a common framework used by multiple services or apps in the platform.
+
+    Code runs in production, behind the scenes, (intended to be) globally used throughout the platform.
+
+    eg. edx-drf-extensions, edx-ace, frontend-platform, paragon
 
 ``backend-service``:
-    A service that is run as a part of openedx but is not really interacted with by the end-users.  eg. xqwatcher or ecommerce-worker
+    A service that is run as part of the platform but is not directly accessed by end-users.
+
+    Code runs in production backend servers, typically within the firewall, and provides APIs for access by public-facing ``webservices``.
+
+    eg. xqueue-watcher, ecommerce-worker
 
 ``backend-tooling``:
-    Scripts or configuration used in conjunction with services or in support of openedx.
+    The repository contains scripts or configuration used in conjunction with backend services or in support of openedx.
+
+    Code does *not* run in production.
+
+    eg. ecommerce-scripts, repo-tools, testeng-ci
+
+``library``:
+    The repository is an installable package that is reusable but doesn't run standalone. eg. a pip installable python package
+
+    Code runs in production, behind the scenes, locally used by a part of the platform.
+
+    eg. frontend-component-footer, edx-milestones, user-util
 
 ``xblock``:
-    Indicating that a repo contains an xblock or xblock related tooling.
+    The repository contains an xblock or xblock related tooling.
+
+    Code runs in LMS and Studio as part of course run content.
+
+    eg. xblock-review, staff_graded-xblock
 
 Example
 -------


oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -136,74 +178,29 @@ For example, in the `edx-platform`_ repo, the file might look like:
...
+#. Stronger collaboration on ownership with business functions since a high-level repo is directly aligned to a business need.
+#. Minimization of bookkeeping maintenance as code and owners evolve. However, we will need to ensure that all lower-level repos are still owned indirectly via high-level repos.
 
-2016-06-29
-----------
+Keeping the granularity of **human owners to individual teams** affords us:
 
-* Original publication
+#. Using existing organizational structures that have scrum processes in-place and cross-functional communication channels to follow through on their ownership responsibilities.
+#. Robust and longer-lasting owner relationships that overcome the absences and attrition of individuals.
```
kdmccormick: Does it actually guarantee long-lasting owner relationships, though? Until the Py3 upgrade, which required combing through openedx.yaml files, I'd regularly see things like `edx/tnl` or `edx/educator` as the owner of a repository. That was earlier this year.

I don't have a preference of team ownership vs. individual ownership. However, I do think we need one of the following, else the ownership definitions will continue to fall out-of-date and only be semi-useful.
1. Automated tooling that ensures that the owning individual is still part of the edX org, or
2. a. Automated tooling that ensures that the owning team exists, *and*
    b. Having someone who is responsible for managing & pruning edX GitHub teams, *and*
    c. Having a reliable way to get in contact with a team based on their GitHub team name (Slack channels often don't line up with teams)

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -29,74 +30,119 @@ Proposes a standard format for repository metadata for Open edX projects.
...
 =============
 
+Keys
+----
+
 Each repo will include a file ``openedx.yaml``, with the following keys:
 
-``owner``: string (required unless ``archived`` is ``True``)
-    The GitHub user responsible for the repository.
+``owner_type``: string (required)
```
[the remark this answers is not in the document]
feanil: Agree, I think we should just make it clear that one of those two fields is required.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -29,74 +30,119 @@ Proposes a standard format for repository metadata for Open edX projects.
...
+    The value of this key must be either ``owning_team`` or ``owning_repo``. It indicates which ownership model applies to this repository, and therefore which of these two keys should exist with a non-empty value.
+
+``owning_team``: string (optional)
+    The GitHub team responsible for this repository. If this value is set, this repository must have one of the following Tags_: ``webservice``, ``frontend-app``, or ``framework``. That is, this repository contains high-level, business-aligned functionality that is suitable for direct ownership designation to an individual responsible team.
 
-``supporting_teams``: list of strings (optional)
```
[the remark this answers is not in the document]
feanil: +1 to keeping the obselete tag on this key with instructions.  Similar to the `track-pulls` key.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -29,74 +30,119 @@ Proposes a standard format for repository metadata for Open edX projects.
...
+    A service that is run as part of the platform but is not directly accessed by end-users.
+
+    Code runs in production backend servers, typically within the firewall, and provides APIs for access by public-facing ``webservices``.
+
+    eg. xqueue-watcher, ecommerce-worker
+
+``backend-tooling``:
+    The repository contains scripts or configuration used in conjunction with backend services or in support of openedx.
+
+    Code does *not* run in production.
```
feanil: The `configuration` and `terraform` reepo feel like they would get this tag if it were not for this line.  Both of those repos have code that "runs in" production in that they build/maintain production.

Is there a different tag you would give those tools or should we just remove this line from the description?

oeps/oep-0002-bp-repo-metadata.rst
feanil: To me, this update to the docs is the first step in making ownership match what we want instead of what we need to settle for.  

I think it's success will be indicated by how quickly it is applied to our code as well as how quickly we start referring to it for planning and coordination.  For this, some tools will definitely need to be built so that we can easily consume all this data.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: Interesting... I tend to appreciate explicitness over implicitness.

Advantages of being explicit of `owner_type`:
1. **Writers** Having the writer/maintainer of the `openedx.yaml` file be explicit of the ownership model for that repo forces them to think about its implications.

2. **Readers** Downstream tools are simplified since they don't have to have `if-then` statements and understand precedence order in case a repo had multiple `owning` fields entered.

3. **Updates** In case we ever added a new ownership model or allowed for fallbacks (both `owning_team` and `owning_repo` are specified but `owning_team` takes primary precedence, for example), we can do so more easily. All Readers will know that a new ownership model has been introduced since there would be a new value for `owner_type`.

Disadvantages of being explicit of `owner_type`:
1. One additional field to enter per repo.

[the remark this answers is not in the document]
feanil: Maybe owner should just be a dict `type` and `value` attributes that are both required?

This way we can be explicit without repetition and there is one field that holds the answer for who is the owner.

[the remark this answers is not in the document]
nasthagiri: Sure. We can do that.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: Good point.  I'll create an `Obsolete Keys` section.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: In our draft attempt of `Ownership Assignment`, we listed `terraform` as a `framework`. 

I see potential value in us knowing which of our repos are never run in production. It may help with resourcing and prioritization efforts for upgrades/etc.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: Suggestion on the name of the `owner` dict?  It may be best if it's a different name from the existing `owner` name, right?  So we can automate finding which repos still need to be updated.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: PR is updated with an `Obsolete Keys` section.  Please review.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: PR is updated with `owner` as a dict.  Please review.

oeps/oep-0002-bp-repo-metadata.rst
feanil: Ah, cool, that sounds good.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -60,34 +69,86 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
...
     If ``True``, this specifies that this repository is archived and no longer maintained by edX.
 
+Obsolete Keys
+*************
+
+``track-pulls``:
+    This key is obsolete, please remove if found.
+
+``supporting_teams``: list of strings
+    This is an obsolete key that was used to specify Github handles for teams supporting the owner of the repository. This is now deprecated in favor of the updated ``owner`` dictionary. As a reasonable migration of this key, set ``owner.type`` to *team* and ``owner.value`` to the first value in ``supporting_teams``.
```
feanil: NIT: Can we also add `Then remove this key.` to the end of this paragraph.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -37,19 +37,28 @@ Bear in mind that particular classes of Open edX repositories may have different
...
 =============
 
+Keys
+----
+
 Each repo will include a file ``openedx.yaml``, with the following keys:
 
-``owner``: string (required unless ``archived`` is ``True``)
-    The GitHub user responsible for the repository.
+``owner``: dictionary (required)
```
kdmccormick: Nit: do we care if the teams/repos are prefixed with `edx/` or not? If the answer is "no we don't care" and we're fine with tools handling both, that's fine with me.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: Done.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: Good point.  I've added prefixing with the Github organization.

nedbat:
I'm curious about the branch name here: "arch/ownership-update".  Usually branches have a prefix of the GitHub user doing the work.  Is this another sign of ownership moving from people to teams?

nedbat:
Hmm, I guess not: Nimisha has often made pull requests here from branches with an "arch/" prefix.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -130,44 +188,29 @@ For example, in the `edx-platform`_ repo, the file might look like:
...
 
 1. Adding a new OEP that few repositories will support, initially. We shouldn't need to update all repositories to detect if they are in accordance, we should be able to assume that they aren't, or automatically detect whether they are.
 2. Repositories may have specific requirements that force them to not implement a best practice. The tools should be able to present that reasoning to anyone looking across repositories, and the reasons should be documented in the repositories themselves.
 3. As much as possible, Best Practices should be autodetected, but because they will often involve a judgement call, autodetection shouldn't be mandatory.
 
+Ownership Rationale
+-------------------
 
-Change History
-==============
```
nedbat: Why are we deleting the change history?

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nedbat: Seems like we still have "type" which must be "team" or "repo", and then there is a "team" key or a "repo" key.  This redundancy just seems like more opportunity for error without adding any more information.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -37,19 +37,28 @@ Bear in mind that particular classes of Open edX repositories may have different
...
+    This key contains information about the assigned owner of this repository.
+
+    ``type``: string (required)
+        The value of this key must be either ``team`` or ``repo``. It indicates which ownership model applies to this repository, and therefore which of these two keys should exist with a non-empty value.
 
-``supporting_teams``: list of strings (optional)
-    The GitHub teams supporting the owner for this repository.
+    ``team``: string (optional)
+        The GitHub team responsible for this repository, prefixed with the Github organization (e.g., ``edx/arch-bom``). If this value is set, this repository must have one of the following Tags_: ``webservice``, ``frontend-app``, or ``framework``. That is, this repository contains high-level, business-aligned functionality that is suitable for direct ownership designation to an individual responsible team.
+
```
nedbat: It's not obvious to me that a "team" owner can only apply to webservice, frontend-app, or framework.  I don't think "must" for the tags is helpful here.  Wouldn't repo-tools be owned by the Open edX team? It's not one of those three things.

nedbat:
BTW, as a stylistic thing, it would be good to capitalize appropriately (Python, Open edX, XBlock, GitHub).  I'm also a fan of using English words instead of Latin abbreviations ("eg" -> "for example").

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
kdmccormick: I prefer `type` and `value`, which removes the redundancy. Unfortunately YAML doesn't have sum types :P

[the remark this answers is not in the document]
nedbat: I'm not sure what you mean by "doesn't have."  Wouldn't this be sufficient:
```
owner:
  repo: edx/edx-platform
```
or:
```
owner:
  team: edx/arch-bom
```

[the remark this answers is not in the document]
kdmccormick: It's reasonable, but doesn't preclude:
```yaml
owner:
  repo: edx/edx-platform
  team: edx/arch-bom
```
A more "sum type" -like constructor would look something like:
```yaml
owner: repo(edx/edx-platform)
```
Of course, YAML isn't typed anyway, so anyone can write anything they want. So we just want something that is maximally readable and minimally ambiguous.

I'm fine with both your suggestion and @feanil 's suggestion, which was:
```yaml
owner:
  type: repo
  value: edx/edx-platform
```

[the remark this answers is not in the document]
nedbat: The spec could preclude both repo and team by saying, "Don't provide both repo and team."  Ultimately it's the tooling that's going to enforce any of these rules.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -60,34 +69,86 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
...
+``frontend-app``:
+    The repository primarily contains code for a frontend application.
+
+    Code runs in production devices (such as web browsers and mobile devices) and is publicly accessible by end-users.
+
+    eg. frontend-app-profile, edx-app-ios
+
+``framework``:
+    The repository contains shared functionality that form a common framework used by multiple services or apps in the platform.
```
albemarle: form -> forms (but maybe you really want "forms part of"?)

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: Our newer OEPs and OEP templates no longer require a `Change History` within the OEP.  The expectation is that folks can look at the OEP's Git commit history to catch up on its historical changes.

Having said that, if we believe this is a controversial change, I can revert this deletion and do so in its own PR.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: The proposal is that `repo-tools` would have an `owner.type` of `repo` and that `repo` would be a higher-order webservice, frontend-app, or framework.  In this case, it can be the `openedxstats` webservice.

Do you have an alternative proposal that works better, while serving our goal of finding a more maintainable ownership assignment model?

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -37,19 +37,28 @@ Bear in mind that particular classes of Open edX repositories may have different
...
+    ``repo``: string (optional)
+        The GitHub repo responsible for this repository, prefixed with the Github organization (e.g., ``edx/XBlock``). If this value is set, this repository must *not* have one of the following Tags_: ``webservice``, ``frontend-app``, or ``framework``. That is, this repository contains lower-level technical functionality that is affiliated with another primary owning service/app/framework.
 
 ``nick``: string (optional)
-    A short-name for the repository used by reporting tools created by the edX Open Source team. Should be 3-4 characters, because this value is used in charts with very limited space on their axes.
+    A short-name for this repository. This is used by reporting tools created by the edX Open Source team. Should be 3-4 characters, because this value is used in charts with very limited space on their axes.
 
 ``tags``: list of strings (optional)
-    Used to group repositories by in reporting tools. Sample tags: core, mobile analytics.
+    This key is used to group repositories by reporting tools. Sample tags: core, mobile analytics.
```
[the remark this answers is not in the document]
nasthagiri: Good point. I updated it to refer to the `Tags` section.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: Updated to `forms part of`. Thanks!

oeps/oep-0002-bp-repo-metadata.rst
nedbat: repo-tools gets used by a few other repos.  repo-tools-data even more so.  A library repo is likely to be used by more than one other repo, isn't that one of the reasons it's been made into a library?

If a team is willing to say, "We own this library," why should this OEP forbid it?  Perhaps most libraries do fit into the "owned by a repo" model, great.

oeps/oep-0002-bp-repo-metadata.rst
nedbat: I think one of the over-arching bits of feedback we get from the community is, "it's hard to know what you've done."  OEPs change rarely. It won't be a burden to maintain a change history.

oeps/oep-0002-bp-repo-metadata.rst
cpennington: As another example: `edx_lint` is a library (and commandline tool) used by a number of other repos. However, it is none of `webservice`, `frontend-app`, or `framework`, and also isn't appropriate to be owned by another repo.

nedbat: Just one more example: edx-drf-extensions is used by a few different IDAs.

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -60,34 +69,86 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
...
+---------
+While the responsibilities of an owner are outside the scope of this OEP, a repository's metadata should provide sufficient information to identify and discover the repository's owner.
+
+Granularity of Code Ownership
+*****************************
+Rather than designating human owners to each and every repository, we designate owners to only high-level and business-aligned repositories as described in ``owner.team`` in Keys_. All other repositories are indirectly owned by owners of the high-level repositories. This indirect relationship is described in the ``owner.repo`` designation in Keys_. See `Ownership Rationale`_.
+
+Granularity of Human Owners
+***************************
+We choose to assign singular teams (i.e., edX "squads") as owners, rather than individuals or other organizational structures (such as working groups or edX "themes"). See `Ownership Rationale`_.
```
iloveagent57: Just a clarifying question: for many edX themes, a squad is the second-lowest level in the hierarchy, with a "team" or "pod" being the lowest level (for example, I am on the Programs theme, in the Masters/MM squad, on the "Grades and assignments" team/pod.  Do you mean "squad" in that same organizational sense here?

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: The intention here is to _not_ assign teams/owners to low-level technical repositories.  That results in harder to maintain assignments as the platform and teams evolve and also makes it harder to keep track of these in our conversations with Product.  Rather, we'd like to have the mapping between "humans" and "technology" at higher-level concepts.

To support this, we will need to make technical decisions on (1) the single-responsibilities of repos and (2) map lower-level repos with higher-order technical concerns.  Here are some examples:
* `edx-drf-extensions` can be listed as a `framework` related to APIs and renamed to something like `edx-rest-api-providers`. (It is listed below as an example `framework`.)
* `auth-backends` can be listed as a `library` related to user-authentication and owned by the `user_authn` app owner within edx-platform.
* `edx_lint` can be its own `framework` related to linting or a `library` mapped to a higher-order framework - depending on what we find in our exercise to assign all the repos.

The _proof is in the pudding_.  As we go through the effort of comprehensively assigning repos, we'll see how the current proposal for tags plays out.  If needed, we can add/edit.  The primary motive here is to assign ownership to humans at higher-order concepts.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: @iloveagent57  Thanks for asking. Our goal is to assign ownership to relatively "healthy and stable" teams so they can absorb the cost of ownership (via rotating pager duties and upkeep) and become long-term technical stewards/architects of what they own - both the thinkers and the doers of what they own.  We consider a team size of 5-6 people as "healthy" and so whatever that translates to within your theme.  I can update this paragraph to clarify that.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: @nedbat I'll go ahead and re-insert the change history in this PR.  We can determine outside of this OEP whether there is still value in keeping this history within the content of OEPs.

oeps/oep-0002-bp-repo-metadata.rst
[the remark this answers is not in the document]
nasthagiri: @kdmccormick and @nedbat Can you explain how the other 2 suggestions could support the 3rd goal I listed above?

```
3. Updates. In case we ever added a new ownership model or allowed for fallbacks (both owning_team and owning_repo are specified but owning_team takes primary precedence, for example), we can do so more easily. 
```

With the `type` and `value` option, wouldn't value then have to be a `dict`?

oeps/oep-0002-bp-repo-metadata.rst
nedbat: I understand the goal of trying to assign ownership in ways that ease maintenance.  I don't think calling edx-lint a framework is a good way to accomplish that. It's not a framework. It's a tool and a library. Why can't a library be owned by a team? Most libraries won't, but some will.

Why not simply say, "as much as possible, repos should be owned by other repos, and that the uppermost repos should be owned by teams"? That's easily understood. You can provide guidance that many libraries naturally fit into ownership by another repo.  It feels arbitrary to say that "frameworks" are top-level and "libraries" are not. It feels contorted to call edx-lint a framework just so that we can allow it to be owned by a team.

nasthagiri: That's fair.  I'll update the OEP to not be restrictive of which `tags` imply team-ownership. This decouples ownership from repo types.

Do note though, `edx-lint` doesn't abide by the current definition of a `library`:  `Code runs in production..`

Once again, I do expect us to update this OEP again once we comprehensively update all the repos and see what additional tags are missing and needed.

nasthagiri:
Thank you all for your thoughtful comments and collective desire to improve ownership of our platform.  I have responded to all the comments and received official approval from this OEP's Arbiter, @cpennington.  I will now squash the commits and merge.

oeps/oep-0002-bp-repo-metadata.rst
kdmccormick: A `tools` tag might be warranted.

PR #128, opened 2019-11-22
OEP-2: Formatting Updates

nasthagiri:
This PR simply reformats OEP-2 according to the latest standards.  There are no content changes included.

### DIFF
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -2,24 +2,25 @@
 OEP-2: Repository Metadata
 ==========================
 
-+---------------+-------------------------------------------+
-| OEP           | :doc:`OEP-2 <oep-0002-bp-repo-metadata>`  |
-+---------------+-------------------------------------------+
-| Title         | Repository Metadata                       |
-+---------------+-------------------------------------------+
-| Last-Modified | 2019-10-29                                |
-+---------------+-------------------------------------------+
-| Author        | Calen Pennington <cale@edx.org>           |
-|               | Feanil Patel <feanil@edx.org>             |
-+---------------+-------------------------------------------+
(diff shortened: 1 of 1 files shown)


oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -2,24 +2,25 @@
...
-+---------------+-------------------------------------------+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :doc:`OEP-2 <oep-0002-bp-repo-metadata>`
+   * - Title
+     - Repository Metadata
+   * - Last Modified
+     - 2019-11-14
```
jmbowman: Minor point - should this be updated to 11/22 now?

oeps/oep-0002-bp-repo-metadata.rst
```diff
@@ -2,24 +2,25 @@
...
+   * - Authors
+     - Calen Pennington, Feanil Patel, Nimisha Asthagiri
```
jmbowman: Are we deliberately shifting away from including email addresses?  OEP-1 just mentions them as optional, but still has it as required for the arbiter.

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: I can go either way on this.  Do reformatting changes require updating the timestamp?  I'm also not sending out reformatting changes for a wider review.  Git already catches these iterative timestamps, right?

oeps/oep-0002-bp-repo-metadata.rst
nasthagiri: Good catch.  I'll add the email address for the arbiter.

I was influenced by the fact that later OEPs didn't include email addresses.  (Side note: going through and updating the formatting of all OEPs may be a one-time worthwhile investment.)

PR #129, opened 2019-12-06
Tweak the OEP templates

nedbat:
- No reason to italicize some of the table sections

- Use the "Open edX" trademark correctly

- Link to the latest Discourse instance

### DIFF
--- a/oeps/oep-templates/adr-based-template.rst
+++ b/oeps/oep-templates/adr-based-template.rst
@@ -27,7 +27,7 @@ OEP-XXXX: OEP Template
 +-----------------+--------------------------------------------------------+
 | Created         | <date created on, in YYYY-MM-DD format>                |
 +-----------------+--------------------------------------------------------+
-| `Review Period` | <start - target end dates for review>                  |
+| Review Period   | <start - target end dates for review>                  |
 +-----------------+--------------------------------------------------------+
 
 Context
--- a/oeps/oep-templates/adr-based-template.rst
+++ b/oeps/oep-templates/adr-based-template.rst
@@ -4,31 +4,31 @@ OEP-XXXX: OEP Template
 
 .. This OEP template is based on Nygard's Architecture Decision Records.
 
-+-----------------+--------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


oeps/oep-templates/adr-based-template.rst
```diff
@@ -27,7 +27,7 @@ OEP-XXXX: OEP Template
 +-----------------+--------------------------------------------------------+
 | Created         | <date created on, in YYYY-MM-DD format>                |
 +-----------------+--------------------------------------------------------+
-| `Review Period` | <start - target end dates for review>                  |
+| Review Period   | <start - target end dates for review>                  |
 +-----------------+--------------------------------------------------------+
```
nasthagiri: @nedbat While you're here, can you update the templates to use the more maintainable formatting for these tables? See the [formatting change made in this PR](https://github.com/edx/open-edx-proposals/pull/128/files#diff-060835d3f05b6411b8ba8d35a0f852c6L5-R23).

oeps/oep-templates/pep-based-template.rst
```diff
@@ -45,10 +45,10 @@ this Open edX proposal (OEP) addresses.
...
 ==========
 
-The motivation is critical for OEPs that will change Open edX. Explain why the
-existing architecture or process is inadequate to address the problem that the
-OEP solves, or why adopting the best practice would significantly improve Open
-edX.
+The motivation is critical for OEPs that will change any part of the Open edX
+ecosystem. Explain why the existing architecture or process is inadequate to
+address the problem that the OEP solves, or why adopting the best practice
+would significantly improve the Open edX world.
```
nasthagiri: Is there a new recommended usage of `Open edX` that we should be aware of?

oeps/oep-templates/adr-based-template.rst
nedbat: Done and pushed here.

oeps/oep-templates/pep-based-template.rst
nedbat: As a trademark, "Open edX" should always be used as an adjective to modify some noun that we produce.  We and legal are just getting more disciplined about this. I should explain this to everyone somehow.

PR #131, opened 2019-12-09
OEP-17: Documentation guidelines

nedbat:


nedbat:
@robrap I'm not sure how best to describe these fields. It duplicates (somewhat) the existing "Report" section.

robrap:
@nedbat: 
-  I think @nasthagiri had mentioned that the bulk of the how-to documentation should live in https://github.com/edx/edx-toggles/ next to the code that should implement.  So, consider that as well, and possibly add minor notes here and pointers to docs that live there.
- Questions that I have when adding these annotations are:
  - What are all the annotations? I usually copy/paste from any place I find them.
  - What are the proper values for toggle_implementation, toggle_category, toggle_use_cases?

To avoid duplication, feel free to just add pointers to other docs where appropriate, but hopefully we can update to get terminology to match where we are duplicating.

sarina:
@nedbat @robrap the last action on this PR was two years ago. Could we close it for now?

robrap:
This OEP has had a bunch of updates with links for how to document, so I am going to close this.

PR #135, opened 2020-02-19
OEP-19: Dev Docs: add clarifications

robrap:
Moving discussions to github.

### DIFF
--- a/oeps/oep-0019-bp-developer-documentation.rst
+++ b/oeps/oep-0019-bp-developer-documentation.rst
@@ -3,7 +3,7 @@ OEP-19: Developer Documentation
 #################################
 
 .. list-table::
-  
+
    * - OEP
      - :doc:`OEP-0019 </oeps/oep-0019>`
    * - Title
@@ -70,12 +70,12 @@ OEPs
 ----
 
 .. list-table::
-   :widths: 15 75 
+   :widths: 15 75
 
    * - **Description**
      - Open edX Proposals (OEPs) are a collection of system-wide technology decisions made by the Open edX community, in the form of best practices, architecture designs, and processes.
    * - **Answers the questions**
-     - 
+     -
        * *What are the best practices and processes for developing on the Open edX platform?*
        * *How does system-wide framework X work on the Open edX platform? (e.g., Analytics, Authorization, etc).*
        * *Why was system-wide technology X or design Y chosen for the Open edX platform?*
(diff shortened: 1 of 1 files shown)


oeps/oep-0019-bp-developer-documentation.rst
```diff
@@ -203,7 +203,7 @@ This section summarizes the requirements for long-term developer documentation.
   * **Search**. For the time-being, we will make use of Google and GitHub's search functionality for finding content since all developer documentation will be in GitHub.  Certain IDEs may also have tools for searching.
 
   * **Index**. In the future, we will invest time in a better automated indexing tool to aggregate and structure the documentation. For the time being, we will manually create and maintain documentation indices at:
-    
+
     * edx-developer-docs_ will include the landing page for developer documentation and be a container of orphaned and cross-repo developer documentation.
 
     * docs.edx.org_ will be the main landing page for *all* Open edX documentation, including a link to the developer documentation index listed above.
```
robrap: **Reviewer Note:** I think we missed providing guidance around proper links, and I have felt that lack when writing new docs.

Questions that I have:
* When/where do I add links to published docs, rather than links to the source documentation?
  * Local links can be relative and hopefully will work in both contexts.
  * For links across repos, I have been erring on providing links to the source documentation (vs published docs). Can this be a decision?
* When linking to code, what is the best way to provide the link?
  * When linking to a specific code example, a github link with `blob/<commit-hash>` and line numbers is ideal.  Avoid `blob/master` links that will break as the file changes.
  * When linking to docstrings in code, what is the best route?
  * Links to `blob/master` with line numbers can easily break as code changes.
    * This could be misleading if it is broken in non-obvious ways by file changes that change the line numbers.
    * It can also be misleading if docs are published for a release and no longer point to links in the release.
  * Links to `blob/master` file **without** line numbers will help find the right file, which could be followed up by a manual search for a class or module-level docstring mentioned in the linking documentation.
    * This solves the problem of possibly pointing to the wrong lines, but is cumbersome.
    * This may still point from named release documentation to master documentation.
    * Based on the options I can think of, I think this might be the least evil.  It keeps documentation more automatically up-to-date, and hopefully someone will recognize when they have jumped too far into the future.
  * Links to `blob/<commit-hash>` with line numbers will provide the correct lines at a point in time, but could easily point to outdated docstrings.
    * Is there a way to automate flagging/fixing these links?
    * Should the link text remind users that the doc may be out of date?
  * Other ideas?

Github introduced the new code navigation features, but I don’t think they have a public API for it (that I could find).

robrap:
This PR can be reopened if we ever make progress in the above linked discussion.

PR #136, opened 2020-02-24
OEP-41: Asynchronous Server Event Message Format

ormsbee:
This OEP describes the general format and conventions Open edX should use in asynchornous event messaging across services. These events would be emitted when a service has taken some action to data that it owns, and that other services would be interested in. This could be a course publish, a user enrollment, updates to catalog metadata, etc.

This OEP does not cover:

* Remote Prodecure Calls, i.e. using messaging to send commands to other services. Events in this document are the services broadcasting, "Here is what I just did!" to interested parties, not "Please do this thing for me!"
* Real-time streaming of student learning events to external platforms, as detailed in [OEP-26](https://open-edx-proposals.readthedocs.io/en/latest/oep-0026-arch-realtime-events.html), though it is possible that the implementation of OEP-26 might build on the infrastructure outlined in this document.
* Browser level notifications and event streaming along the lines of Server-Sent-Events, GraphQL subscriptions, or frameworks built on top of WebSocket. The interactions here are strictly server to server interactions between services serving the same site. Again, it is entirely possible that an implementation of browser level notifications would be built on top of the infrastructure outlined here.
* The specific transport and libraries used for this messaging. That will be specified in a follow-on OEP.

### DIFF
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -16,7 +16,7 @@ OEP-41: Asynchronous Server Event Message Format
    * - Arbiter
      - Christopher Pappas <cpappas@edx.org>
    * - Status
-     - Draft
+     - Accepted
    * - Type
      - Architecture
    * - Created
@@ -24,9 +24,7 @@ OEP-41: Asynchronous Server Event Message Format
    * - Review Period
      - 2020-08-17
    * - Resolution
-     - ?
-   * - References
-     - ?
+     - Accepted
 
 --------
 Abstract
@@ -120,10 +118,12 @@ example:
 
   {
     "id": "b3b3981e-7544-11ea-b663-acde48001122",
-    "source": "https://discovery.edx.org/events"
+    "type": "org.openedx.catalog.course.created.v1",
     "specversion": "1.0",
     "time": "2020-02-23T09:00:00Z",
-    "type": "org.openedx.catalog.course.created.v1"
+    "source": "/openedx/discovery/web",
+    "sourcehost": "edx.devstack.lms",
+    "minorversion": 2,
     "data": {
         // message specific payload here
     }
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -151,10 +151,7 @@ string using Python's default behavior: lowercase and and dash-separated.
 `source <https://github.com/cloudevents/spec/blob/master/spec.md#source-1>`_
 -----------------------------------------------------------------------------
 
-Example:
 
-If it is being hosted under a different domain, is the source expected to be
-different?
 
 
 `specversion <https://github.com/cloudevents/spec/blob/master/spec.md#specversion>`_
@@ -255,6 +252,8 @@ majority of the time, but some events might have ``text/xml``.
 `time <https://github.com/cloudevents/spec/blob/master/spec.md#time>`_
 ----------------------------------------------------------------------
 
+Example: ``"2020-02-23T09:00:00Z"``
+
 Timestamp that the event occurred, in UTC using `RFC 3339
 <https://tools.ietf.org/html/rfc3339>`_. If this event was sent because we
 created a new row in the database, we should pull this ``time`` directly from
@@ -277,7 +276,7 @@ A callback is when you create a message with an ID or URL that you expect the
 consumer to make a synchronous call to when it receives the message. This is
 commonly used when an event represents some change that is too large to
 practically fit into the 64K message. For instance, we currently emit a generic
-"course_published" Django signal whenever data is published in Studio, leading
+``course_published`` Django signal whenever data is published in Studio, leading
 to a cascade of calls from various apps to the ModuleStore in order to extract
 the content data that they need.
 
@@ -298,7 +297,7 @@ load that my service is placing on the REST endpoint serving this user
 information. Is that safe? Who knows?
 
 One thing to consider is whether we can emit multiple events that better target
-specific consumer use cases. Let's take the "course_published" event as an
+specific consumer use cases. Let's take the ``course_published`` event as an
 example. Some listeners only care about schedule changes, because they have to
 account for when a course starts and ends. Search indexing really only wants to
 know the exact bit of content that was modified so that it can update that text.
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -140,9 +140,10 @@ considered optional in the CloudEvents spec).
 `dataschema <https://github.com/cloudevents/spec/blob/master/spec.md#dataschema>`_
 ----------------------------------------------------------------------------------
 
-Example:
+Example: ???
 
-
+WIP: This would link to documentation? Fill this out after we flesh out that
+portion of this OEP.
 
 
 `id <https://github.com/cloudevents/spec/blob/master/spec.md#id>`_
@@ -159,10 +160,12 @@ string using Python's default behavior: lowercase and and dash-separated.
 `source <https://github.com/cloudevents/spec/blob/master/spec.md#source-1>`_
 -----------------------------------------------------------------------------
 
-Example:
+Example: ``"https://openedx.org/events/catalog"``
 
-
-
+WIP: It's required that this field exist and recommended that it be an absolute
+URI. Use an openedx.org domain? Should there be extra information to indicate
+things like the site it comes from (in the whitelabel sense?). Or have that info
+in the data section?
 
 
 `specversion <https://github.com/cloudevents/spec/blob/master/spec.md#specversion>`_
@@ -289,6 +292,16 @@ actions. Do call ``datetime.now()`` if the event happens and has no
 corresponding database changes. If you are sending out multiple event messages
 describing the same occurance (e.g. a version 1 and version 2 of an event), they
 should have the *exact* same timestamp.
+
+Payload Metadata
+================
+
+The ``data`` attribute contains the application-specific message data.
+
+WIP: Do we require a common metadata header everywhere? In which case, we
+wouldn't be able to use other content types, which is probably fine? Need
+minor version info here, plus maybe site information (e.g. white label?). Some
+kind of global context information?
 
 
 Message Content Data Guidelines
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -367,6 +367,10 @@ event when a catalog entry for a course is created, while Studio
 first authored there. These are similar, related events, but they are not the
 same event type.
 
+To help prevent naming collisions, the set of messages will be centralized into
+a separate repository (or possibly a separate repository per logical subdomain).
+The details of this would be worked out in a follow-on OEP.
+
 
 Avoid Callbacks
 ---------------
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -118,7 +118,7 @@ example:
 
   {
     "id": "b3b3981e-7544-11ea-b663-acde48001122",
-    "source": "???"
+    "source": "https://discovery.edx.org/events"
     "specversion": "1.0",
     "time": "2020-02-23T09:00:00Z",
     "type": "org.openedx.catalog.course.created.v1"
@@ -160,13 +160,10 @@ string using Python's default behavior: lowercase and and dash-separated.
 `source <https://github.com/cloudevents/spec/blob/master/spec.md#source-1>`_
 -----------------------------------------------------------------------------
 
-Example: ``"https://openedx.org/events/catalog"``
+Example: ``"https://courses.edx.org/events"``
 
-WIP: It's required that this field exist and recommended that it be an absolute
-URI. Use an openedx.org domain? Should there be extra information to indicate
-things like the site it comes from (in the whitelabel sense?). Or have that info
-in the data section?
-
+This is an absolute URI that represents the URL of the server emitting the
+event. It does not have
 
 `specversion <https://github.com/cloudevents/spec/blob/master/spec.md#specversion>`_
 ------------------------------------------------------------------------------------
@@ -186,6 +183,10 @@ describing the type of event, but recommends that it be prefixed with a
 reverse-DNS name for namespacing purposes. We will use a period-separated
 hierarchical name comprising with the format ``{Reverse DNS}-{Architecture
 Domain}-{Subject}-{Action}-{Major Version}``.
+
+``version``
+-----------
+
 
 
 Reverse DNS
@@ -277,6 +278,19 @@ format of ``data`` attribute. It should be ``application/json`` the vast
 majority of the time, but some events might have ``text/xml``.
 
 
+``minorversion``
+----------------
+
+Example: ``2``
+
+This is the one extension attribute we would be introducing to CloudEvents and
+represents a minor version in semver reckoning, meaning that it increments when
+we have made backwards compatible additions to the message payload in the
+``data`` attribute. Values can only be integers. The initial value for
+``minorversion`` should be ``0``. There is no corresponding ``majorversion``
+because that information is encoded into the message type.
+
+
 `time <https://github.com/cloudevents/spec/blob/master/spec.md#time>`_
 ----------------------------------------------------------------------
 
@@ -293,15 +307,6 @@ corresponding database changes. If you are sending out multiple event messages
 describing the same occurance (e.g. a version 1 and version 2 of an event), they
 should have the *exact* same timestamp.
 
-Payload Metadata
-================
-
-The ``data`` attribute contains the application-specific message data.
-
-WIP: Do we require a common metadata header everywhere? In which case, we
-wouldn't be able to use other content types, which is probably fine? Need
-minor version info here, plus maybe site information (e.g. white label?). Some
-kind of global context information?
 
 
 Message Content Data Guidelines
--- a/oeps/oep-0041-arch-async-server-event-messaging.rst
+++ b/oeps/oep-0041-arch-async-server-event-messaging.rst
@@ -137,6 +137,14 @@ description of the conventions Open edX will use for that field. All Open edX
 events should have all of the fields specified below (even if some are
 considered optional in the CloudEvents spec).
 
+`dataschema <https://github.com/cloudevents/spec/blob/master/spec.md#dataschema>`_
+----------------------------------------------------------------------------------
+
+Example:
+
+
+
+
 `id <https://github.com/cloudevents/spec/blob/master/spec.md#id>`_
 ------------------------------------------------------------------
 
@@ -150,6 +158,9 @@ string using Python's default behavior: lowercase and and dash-separated.
 
 `source <https://github.com/cloudevents/spec/blob/master/spec.md#source-1>`_
 -----------------------------------------------------------------------------
+
+Example:
+
 
 
 
@@ -207,13 +218,21 @@ Example: ``course``
 
 The name of an entity that the event applies to. Examples might be ``course``,
 ``student``, ``enrollment``, ``order``, etc. Subjects may be namespaced, so
-``special_exam.proctored.allowance`` could be a subject. A subject should always
-mean the same thing within a Domain, but can mean different things across
-domains. For instance, what the LMS (``learning`` domain) calls a ``course``
-maps to what the ``catalog`` domain would call a ``course_run``. We should try
-to be consistent where possible, but each domain ultimately gets to decide what
-its terms means, and we should be careful when mapping a concept in one domain
-to concepts in another domain.
+``special_exam.proctored.allowance`` could be a subject.
+
+A subject should always mean the same thing within a Domain, but can mean
+different things across domains. For instance, what the LMS (``learning``
+domain) calls a ``course`` might map to what the ``catalog`` domain would call a
+``course_run``. We should try to be consistent where possible, but each domain
+ultimately gets to decide what its terms mean, and we should be careful when
+translating a concept from one domain to another. For instance, the
+``content_authoring`` and ``learning`` domains might both have a concept of a
+"due date" for an assignment. But while the ``content_authoring`` due date is
+determined only by the content author, the ``learning`` due date might take into
+account a student's cohort, individual due date extensions, accessibility
+allowances, and any number of other things. Both domains may call it ``due``,
+but the due date information from ``content_authoring`` is just an input to the
+more complex due date information in ``learning``.
 
 Action
 ~~~~~~
@@ -222,10 +241,16 @@ Example: ``created``
 
 This is the action that occurred for the event. Some of most common ones will be
 ``created``, ``updated``, and ``deleted``, but many applications will want more
-specific actions like ``declined``, ``started``, ``verified``, etc. Actions
-should be past tense, to better align with our existing conventions around
-Django signals and learning analytics events (we're not completely consistent,
-but we tend towards past tense).
+specific actions like ``declined``, ``started``, ``verified``, etc. Applications
+should prefer these more specific actions when possible, since they are usually
+clearer for listeners and don't requiring knowing implementation details. For
+instance, it is preferable to have a ``registration.completed`` event than to
+have a ``registration.updated`` event and leave it to clients to check the
+resulting message's payload for ``status: "complete"``.
+
+Actions should be past tense, to better align with our existing conventions
+around Django signals and learning analytics events (we're not completely
+consistent, but we tend towards past tense).
 
 Major Version
 ~~~~~~~~~~~~~
@@ -269,6 +294,35 @@ should have the *exact* same timestamp.
 Message Content Data Guidelines
 ===============================
 
+These are general guidelines to consider when creating your messages. There can
+always be exceptional circumstances and use cases that require going against one
+of these guidelines, but try to default to these guidelines unless you're really
+sure about what you're doing.
+
+One Producer Service Per Event Type
+-----------------------------------
+
+Each event type should be emitted by one, and only one, service. That service is
+the source of truth for whatever entity the event describes. If course-discovery
+emits an event describing when a course starts (e.g.
+``org.openedx.catalog.course.start_date.changed``), it is *not* appropriate for
+the LMS to send potentially conflicting information using that same event type.
+
+Two services may have similar sounding events. The course-discovery service
+(``catalog`` domain) might emit a ``org.openedx.catalog.course.created.v1``
+event when a catalog entry for a course is created, while Studio
+(``content_authoring`` domain) might emit a
+``org.openedx.content_authoring.course.created.v1`` event when course content is
+first authored there. These are similar, related events, but they are not the
+same event type.
+
+Events are Created by the Owning Domain
+---------------------------------------
+
+
+
+
+
 Avoid Callbacks
 ---------------
 
@@ -283,18 +337,18 @@ the content data that they need.
 Callbacks threaten performance and stability because they reduce a service's
 ability to control its own load. For instance, a sudden increase in Courseware
 traffic might generate a burst of student analytics events. If this stream of
-events overwhelms my service's ability to consume them, the queue may start to
-back up with unread events. Yet this shouldn't cause my service to fail, since
+events overwhelms your service's ability to consume them, the queue may start to
+back up with unread events. Yet this shouldn't cause your service to fail, since
 it still gets to control how quicky it consumes events off of that queue. It has
 the freedom to either slowly catch up (if the burst was a momentary spike), or
-to scale up additional workers to handle the higher throughput. My service's
+to scale up additional workers to handle the higher throughput. Your service's
 decision to scale up or down does not directly impact other services.
 
 Things change when we introduce a callback to this same scenario. Say the
 analytics events now include a callback URL to get basic user information. In
-this scenario, doubling the consumers that my service has now also doubles the
-load that my service is placing on the REST endpoint serving this user
-information. Is that safe? Who knows?
+this scenario, doubling the consumers that your service has now also doubles the
+load that your service is placing on the REST endpoint serving this user
+information. Is that safe to do?
 
 One thing to consider is whether we can emit multiple events that better target
 specific consumer use cases. Let's take the ``course_published`` event as an
@@ -311,6 +365,71 @@ when a user changes their profile information might include a URL to the S3
 location of their new profile picture. Just keep in mind that messages may be
 read long after they're generated, and any presigned S3 URLs you generate might
 be expired by the time a consumer gets them.
+
+
+Architectural Goals
+===================
+
+Asynchronous messaging is often promoted as a way to decouple services from each
+other and make them more resilient to failure. What does that mean for Open edX
+in concrete terms? How does this OEP align with the Achitecture Manifesto?
+
+Autonomy beats Coordination -> Control over domain messaging, individually versioned messages.
+
+Extensions/Plugins -> Encouraging further decoupling by elminating even code dependencies
+
+Data Redundancy beats Coupling Acorss Contexts
+
+Loosely Coupled Codebase (versioning, repeating yourself, separate classes for minor version
+updates, mixed together?)
+
+Asynchronous Messaging beats Synchronous Requests
+
+Eliminate Expensive, Batch Synchronization
+------------------------------------------
+
+We currently have processes that make extremely expensive requests across
+services to synchronize data, particularly from course-discovery. Aside from
+being wasteful (relatively few entries actually change between these batches),
+it is also difficult to properly scale for. Most auto-scaling is built to
+trigger once particular thresholds around CPU are hit, and does not deal well
+with momentary bursts of load that are 5-10X normal levels. We are forced into
+some combination of over-provisioning and/or accepting periodic spikes in
+overall service latency. In a situation where we are already running near
+capacity, a spike like this might be enough to trigger a cascade of failures.
+
+Eliminate Blocking, Synchronous Requests
+----------------------------------------
+
+
+
+
+Reduce the need for Plugins
+---------------------------
+
+`Django app plugins <https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/plugins>`_
+exist to help decouple core edx-platform code from third party extensions. But
+in some cases, the only reason the plugin needs to exist at all is because there
+is no other way to get notifications for important lifecycle events like user
+registration, course enrollment, score changes, etc. Having a stable set of
+event APIs will allow many of these to exist as independently deployed services
+that don't need to be run in the same process as the LMS.
+
+
+
+Simplify Integration to External Systems
+----------------------------------------
+
+( Remove direct dependencies between parts of the system. )
+
+
+
+
+
+
+* easier to substitute other systems
+* don't have to add as much to edx-platform itself.
+
 
 
 Scratch Notes
@@ -334,6 +453,8 @@ cases, schedule data, etc.)
 
 1. Self-contained, without history
 2. PII
+Only one producer per event.
+
 4. No RPC
 5. Exchange = Domain, routing key = event type. Namespacing excessive?
 6. Shouldn't emit before data is committed to the database. on_commit practice?
@@ -357,6 +478,7 @@ Documentation
 
 (AsyncAPI)
 
+Security
 
 Top level concerns:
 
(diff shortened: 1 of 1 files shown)


oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,203 @@
+=============================================
+OEP-41: Asynchronous Server Event Messaging
+=============================================
+
+.. list-table::
```
cpennington: What is this `list-table` magic? Amazing!

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+
+`id <https://github.com/cloudevents/spec/blob/master/spec.md#id>`_
+------------------------------------------------------------------
+Example: ``"b3b3981e-7544-11ea-b663-acde48001122"``
+
+Message ID. CloudEvents requires this to be a string that is unique for any
+given source. We will use a UUID1 here, since it is easy to generate without
+worrying about collisions, and it offers slightly more useful introspection
+(timestamp and machine) than a random UUID4. The UUID will be encoded as a
+string using Python's default behavior: lowercase and and dash-separated.
```
ormsbee, line 149: Typo: and and

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+
+`type <https://github.com/cloudevents/spec/blob/master/spec.md#type>`_
+----------------------------------------------------------------------
+Example: ``"org.openedx.catalog.course.created.v1"``
+
+This is the name of our event. CloudEvents only requires that this be a string
+describing the type of event, but recommends that it be prefixed with a
+reverse-DNS name for namespacing purposes. We will use a period-separated
+hierarchical name comprising with the format ``{Reverse DNS}-{Architecture
+Domain}-{Subject}-{Action}-{Major Version}``.
```
ormsbee, line 181: '.'s not '-'s

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+around Django signals and learning analytics events (we're not completely
+consistent, but we tend towards past tense).
+
+Major Version
+~~~~~~~~~~~~~
+Example: ``v1``
+
+The last portion is the major version of the message, starting with ``v1``. All
+messages will have a major and minor version, with the minor version encoded in
+the ``data`` attribute (todo: link to where we discuss that). New fields may be
```
ormsbee, line 253: Switched this to a new attribute instead.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+<https://github.com/cloudevents/spec/blob/master/spec.md>`_ specification, with
+additional usage conventions that all Open edX services should adhere to.
+CloudEvents was developed by the Cloud Native Computing Foundation, with
+participation from major companies like Google, Microsoft, IBM, Red Hat, Oracle,
+and Huawei. While the finalized 1.0 spec is only months old and the Python
+tooling is primitive (advertised to break at any time), CloudEvents is also a
+pretty minimal message envelope spec, with the basic required fields being
+things that we'd want to have anyway.
+
+Event messages will be UTF-8 encoded JSON documents of 64 KB or less. An
```
christopappas, line 114: Reading from top to bottom (i do see a thing about callbacks but haven't read yet), but I do wonder about the 64kb limit. Not sure if that'll be enough for huge nested discovery data bits, especially if we want our messages to be stateless (And therefore contain ALL the information). probably work-around-able, but maybe something to be mindful of

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+
+
+-------------
+Specification
+-------------
+
+Message Format
+==============
+
+The message format will use the `CloudEvents 1.0
```
christopappas, line 104: my initial thoughts on the spec are: good. i think it’s worth adopting something other places are using to start with, and then we can make any additions/changes as we see fit 👍

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: Do you mind posting a concrete example (with ID and text changed if that's necessary)? I know course-discovery has some heavily connected data models, but it's possible that we might be able to tease apart the updates into multiple events that are more reasonably sized. I think this sort of thing is likely to come up a lot, and I'd love to have some good examples here.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+but the due date information from ``content_authoring`` is just an input to the
+more complex due date information in ``learning``.
+
+Action
+~~~~~~
+Example: ``created``
+
+This is the action that occurred for the event. Some of most common ones will be
+``created``, ``updated``, and ``deleted``, but many applications will want more
+specific actions like ``declined``, ``started``, ``verified``, etc. Applications
```
christopappas, line 236: 🤔  -- Say I have some new python microservice and I want to start consuming messages from the LMS, I could probably have some code in a consumers file like:

```
if message.get('action') == 'created':
    run_my_ingest_logic()
elif message.get('action') == 'update':
    run_my_update_logic()
else message.get('action') == 'deleted':
    run_my_delete_logic()
```

How will I know what 'action' verbs messages from a particular service can include? Are they easily discoverable? Do I need to look at the source code of the producing service?

If we have a set vocabulary (even if only created, updated, deleted), it does simplify what the consumer needs to know about implementation details of the producing service. Granted, 90% of things are probably updates, but perhaps knowing `updated` + the subject and domain might be enough info to write your consumer code to do the thing it wants

Thoughts?

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+instance, it is preferable to have a ``registration.completed`` event than to
+have a ``registration.updated`` event and leave it to clients to check the
+resulting message's payload for ``status: "complete"``.
+
+Actions should be past tense, to better align with our existing conventions
+around Django signals and learning analytics events (we're not completely
+consistent, but we tend towards past tense).
+
+Major Version
+~~~~~~~~~~~~~
```
christopappas, line 248: Question about versioning... I think we need to make sure we outline a process for deprecating old major message versions. I'm imagining a future where we're on V4 of some message in the LMS, but still need to send messages V1 2 and 3 (otherwise we break someone's consumer), resulting in us sending a flurry of messages all the time something something queue size something something memory limit :)

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: Unpacking a few things...

**Dispatching by Event Type**

I think we're likely to have some kind of simple registration scheme where you can specify what kind of message you want, like:

```python
@handle_event("org.openedx.catalog.course.created.v1")
def do_something(event):
    # ...
```

**Knowing what a Service Offers**

This is something that we're going to have to fill with docs, but there are potentially going to be a slew of possible message types at different minor versions throughout the system. So I do think we also need to be able to generate a list of them, e.g. via some management command. I'm still fuzzy on the details, will try to add that to this OEP in the next week. I'd like to create a fairly small library, probably based on `attrs` that lets you cleanly separate minor versions and run basic sorts of introspection for reporting purposes. I'm not sure how to tie it to documentation at the moment (AsyncAPI?).

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: I agree, I need to really flesh out the versioning story here, especially given how granular these contracts effectively are since individual messages have versions.

That being said (and I hope these words don't bite me later), I don't think we should be too concerned about writing v1, 2, 3, 4 of a message. Backwards compatible minor version bumps should be preferred, of course, but writing multiple major versions of a messages grows slowly and linearly over time. I'm hoping that simple event emission is less of a dev support burden than maintaining bidirectional APIs. My big concern for message spew are if we do things where we the number of messages sent grows with the size of the thing being worked on, for instance if we have some action on the course that would unenroll a thousand students, and we were trying to send all those messages in a synchronous transaction.

I honestly hadn't really thought about that before now. That could really bite us if we start tying messages to model level interactions. Argh. Thank you.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+==================
+
+All top level field names are specified by the CloudEvents spec, but most fields
+defined by that standard have some leeway in how they're used for any given
+system. Each field below has a link to CloudEvents requirements, as well as a
+description of the conventions Open edX will use for that field. All Open edX
+events should have all of the fields specified below (even if some are
+considered optional in the CloudEvents spec).
+
+`id <https://github.com/cloudevents/spec/blob/master/spec.md#id>`_
```
chavesj, line 141: The heading for each CloudEvents Field is inconsistent; some headers are hyperlinked and some aren't.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+course-discovery PR?) to test that this works properly...
+
+Every domain will map to a Kombu Topic Exchange, and services will be Producers
+that send messages to those Exchanges. The Exchange name and message Routing Key
+will be derived from the event ``type``. The parts of the ``type`` up to and
+including the Domain will be the name of the exchange. The rest of the ``type``,
+starting from the Subject will be the Routing Key. So the
+``org.openedx.catalog.course.created.v1`` event would go to the
+``org.openedx.catalog`` Exchange, and use ``course.created.v1`` for the Routing
+Key.
```
ormsbee, line 419: Thinking on this more, I wonder if it's better to just make it one unified exchange. One of the original ideas behind breaking up the exchanges was to give a sense of ownership (e.g. "only `catalog` can write to this exchange"), but the namespacing of the message types themselves might be sufficient for that. And it would be one less source of configuration.

Another consideration might be if there are security/privacy concerns and certain specialized exchanges. For instance, there might be a special exchange that is only for anonymized learner analytics events...?

Anyhow, I'll probably collapse this part of the proposal and just have one Exchange for now and use Routing Keys for all the separation. Please comment here if you have an opinion on this.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+
+Message ID. CloudEvents requires this to be a string that is unique for any
+given source. We will use a UUID1 here, since it is easy to generate without
+worrying about collisions, and it offers slightly more useful introspection
+(timestamp and machine) than a random UUID4. The UUID will be encoded as a
+string using Python's default behavior: lowercase and and dash-separated.
+
+`source <https://github.com/cloudevents/spec/blob/master/spec.md#source-1>`_
+-----------------------------------------------------------------------------
+Example: ``"https://discovery.edx.org/events/"``
```
regisb, line 153: The source identifier should not change over time. Yet, it is quite possible to switch an application from http to https, and vice versa, based on just a configuration value, so I do not recommend including the scheme in the source name.

Also, should we use the external or the internal domain name? The edX discovery service might be available at discovery.edx.org, but it might also be available at discovery1.edx.org. If the discovery service runs inside a docker-compose or Kubernetes cluster, its internal domain name will simply be "discovery".

Thus, I suggest  we do not rely on http/https urls as the source name. Instead, I propose an identifier similar to the "application-specific identifiers" described in the spec: https://github.com/cloudevents/spec/blob/master/spec.md#source-1

Something along the lines of: "/openedx/discovery/events".

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+location of their new profile picture. Just keep in mind that messages may be
+read long after they're generated, and any presigned S3 URLs you generate might
+be expired by the time a consumer gets them.
+
+Transport Layer
+===============
+
+Events will be sent and received using Kombu, the messaging library that celery
+uses. We will use Redis as the backing server. This allow us to implement this
+messaging scheme without adding any new server dependencies.
```
regisb, line 404: This means that fanout will only be supported with pub/sub: https://github.com/celery/kombu#transport-comparison
In practice, it means that some disconnected clients will miss messages when they reconnect. Isn't that an issue? Mongodb is also part of the Open edX stack -- I think it would be worth considering this alternative.

ormsbee: I guess I kind of assumed that we'd use topic exchanges for the most part, as opposed to fanout. I really don't have any operational experience using Redis in this way though, so I could be totally mistaken. Part of the reason why there hasn't been an update to this OEP for a while is because I'm working on a prototype to kick the tires on a few of these things.

ormsbee: As for MongoDB... it's beyond the scope of this OEP, but I was hoping that one day we could actually remove MongoDB from the overall stack. There are only two central uses of it right now: Modulestore/Contentstore (where we mostly use it as a key value blob store), and the forums.

That being said, I proposed Redis as the transport mostly because it seemed the easiest thing to administer that's already in our stack, and we already use it as the default for celery. At the code layer, I'd want to just use Kombu and be transport agnostic anyhow. If folks want to use MongoDB or even RabbitMQ, that should be fairly straightforward to set up.

regisb: It's important to understand that with pub/sub (and thus with Redis) we do not get exactly once delivery, but only at most once delivery. See https://redislabs.com/redis-enterprise/use-cases/messaging/ ("Messaging semantics must be clear")

For instance:

0. subscriber and publisher boot up
1. subscriber crashes
2. publisher emits event
3. event is not received by subscriber
4. subscriber comes back up

This is a very frequent scenario, for instance in deployments without downtime. Sure, you might be doing blue/green deployments properly, but crashes happen nonetheless. If there is a simple syntax error in the subscriber, then all events that will be received will be thrown away. Sure you might guarantee that you have a fallback option that catches errors and re-schedules events in 10 minutes -- but will that work? What if there is an error in the fallback code?

My point is that we will have to engineer a very complex solution just to make up for the failings of the pub/sub model.

I understand that you want to simplify the Open edX, and I can sympathize with that. But please don't displace complexity by moving database features to the code base. 

Apparently, [Redis Streams](https://redis.io/topics/streams-intro) is a redis-backed solution for at least once delivery. I didn't know this alternative to Kafka existed. However, Redis Streams does not seem to be compatible with Kombu. Maybe it's time to reconsider the use of Kombu?

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: Thank you for bringing this up! This was quite possibly the hardest field for me to reason about, and I'm grateful to get feedback on it. I'm not really happy with what I've got written here, but I'm not sure about the implications of something like "/openedx/discovery/events" either.

I originally had an app specific hierarchy like you propose, but I was later swayed by the CloudEvents spec recommendation that it be an absolute URI (which I realize is a recommendation and not a requirement). I'm definitely not tied to the current representation, but I do want to try to figure out exactly what it is we want to accomplish with this field. The only thing the spec really requires is that this value is a string and that `source + id` is unique (which we get for free anyway since we use UUIDs), so we have a lot of freedom in how we want to interpret this field.

One goal could be for the source to represent a sort of top level schema, e.g. "This is a family of events associated with a subset of Open edX." I believe that's what they're going for with their `https://github.com/cloudevents` example. I feel like we could accomplish this with `dataschema` if we wanted to. Another possible goal would be to map to something more concrete and physical with respect to identifying the producers. The spec also gives the example of having a unique identifier for every distinct producer.

We already encode the logical hierarchy in the message type, e.g. `org.openedx.catalog.course.created`. I felt that it would be redundant to have the source represent that concept. So I wanted to focus on the issue of, "Where did this event originate?" for debugging purposes.

> The source identifier should not change over time.

I think there are at least three use cases where the "source" service might change over time:

1. Service is renamed, e.g. course-discovery to catalog.
2. A service is refactored or broken up into multiple services that still share responsibility for some domain. Downstream processes listening for events still just listen for the event type they're interested in and don't notice any changes.
3. An Open edX service is entirely replaced by another system that emits the same events to make it easier to maintain compatibility. If the LMS is going to listen for course run `start_date` changes coming from the catalog, you could imagine something like Richie taking over that role of emitting those events for some organizations.

So for those use cases, I'm actually assuming that the source for a given message can change, and that it's a bad smell if any of our code relies on it being stable or on introspecting it. I went with absolute URIs because I thought this would simplify the namespacing and make it easier to debug issues. But I like your suggestion better and I think it could get us all the things that I really want from debugging, if we did something like:

source = "/{namespace}/{app}/{worker|web}"

So in the current scheme of things:
* "/openedx/discovery/web" for something coming from Django web requests to the course discovery service.
* "/openedx/discovery/worker" if the event is coming from one of its celery workers.

If we're logging that, it would be easy to track down where the message is coming from, and especially if we have multiple producers of the event. It would also have the added benefit that if something popular-but-beyond-our-control emits events with a weird data quirk, we could find and "fix" those events by inspecting their source. Though that should be avoided whenever possible.

A downside is that the whole "aid debugging" goal completely fails if people start relying on the source and producers have to start lying to not break things...

Does that sound good to you?

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: I agree that we need to account for subscribers crashing, going away for a while, and then coming back and not missing all the events in the intervening time. I agree that it's a frequent scenario. I have no desire to re-implement a message queue to get things onto our message queue.

My understanding (and this could be wrong), is that Kombu only uses Redis PUB/SUB for fanout exchanges, and that topic exchanges are implemented with Redis lists and sets (i.e. LPUSH, BRPOP), which would then survive client disconnect. I don't know the limits of this because I don't really know the specifics how how Kombu arranges these data structures internally, so I definitely need to test this.

> Apparently, Redis Streams is a redis-backed solution for at least once delivery. I didn't know this alternative to Kafka existed. However, Redis Streams does not seem to be compatible with Kombu. Maybe it's time to reconsider the use of Kombu?

So I am biased here. I helped to write one of the Python drivers for Kafka in the very early days of Datadog, and I'm enormously fond of the project. I think their design trade-offs were incredibly clever, though the ZooKeeper stuff did make me miserable for a few weeks. We've looked at Redis streams briefly when talking about learner analytics tracking logs, as it has many of the features we want without adding something drastically new to our stack. Being able to rewind time and reprocess things is like a crazy super power, and capped streams would be a great compromise between allowing for failure without having data exposed forever.

That being said, I think that event logs like Kafka and Redis Streams are more effort to set up for our common use cases, where we have a wide range of event types and relatively lower throughput. With Kombu, it's really easy to match on a handful of patterns using routing key wildcards, say if you want to audit all GDPR related events throughout the system. Using Kafka or Redis Streams requires more up-front planning about how those streams/topics are going to be arranged.

If it turns out that Redis + Kombu doesn't meet our reliability requirements, I'd probably next see if Kombu + RabbitMQ would work. I'm afraid that adopting Kafka or Redis Streams as a general tool will require us to plan more and write more in terms of code. But I haven't followed the Kafka space closely in a while, so if there are libraries or frameworks that you recommend, I'd be happy to look at them after I do a round of Kombu + Redis testing.

oeps/oep-0041-arch-async-server-event-messaging.rst
regisb: Sounds very good! :ok_hand: Thanks for the detailed answer :)

oeps/oep-0041-arch-async-server-event-messaging.rst
regisb: > My understanding (and this could be wrong), is that Kombu only uses Redis PUB/SUB for fanout exchanges, and that topic exchanges are implemented with Redis lists and sets (i.e. LPUSH, BRPOP), which would then survive client disconnect. I don't know the limits of this because I don't really know the specifics how how Kombu arranges these data structures internally, so I definitely need to test this.

Yes, this is most likely. The problem with Kombu + Redis (or Rabbitmq for that matter) is that only a single subscriber can receive every message. In other words: there is no fanout. It means that we will not be able to process a single message for different purposes. Let's take an example: when we receive a "subscribed to course" event, we would like both to:

1. warn the course staff about a new subscription
2. forward this event to our analytics system

Without fanout, a worker doing step 1. will have to emit another event in order to be processed by a worker performing step 2.

I am putting Rabbitmq in the same bag as Redis because fanout is implemented by sending multiple copies of the same message to different queues: https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.transport.virtual.exchange.html#fanout This really seems backwards once you've played around with Kafka.

It's cool that you've worked with Kafka before! In my experience few people in the python community have. It's a fantastic tool that is really hard to get right, in part because of the Zookeeper requirement, in part because of the [horribly](https://docs.confluent.io/current/installation/configuration/topic-configs.html) [complex](https://docs.confluent.io/current/installation/configuration/producer-configs.html) [configuration](https://docs.confluent.io/current/installation/configuration/consumer-configs.html). But that's a conversation for another time...

> That being said, I think that event logs like Kafka and Redis Streams are more effort to set up for our common use cases, where we have a wide range of event types and relatively lower throughput. With Kombu, it's really easy to match on a handful of patterns using routing key wildcards, say if you want to audit all GDPR related events throughout the system. Using Kafka or Redis Streams requires more up-front planning about how those streams/topics are going to be arranged.

I don't think we need more planning, but a different implementation. With wildcard pattern matching, I expect you wanted to write something like:

```
def wildcard_consumer(event_name, *args, **kwargs):
    event_type = event_name.split(",")[1]
    if event_type == "subscribe":
        subscribe_consumer(*args, **kwargs)
    elif event_type == "unsubscribe":
        unsubscribe_consumer(*args, **kwargs)
    ...
```

Is that right? With Redis Streams, the event type would be passed as an argument of the event:

```
def assigned_consumer(event_name, event_type, *args, **kwargs):
    if event_type == "subscribe":
        subscribe_consumer(*args, **kwargs)
    elif event_type == "unsubscribe":
        unsubscribe_consumer(*args, **kwargs)
    ...
```

So the consumer would only process events from a given stream, but we could pack different events in that stream. Would that work for you?

> If it turns out that Redis + Kombu doesn't meet our reliability/persistence-on-disconnect requirements, I'd probably next see if Kombu + RabbitMQ would work. I'm afraid that adopting Kafka or Redis Streams as a general tool will require us to plan more and write more in terms of code. But I haven't followed the Kafka space closely in a while, so if there are libraries or frameworks that you recommend, I'd be happy to look at them after I do a round of Kombu + Redis testing.

I don't have any experience with Redis Streams, but in my understanding the required tooling is very small. Basically, all we need is Redis v5.0+ and an up-to-date python client. I think it would be a shame to lose the ability to fanout so early in the design process. Also, adding back Rabbitmq to the stack would be a real step backwards, as it was removed more than a year ago.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+roughly match deployed services. Domain names should be lower cased and use
+underscores if they are more than one word.
+
+Subject
+~~~~~~~
+Example: ``course``
+
+The name of an entity that the event applies to. Examples might be ``course``,
+``student``, ``enrollment``, ``order``, etc. Subjects may be namespaced, so
+``special_exam.proctored.allowance`` could be a subject.
```
bryanlandia, line 214: @ormsbee So, recording events to the tracking or sending to a BI backend is not the same thing as interservice communication but it does seem that the event naming could/should be consistent.  Do you imagine deprecating the existing event naming schemas, so something like `edx.bi.user.account.registered` would become something like `org.openedx.learning.student.account.created.v1`?  To my mind, it would be good to maintain consistency of event naming schemas.

ormsbee: @bryanlandia: Good point, and I'll add a section to the OEP to try to address it. My general philosophy on events is that backwards compatibility trumps naming consistency. A couple possible approaches:

### 1: Namespace and Encapsulate in CloudEvents

We could put all existing events in a namespace like `org.openedx.legacy.{original event name}`, extract one or two relevant fields, and otherwise echo the event as is in the `data` section.

### 2: Create Parallel Exchange(s)

Instead of trying to retrofit the events to match the new conventions, we create one or two new exchanges. We use the existing event type names as routing keys, but the events are transmitted in exactly the same format they are now, and no attempt is made to modify them to meet any of the conventions proposed in this OEP for new events. We might have a couple of extra exchanges, e.g. one for learner tracking events and one for BI stuff.

New events are added with the new conventions, which I hope will improve versioning and usability overall. Some actions trigger both old and new events. Listeners can register on multiple exchanges. While the listener code might have to change, the code that actually processes the response remains exactly the same.

----

My inclination would be option 2 here. What you you think?

oeps/oep-0041-arch-async-server-event-messaging.rst
christopappas: Have a look at the course metadata returned by course discovery for ACCA+FA1-MA1.X -- the json metadata returned for this is 107kb

Now, this definitely is related to course runs being embedded in the metadata, and like you mention, we definitely could be judicious about how we serialize our messages... although that does bring up the stateless topic: when we're trimming metadata that is returned from a serializer, we'll need to make sure that whatever is returned is stateless in addition to being sub 64kb 

Let me know if this is enough to go on, it's just one example. Happy to dig up more!

oeps/oep-0041-arch-async-server-event-messaging.rst
christopappas: Question that came to mind: what happens when we decide to change the action of a message that is already being consumed by others. We register a new message version right?

If so we'll definitely want to stress the importance to people to get their action verb right the first time.

oeps/oep-0041-arch-async-server-event-messaging.rst
christopappas: Alternative for a special exchange that has PII-like stuff: encryption? I think having 1 exchange simplifies things from an architectural POV, and perhaps encrypting values of sensitive values gives the added privacy needed? There then is the question of how to decrypt... and maybe that would be annoying (sharing keys around or something). If a key gets rotated, everyone needs to update their service to decrypt. Maybe a bad idea, putting it here anyway in case it sparks discussion

Overall: I like having 1 exchange

christopappas:
When you get a chance, if you wouldn't mind updating the review dates, that would be 👍  :)

oeps/oep-0041-arch-async-server-event-messaging.rst
bryanlandia: ugh, sorry @ormsbee overlooked your response here and appreciate your thoughts.  I'll give this some thought and try to get back to you by end of week.

oeps/oep-0041-arch-async-server-event-messaging.rst
feanil: I like the new format you're thinking about @ormsbee , do you think source sort of source machine identifier could be added too?  I'm not sure what the intent of the source identifier shouldn't change is.  One way I'm thinking about it is that the source shouldn't change for a given producer but we could think of each machine as a new source that comes and goes.  This makes it harder for people to depend on it but gives us a lot of info on where things are coming from.

ormsbee: @feanil: 

> I like the new format you're thinking about @ormsbee , do you think source sort of source machine identifier could be added too?

I'm inclined to put that in a different field, if we do. Maybe `sourcehost`, and put the hostname in there? Having it as a separate field would make it easy to log/facet/filter by host, independent of message type, or if for some reason multiple worker types are running on the same host.

> I'm not sure what the intent of the source identifier shouldn't change is. One way I'm thinking about it is that the source shouldn't change for a given producer but we could think of each machine as a new source that comes and goes. This makes it harder for people to depend on it but gives us a lot of info on where things are coming from.

I think that in the above scenario, `source` is sort of the "logical/semantic source" while `sourcehost` would be the "physical source". My hypothesis (for which I have zero data) is that having that split will make it easier to handle slight incompatibilities that might arise if other alternate services (like Richie) decide to start emitting messages that the LMS reads. If one party doesn't have much control over the other, you could imagine a bit of handling on the client side like:

```python
# The Swampcastle Catalog doesn't emit UTC dates like you're supposed to...
if source == 'swampcastle/course_catalog/web':
    # some fix here...
```

I'm a little torn, because a) it's not a great pattern to encourage; and b) I'm not sure it'll ever be an issue. But I think that it's plausible and that we do get other benefits by decoupling the logical (code) and physical (server) in reporting this information.

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: > Yes, this is most likely. The problem with Kombu + Redis (or Rabbitmq for that matter) is that only a single subscriber can receive every message. In other words: there is no fanout. It means that we will not be able to process a single message for different purposes.

I just wrote a couple of experiment scripts, and it does look like we get this capability with Kombu. I set up three instances of [consumers](https://gist.github.com/ormsbee/23616856bbbb11de35dd88f59e984c32) with different queues that connected to the same exchange. One consumer was set to listen for the routing key `org.openedx.catalog.banana.thrown.v1`, another for `*.banana.*`, and the third for `nomatch.key`. When a separate [producer](https://gist.github.com/ormsbee/d5c684b127ffb3accd9cad0d76d070cf) process sent `org.openedx.catalog.banana.thrown.v1` messages, the first two consumers got them while the third did not.

I mean, yes, under the covers it's likely being implemented as pushes to separate queues, but does that really matter for our purposes? The interface is fairly light from the client end. It also seems lighter than having every client examine every message in the stream to filter out the ones they actually care about.

I do have some longer term concerns about robustness with the way Kombu uses redis (I've tested disconnects and catching up once a client reconnects, and things seem to be okay, but I haven't stress tested it). I agree that being able to deliver the same message to multiple consumers is critical, and that we absolutely do not want some crazy chain of clients echoing events out. But based on my reading and experiments, Kombu seems to meet our requirements in these areas.

Would you feel comfortable starting with Kombu if it was kept as a swappable implementation detail? The touch points are fairly minimal (get a message, send a message, ack a message, subscribe to a pattern of messages, maybe some exchange-like concept)–I expect it'll be fairly straightforward to keep those details contained to a small module.

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: So I think there's an important (if sometimes blurry) distinction between stateless and denormalized. For instance, if a course instructor's `given_name` changes, I think we should send the whole staff record (`salutation`, `family_name`, etc.) I _don't_ think we should send the full data for every course run they're an instructor in.

My guess is that downstream systems that want to replicate data usually care only about a small portion of the data model, and that they're better served by stateless-but-normalized updates. So for instance, if a course run is added to a program, it's helpful for the LMS to have a message that says, "Hey, the new list of course runs for this program is [...]". But that can be a simple normalized list of IDs without all the details.

There will be exceptions of course. We're never going to echo across the list of students or the list of enrollees for a course, though it might be practical to echo the full enrollment list for an individual student. Though even then, possibly not, since that is not currently bounded.

There's also the matter of event granularity / multiple logical events. For instance, enrolling in a course and unenrolling in a course are semantically distinct and more useful as separate events than a generic, "something was updated with an Enrollment record" event. It's also probably worth specifying "enrolling myself" vs. "re-enrolling" vs. "the professor enrolled me without asking", which isn't strictly captured in that one table.

I do think there are a bunch of odd corners of this that we'll eventually have to figure out. Such as: What happens when there are related updates (e.g. a new course run is created, _and_ added to program)? Message processing is not always guaranteed to be in-order. Even if it was, it's always possible for two related messages to be distributed to two different consumers that will not see each others changes in time. In practice, I think it'll work out okay when we're trying to replicate out small bits, but even then we may have to relax certain integrity constraints...

ormsbee: @macdiesel, @doctoryes: FYI in case you had thoughts on this.

oeps/oep-0041-arch-async-server-event-messaging.rst
feanil: I like the clarity of `sourcehost` and I think that would be useful.  For you concern about sources with incompatibilities, would some sort of schema validation/registry not resolve that? Though I do agree having a logical service this event came from is a useful piece of metadata in general.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,486 @@
...
+
+This OEP describes the conventions Open edX should use for asynchronous event
+messaging across services. These events would be emitted when a service has
+taken some action to data that it owns, and that other services would be
+interested in. This could be a course publish, a user enrollment, updates to
+catalog metadata, etc.
+
+This OEP does not cover:
+
+* Remote Prodecure Calls, i.e. using messaging to send commands to other
```
timmc-edx, line 43:
```suggestion
* Remote Procedure Calls, i.e. using messaging to send commands to other
```

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: Finally getting back revising this. I don't think schema validation/registry will be enough to completely resolve those types of incompatibilities. In previous lives, I've seen all kinds of random incompatibilities published by vendors for what should have been the same spec like:

* Putting HTML escape codes vs. not.
* Using a literal non-breaking space character vs. a space (i.e. U+00A0 vs. U+0020).
* Encoding strings in Latin-1 vs. UTF-8.
* Sending offsets 4-byte aligned vs. not.
* Assuming indexes are 1 vs. 0-based.

oeps/oep-0041-arch-async-server-event-messaging.rst
ormsbee: @regisb: Just as a heads up: I'm going to remove all mention of the Transport layer from this OEP. Message queue implementation has bubbled up the list of priorities and is actively being worked on this quarter. As part of that, @feanil is leading an investigation/evaluation of transport backends, which will include Redis Streams, Kafka, and Pulsar.

ormsbee:
@feanil, @christopappas, @cpennington, @regisb, @bryanlandia: I've updated this to address earlier comments about `source`, as well as some edits and filling in on the motivation side. I've also removed transport details and tried to focus this OEP more narrowly on the message format itself, with additional transport related guidance expected to come from ARCH-BOM in the coming weeks.

oeps/oep-0041-arch-async-server-event-messaging.rst
regisb: Awesome, thanks for the heads-up @ormsbee. I'm curious to see the results of @feanil's research.

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,463 @@
...
+  {
+    "id": "b3b3981e-7544-11ea-b663-acde48001122",
+    "source": "https://discovery.edx.org/events"
+    "specversion": "1.0",
+    "time": "2020-02-23T09:00:00Z",
+    "type": "org.openedx.catalog.course.created.v1"
+    "data": {
+        // message specific payload here
+    }
+  }
```
ormsbee, line 130: Self note: update example to make sure it's consistent with more recent changes (e.g. `source`)

oeps/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -0,0 +1,463 @@
...
+ecommerce service is its own subdomain and should not be emitting ``catalog`` or
+``learning`` events. It is sometimes the case that a subdomain encompasses
+multiple services (e.g. Studio and Blockstore both operate on the
+``content_authoring`` subdomain).
+
+
+One Producer Service Per Event Type
+-----------------------------------
+
+Each event type should be emitted by one, and only one, service. That service is
```
jhan217, line 356: How will people be able to figure out if a given event name is already in use? Should events be namespaced to the owning service/sub-domains?

ormsbee: @feanil had some thoughts around having either a centralized schema registry repo, or else having a separate one per business subdomain (e.g. "catalog", "learning"). That's all very fuzzy to me at this point.

ormsbee: I'll put a small note in here to this effect before merging this tonight. Thank you!

jhan217: Yeah just seems like if we want events to be namespaced and have only one owner, we should either figure out a way to programmatically enforce it or at least give people a way to see if the namespace they want for their event is already taken.

PR #140, opened 2020-03-20
OEP-44: Changelogs

robrap:
ARCHBOM-1075


oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+Many repositories do not have a CHANGELOG.rst, and would require one to comply with this OEP. In most cases, this would simply be a boiler-plate CHANGELOG.rst that points to the GitHub Releases url for that repository.
+
+In some cases, like `edx/configuration CHANGELOG`_, adding dates would add clarity to the order of the log and would stop people from adding to the bottom, which sometimes happens. This file should also be changed to rST format, in keeping with OEP-19.
+
+.. _edx/configuration CHANGELOG: https://github.com/edx/configuration/blob/master/CHANGELOG.md
+
+Rejected Alternatives
+=====================
+
+An alternative for Python Library Respositories, rather than using GitHub Releases, would be to add changelog entries directly into the CHANGELOG.rst. Given that `Frontend Libraries`_ are already using GitHub Releases, and many Python libraries already use the GitHub Releases, it seems simpler to be consistent.
```
robrap: **Note**: We currently keep [this Confluence page of Python Package Changelogs](https://openedx.atlassian.net/wiki/spaces/TE/pages/1001521320/Python+Package+Changelogs) for helping with upgrades.

Although each line points to either a CHANGELOG.rst or GitHub releases, I have found some circumstances where both are used, and they are not in sync (e.g. GitHub releases has the latest descriptions and the CHANGELOG.rst is out of date).

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+* Format dates as YYYY-MM-DD. As an example, 2020-03-21.
+* If you do nothing else, list deprecations, removals, and any breaking changes in your changelog.
+* It is both acceptable and appropriate to clean up changelog entries at any time to add clarifications or customize messaging for the audience of the changelog when needed.
+
+Example Changelog
+~~~~~~~~~~~~~~~~~
+
+Here is an example CHANGELOG.rst.
+
+.. note:: Although this example includes versions, based on the decisions below our changelog entries would typically only include dates, because our versioned changelogs are managed using GitHub Releases.
```
nedbat: I thought the recommendations here were to either have explicit text in changelog.rst, or a link to the GitHub releases page?  But this sentence seems to anticipate a changelog file that has dates, but no version numbers?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+  .. _OEP-44: https://open-edx-proposals.readthedocs.io/en/latest/oep-0044-bp-changelogs.html
+  .. _Semantic Versioning: https://semver.org/spec/v2.0.0.html
+
+  [2.0.0] - 2020-03-21
+  --------------------
+
+  Removed
+  ~~~~~~~
+
+  * Removed deprecated function `xxx`. See DEPR-XXX for details.
```
nedbat: Could we expect more detail than this? Often deprecation announcements include mention of the thing to use instead, for example. It would be great for changelog entries to include more information that users will need.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+  * Feature x. See announcement here: https://xxx
+
+Notable Differences
+~~~~~~~~~~~~~~~~~~~
+
+The following are notable differences between the recommendations of this OEP and the `keepachangelog.com`_ recommendations:
+
+* We use GitHub Releases in certain cases, as you'll see below, for developer efficiency purposes.
+* Because we typically release every change for libraries, we don't get any benefit from an "Unreleased" section.
```
nedbat: We don't seem to mention Unreleased except in this sentence. Perhaps we should make it explicitly optional, with reasons repo owners might choose to include it or not?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+All versioned code should use a MAJOR.MINOR.PATCH format for versioning, as defined by the `Semantic Versioning v2`_ specification. Here is a copy of the most relevant part of the summary:
+
+Given a version number MAJOR.MINOR.PATCH, increment the:
+
+* *MAJOR* version when you make *incompatible API changes*,
+* *MINOR* version when you *add functionality in a backwards compatible manner*, and
+* *PATCH* version when you make *backwards compatible bug fixes*.
+
+Every published version should have a corresponding entry in the changelog, but not necessarily in CHANGELOG.rst.
```
nedbat: I don't know what this means.  How are "the changelog" and "CHANGELOG.rst" different?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+One of the most important rules when choosing whether to increment the MAJOR, MINOR, or PATCH version, is that backward incompatible changes always result in a new MAJOR version, and never a MINOR or PATCH version.
+
+* You must increment the MAJOR version for backward incompatible changes, even if you think it is unlikely to affect anyone. You can simply note in the changelog entry why it is backward incompatible and why you think it is unlikely to affect anyone.
+* If a backward incompatible change is found in a MINOR or PATCH version after publishing, first update the changelog and communicate immediately.  Next, a new MINOR or PATCH version should immediately be released which reverts this change, and it should be moved to a new MAJOR version.
```
nedbat: I wonder if versioning logic like this belongs in this OEP.  It could become a more difficult part to get agreement on.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+* If you decide that reverting will cause more harm for some reason, make sure you clearly communicate this decision.
+
+The reason it is important that only MAJOR versions include backward incompatible change is because when people upgrade dependencies, some will only read changelogs for MAJOR versions because of the potential backward incompatibilities, and may not read notes for MINOR or PATCH upgrades.
+
+Frontend Libraries
+~~~~~~~~~~~~~~~~~~
+
+Like all Open edX repositories, frontend libraries should also include a CHANGELOG.rst. However, as you'll see below, this file may simply point to the GitHub Releases. Here is an example CHANGELOG.rst::
```
nedbat: Why is this true particularly for frontend libraries?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+  All notable changes to this project will be documented as `GitHub Releases`_.
+
+  This project adheres to `Semantic Versioning`_.
+
+  .. _GitHub Releases: https://github.com/edx/frontend-build/releases
+  .. _Semantic Versioning: https://semver.org/spec/v2.0.0.html
+
+Many Python libraries publish packages to PyPI when a version is tagged. Publishing a GitHub Release both tags the version and allows one to write an appropriate changelog message. This is the recommended process at this time.
+
+Every published release should have a corresponding changelog entry.
```
nedbat: Do we have to clarify what "published" means? Or is "release" enough here?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+* The changelog entries will live in the CHANGELOG.rst file, following the recommendations detailed in `Changelogs and Changelog Entries`_.
+* Changelog entries should start with a date, rather than a date and version, because these repositories are not versioned.
+
+Changelog for edx-platform
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The edx-platform repository should have a CHANGELOG.rst, just like all other repositories.
+
+* Many notes that could potentially live in the edx-platform CHANGELOG.rst get added to the `Open Edx Release Planning`_ pages which have historically been created for each named release. Worst case, edx-platform notes are manually added to the CHANGELOG.rst instead, and at least the documentation stays more closely tied to the code.
```
nedbat: I don't know what "Worst case" is clarifying?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+* Entries could be grouped by day if they could be added at a faster cadence (see `Tooling and Automation`_). Otherwise, date ranges should be used.
+
+* Attempting to add changelog entries at a very faster cadence without automation is likely to cause too many merge conflict issues. However, this should not stop people from adding entries immediately so we can learn.
```
nedbat: Is there anything we can do to help with merge conflicts? For the original edx-platform changelog, I had recommended not putting your entry at the very top of the file.  (Some developers then wanted to know what "near the top" meant, but maybe we've grown since then :)  )

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+* As with all changelogs, it is acceptable and appropriate to clean up or enhance or add changelog entries as necessary.
+
+* More work is clearly required to determine how to iterate and improve the process for a CHANGELOG.rst in edx-platform.
+
+* The CHANGELOG.rst will also likely need to be broken into multiple files, but that can be a local decision.
```
nedbat: Why would it need to be broken into multiple files? Also, "local" decision? This is in the edx-platform section.

oeps/oep-0044-bp-changelogs.rst
nedbat: I thought one of the suggestions in the OEP was to keep an explicit CHANGELOG.rst.  Why is it mentioned here as a rejected alternative?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+Why keep a changelog?
+~~~~~~~~~~~~~~~~~~~~~
+
+To make it easier for users and contributors to see precisely what notable changes have been made between each release (or version) of the project.
+
+Who needs a changelog?
+~~~~~~~~~~~~~~~~~~~~~~
+
+People do. Whether consumers or developers, the end users of software are human beings who care about what's in the software. When the software changes, people want to know why and how.
```
nedbat: Is there some way to measure success? It would be easy to count the number of CHANGELOG.rst files, but are the changelogs good? Are they accomplishing the goal of adequately educating the readers? What can we do to assess our progress toward good changelogs?

nedbat: Could this OEP have a rubric for the kinds of information that should be included in a changelog entry?

oeps/oep-0044-bp-changelogs.rst
robrap: This was specifically for Python Library Repositories.  I tried making this more clear, but you can let me know if it is. :)

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,294 @@
...
+
+These make great section headers, but similar language could also be used in messages to clarify the type of change.  For commits, you may want to use "adds" in place of "added", etc.
+
+The recommendation also includes "**Security** in case of vulnerabilities", but that is being called out separately because the Security Group needs to manage the timing of adding any such designation to any public changelog.
+
+Other Recommendations
+~~~~~~~~~~~~~~~~~~~~~
+
+* Format dates as YYYY-MM-DD. As an example, 2020-03-21.
+* If you do nothing else, list deprecations, removals, and any breaking changes in your changelog.
```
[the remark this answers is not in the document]
robrap: I updated the text. See if you like it better.

oeps/oep-0044-bp-changelogs.rst
robrap: I removed the versions from the example and updated the intro line, so hopefully this is more clear.

oeps/oep-0044-bp-changelogs.rst
robrap: I added more optional details.  See if you think this helps.

oeps/oep-0044-bp-changelogs.rst
robrap: I simply deleted this line.

oeps/oep-0044-bp-changelogs.rst
robrap: I made this more clear.

oeps/oep-0044-bp-changelogs.rst
robrap: By "versioning logic", do you mean these process steps?  I updated the language to make it less about what you must do, and more as help for what you might do.

Or, do you mean the entire topic of semantic versioning?  I don't expect a lot of disagreement there, but I am happy to move this to a separate OEP if you think that makes more sense.

oeps/oep-0044-bp-changelogs.rst
robrap: I made a one word change.  Given that, I'm not sure what you mean by "this".  I'm not seeing where this isn't clear, so I need more information.

oeps/oep-0044-bp-changelogs.rst
robrap: I updated the language.  Let me know if it is not clear.

oeps/oep-0044-bp-changelogs.rst
robrap: I took two steps here:
1. More importantly, I deleted this whole section which was both speculative and really does belong in an edx-platform ADR when and if any of this would provide help.
2. I updated the language around "worst case".  I am going to copy the text to this comment in case I ever want to see it again, but I think this was mostly about me getting stuff out of my head.

```
Changelog for edx-platform
~~~~~~~~~~~~~~~~~~~~~~~~~~

The edx-platform repository should have a CHANGELOG.rst, just like all other repositories.

* Many notes that could potentially live in the edx-platform CHANGELOG.rst get added to the `Open Edx Release Planning`_ pages which have historically been created for each named release. Instead, adding entries to the edx-platform CHANGELOG.rst would follow the best practice of keeping documentation close to its code.

* Entries could be grouped by day if they could be added at a faster cadence (see `Tooling and Automation`_). Otherwise, date ranges should be used.

* Attempting to add changelog entries at a very faster cadence without automation is likely to cause too many merge conflict issues. However, this should not stop people from adding entries immediately so we can learn.

* As with all changelogs, it is acceptable and appropriate to clean up or enhance or add changelog entries as necessary.

* More work is clearly required to determine how to iterate and improve the process for a CHANGELOG.rst in edx-platform.

* The CHANGELOG.rst will also likely need to be broken into multiple files, but that can be a local decision.

.. _Open Edx Release Planning: https://openedx.atlassian.net/wiki/spaces/COMM/pages/13205845/Open%2BedX%2BRelease%2BPlanning
```

oeps/oep-0044-bp-changelogs.rst
robrap: I don't think I've grown.  Do you mean if there are 5 bullet points for the current date's entry, you should insert your new bullet somewhere randomly between the first and last?  Or do you mean something different?

Either way, I pulled this code, but I am curious.

oeps/oep-0044-bp-changelogs.rst
robrap: I pulled this section so it really could live in an ADR when and if it is actually necessary.  Maybe it never will be.  This was about whether the changelog gets too long and we want to break it by year or anything else.

oeps/oep-0044-bp-changelogs.rst
robrap: 1. I'm sure there are ways to measure success, but I don't see how to do it without either automation or surveying the audiences of changelogs, and this hasn't been a priority. My first priority was to provide clarity so someone could simply know if they were doing or not doing what was expected.  I imagine we might be able to add to the new Repository Health Dashboard some metric related to the number of entries vs the number of merges, etc., but I don't want to block this PR on this.  I could add an "Open Questions" or "Next Steps" section that acknowledges this direction we'd like to go in.
2. I'd love help adding a rubric if you have ideas.  I think what I added gives a strong indication of the types of changes to consider, but happy to provide something else or in addition.

regisb: I think a good measure for success would be for Open edX newcomers to be able to access, read, understand and link to the changelog. I understand this is not an objective measure, but after all we are building something for humans here.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,285 @@
...
+
+* A new MINOR or PATCH version can be immediately released to revert the change, and the change can be restored with a new MAJOR version. Or,
+* If you decide that reverting will cause more harm for some reason, and you do not intend to revert, please let this be a part of your communication.
+
+The reason it is important that only MAJOR versions include backward incompatible change is because when people upgrade dependencies, some will only read changelogs for MAJOR versions because of the potential backward incompatibilities, and may not read notes for MINOR or PATCH upgrades.
+
+Frontend Libraries
+~~~~~~~~~~~~~~~~~~
+
+Like all Open edX repositories, frontend libraries should also include a CHANGELOG.rst. However, as you'll see below, this file will simply point to the GitHub Releases. Here is an example CHANGELOG.rst::
```
regisb: > However, as you'll see below, this file will simply point to the GitHub Releases.

:disappointed: 

This is really too bad. It's great to be able to find all changes in a single file, to be easily read by humans, something that is not allowed by semantic-releases. Also, semantic-releases picks changes from git commits, which directly contradicts the principle according to which changelogs are made for humans, not machines. A changelog entry should be allowed to be longer than 50 characters, which is the standard git commit title maximum length.

Any chance we can improve on that?

oeps/oep-0044-bp-changelogs.rst
nedbat: I guess this still doesn't feel like a rejected alternative to me.  A Python library might decide to keep an explicit CHANGELOG, and not use GitHub releases.  Why would that be bad?  You are right that it is bad to have both an explicit CHANGELOG, and releases with notes, since they can be out of sync.  Is that the rejected alternative?

oeps/oep-0044-bp-changelogs.rst
nedbat: Great thanks.  It could also be part of the bullet the first mentions the deprecation.

oeps/oep-0044-bp-changelogs.rst
nedbat: We can wait to see what further review brings. The mechanics of semantic versioning feels orthogonal to how changelogs should be managed, but I can see that the topics are related.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,285 @@
...
+  =========
+
+  All notable changes to this project will be documented as `GitHub Releases`_.
+
+  This project adheres to `Semantic Versioning`_.
+
+  .. _GitHub Releases: https://github.com/edx/frontend-build/releases
+  .. _Semantic Versioning: https://semver.org/spec/v2.0.0.html
+
+Many Python libraries publish packages to PyPI when a version is tagged. Publishing a GitHub Release both tags the version and allows one to write an appropriate changelog message. This is the recommended process at this time.
```
nedbat: Do we have a recommendation for how to track the work-in-progress until the release, or are we assuming for a Python library that every change will result in publishing a new version?

oeps/oep-0044-bp-changelogs.rst
nedbat: Yes, I meant just choose a random spot that would still be appropriate.  At the time, we didn't demarcate the changelog by date (it was like edx-configuration's is now).

oeps/oep-0044-bp-changelogs.rst
nedbat: My question was: why particularly will frontend libraries use a CHANGELOG that simply points to GitHub Releases?  If this is a good solution for them, why isn't it the right choice for other kinds of repos?

oeps/oep-0044-bp-changelogs.rst
nedbat: For measuring success, we could self-audit.  Three months after ratifying the OEP, take a survey of repos.  How many claim to conform?  Of those, read the changelogs: do we think they have the right information?

robrap:
Thanks for comments. FYI: The review period will start in the next couple of weeks, sometime after Django 2.2. is out, so I may not be responding for a bit.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,285 @@
...
+
+As noted above, please update the changelogs as necessary to add clarifications or customize the messaging for the audience of the changelog, especially because these messages originally come from commit messages, which may have been written for a different audience.
+
+.. _semantic-release: https://github.com/semantic-release/semantic-release
+.. _GitHub Release changelog for edx/frontend-build: https://github.com/edx/frontend-build/releases
+
+Python Libraries
+~~~~~~~~~~~~~~~~
+
+Like all Open edX repositories, Python libraries should also include a CHANGELOG.rst. This file may also simply point to the GitHub Releases. Here is an example CHANGELOG.rst::
```
nedbat, line 216: Without extra tooling GitHub Releases are *not* a good changelog solution.  They provide no help for writing the changelog.

robrap: Can you clarify what you mean?  Here are the Github releases for edx-rest-api-client, as an example:  https://github.com/edx/edx-rest-api-client/releases.  When you publish the release notes, the library is published to PyPI.  Is that "extra tooling"?  If not, can you clarify what "extra tooling" means to you, and why this current process is "_not_ a good changelog solution."  Note: for me as the audience of these changelog entries, they also have always been sufficient for my needs during upgrades.

nedbat: In my testing, the release shows only the latest commit by default.  You get a blank text field to write into if you want, but do we think people will write new text there? I thought we were trying to reduce the duplicated work.  If people are willing to write separately about the changes, why not do it in a CHANGELOG file?

If there's more than one change in a release, GitHub releases doesn't seem to help to summarize them all.  Are we assuming that every commit will become a release?

nedbat: By "extra tooling" I meant something like the "semantic-releases" tool the frontend libraries are using, which populates the release description with commit messages.

oeps/oep-0044-bp-changelogs.rst
robrap: I will remove this and just add a note about not allowing notes to be out-of-sync.

oeps/oep-0044-bp-changelogs.rst
robrap: I will add another bullet on another day.  Not sure if this is what you are looking for.

oeps/oep-0044-bp-changelogs.rst
robrap: 1. As noted in several places in the OEP, initial changelog entries can and should be updated as-needed, so there are no limitations.
2. Regarding a single file, I won't stop anyone from finding or writing a tool that reads the GitHub Releases and produces a file. :)

oeps/oep-0044-bp-changelogs.rst
robrap: Mostly we publish every change.

oeps/oep-0044-bp-changelogs.rst
regisb: 1. To be blunt, I don't expect Open edX developers to come back and edit their commit messages. They simply don't have any incentive to do so. Case in point: https://github.com/edx/frontend-build/releases The release messages make no sense at all for those who are not already experts of this component.
2. If we publish a changelog from parsing the github releases, it will not be in sync with the git tags. Let me suggest something else: I'll write a tool that parses the changelog and will generate the github release message.

oeps/oep-0044-bp-changelogs.rst
regisb: This defeats the entire purpose of a changelog. If I want to read the git history of a repo I run `git log`. There are too many commits that do not belong in a changelog. A changelog is primarily for users of a tool.

robrap: What you are calling "too many commits" is simply a result of our current process.  For our libraries, we release a new version of the library as soon as we need a new feature or fix.  We do not slowly accumulate changes until we are ready for some grand release.  So, in practice, there is typically one PR per version/release.
I think we would both agree that users of the library need to know what is contained in each version. So, maybe you are asking for a different process with fewer versions?

oeps/oep-0044-bp-changelogs.rst
robrap: 1a: When you say "edit their commit messages", do you mean edit their changelog messages?
1b: I don't know if you are asking for different changelog messages, or fewer versions.  Assuming we agree that there would be a changelog for each version, maybe you could provide example changelog messages for the recent versions of frontend-build that you would be happy with?  Or, if you simply want fewer versions, how do you propose we adjust our process to make that happen?
2. I am all for new tools that help improve all this.  Just keep in mind that the current tool is maintained, and any new tools would need a maintainer.  Also, the tool would need to perform additional tasks, including keeping the two in sync over time if one changes, and the process for publishing to npm would change, etc.  This is all possible, but this is certainly a larger project.

regisb: > 1a: When you say "edit their commit messages", do you mean edit their changelog messages?

Yes. If I understand correctly, the "semantic-release" utility uses the commit title as a changelog entry, so in the proposed approach commit titles are changelog messages, and vice versa.

> 1b: I don't know if you are asking for different changelog messages, or fewer versions. Assuming we agree that there would be a changelog for each version, maybe you could provide example changelog messages for the recent versions of frontend-build that you would be happy with? Or, if you simply want fewer versions, how do you propose we adjust our process to make that happen?

I'm perfectly happy with many new versions a day. But the changelog messages should be targeted towards users. As such, changelog entries such as "fix tests" or "internal refactor of FooBar class" are totally useless to me. When there are only such changes in a release, I'd be happy to read "no functional change" for a release. When there are indeed changes, they could be phrased as such:

- "Fix 500 error during logout" (and not: "Fix SyntaxError in UserLogout class")
- "Extend the public API with a `set_foobarizer` method: this is particularly useful for application developers trying to foobarizing their users. See [the docs](http://thedocs.com/set_foobarizer). (and not: "Add `set_foobarizer` method to `api.Foo`)
- [Breaking change] Get rid of the `dry_run` parameter in the public `foobarize` API method. This parameter is deprecated in favour of the `no_apply` parameter. See [the docs](http://thedocs.com/no_apply). ("Rename `dry_run` parameter)

2. Hey, you are the one who suggested we create a new tool ;) There is no need to keep a CHANGELOG.md file and the Github releases in sync. Having a changelog would already be a major improvement over the current process. Older changelog entries are typically not updated very often. Also, the Changelog is versioned, so it's usually regarded as a better source of authority than to Github releases. So I think it's ok if we sometimes make changes to older entries of a changelog file and not update the corresponding releases.

oeps/oep-0044-bp-changelogs.rst
regisb: I don't think there are too many commits in the repository, and I think it's great that we release a new version for every fix and feature. See my comment above for how I think new versions should be described.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,246 @@
...
+
+  .. _GitHub Releases: https://github.com/edx/frontend-build/releases
+  .. _`OEP-44: Changelogs`: https://open-edx-proposals.readthedocs.io/en/latest/oep-0044-bp-changelogs.html
+  .. _Semantic Versioning: https://semver.org/spec/v2.0.0.html
+
+Every package published to PyPI should have a corresponding changelog entry for the new version.
+
+Many Python libraries publish packages to PyPI when a version is tagged. Publishing a GitHub Release both tags the version and allows one to write an appropriate changelog message. This is the recommended process at this time to avoid issues where the CHANGELOG.rst and GitHub Releases are not in sync.
+
+For a given Python Library repository, if a decision is made to add changelog entries to the CHANGELOG.rst, you **must** provide a way to ensure the CHANGELOG.rst and GitHub Releases are not out of sync. At this time, we do not have tooling for this.
```
nedbat, line 197: If we are manually managing a CHANGELOG.rst file, why do we need GitHub releases at all?  Do you mean the git tags?

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,246 @@
...
+  The release notes format should include information recommended in `OEP-44: Changelogs`_.
+  This project adheres to `Semantic Versioning`_.
+
+  .. _GitHub Releases: https://github.com/edx/frontend-build/releases
+  .. _`OEP-44: Changelogs`: https://open-edx-proposals.readthedocs.io/en/latest/oep-0044-bp-changelogs.html
+  .. _Semantic Versioning: https://semver.org/spec/v2.0.0.html
+
+Every package published to PyPI should have a corresponding changelog entry for the new version.
+
+Many Python libraries publish packages to PyPI when a version is tagged. Publishing a GitHub Release both tags the version and allows one to write an appropriate changelog message. This is the recommended process at this time to avoid issues where the CHANGELOG.rst and GitHub Releases are not in sync.
```
nedbat, line 195: Without external tooling, GitHub Releases are painful to use.  They are pre-populated with only the last commit message on the branch.  They are an extra editing step after the tag has been created, with no possibility for review.

I would think to reduce developer effort, and avoid losing information, the best recommendation would be to manually edit a CHANGELOG.rst file.

oeps/oep-0044-bp-changelogs.rst
robrap: @regisb @nedbat: Lots of great points that have me thinking about a bunch of potential updates.  I will put up changes in the next day or so, or ask more questions.  Thanks.

robrap: @regisb:
1. You all changed my tune and I am recommending against Github Releases.
2. I think the frontend library used can write to a changelog instead, but someone needs to trial that.
3. I lifted a bunch of your examples and added them to the OEP.  Happy to add you as co-author if you want.  That's up to you.
Please mark as resolved when and if you are satisfied with this thread.  Thanks.

oeps/oep-0044-bp-changelogs.rst
robrap: I haven't found it painful to use, but started recommending against it and provided some reasons.  I'm on board.

oeps/oep-0044-bp-changelogs.rst
robrap: We don't.  I simplified the OEP a lot.  Marking as resolved.

regisb:
I really appreciate your new version @robrap, thanks a bunch for your work. No need to add me as an author -- I'll review this as soon as possible.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,212 @@
...
+.. list-table::
+
+   * - OEP
+     - :doc:`OEP-0044 </oeps/oep-0044>`
+   * - Title
+     - Changelogs
+   * - Last Modified
+     - 2020-04-30
+   * - Authors
+     - Robert Raposa
```
davidjoy, line 14: Would be useful to include an email address here.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,212 @@
...
+============
+
+Since these are only recommendations, strictly speaking, there are no consequences.
+
+If you want to follow the best practice, update your repository to follow this guidance.
+
+Rejected Alternatives
+=====================
+
+For repositories without versioned code, like edx-platform and other IDAs, the only real alternatives to a CHANGELOG.rst are:
```
regisb, line 192: What would it take to add a CHANGELOG.rst to edx-platform? I understand that we can't count on version numbers, and release names are too far apart to simply dump changes in a single file. But I believe @nedbat had in mind some tooling to facilitate the addition of changelog entries. Maybe these changes could be aggregated every now and then (every week? month?) in a single changelog file for other users to keep track of?

oeps/oep-0044-bp-changelogs.rst
regisb: I can't mark this as resolved but yes, I'm very satisfied.

oeps/oep-0044-bp-changelogs.rst
regisb: :+1:

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,212 @@
...
+   * - Rename dry_run parameter
+     - **BREAKING CHANGE** Remove the dry_run parameter in the public foobarize API method. This parameter is deprecated in favour of the no_apply parameter. See docs for details.
+   * - Add set_foobarizer method to api.Foo
+     - Add a set_foobarizer method to Foo's public API. This is particularly useful for developers trying to foobarize their users. See docs for details.
+
+Github Releases
+~~~~~~~~~~~~~~~
+
+* Github Releases should not replace the CHANGELOG.rst.
+* If used, Github Releases should stay in sync with the CHANGELOG.rst.
```
robrap, line 164: FYI: Github adds "Latest Release" next to the latest Github Release.  So, if newer tags just go to CHANGELOG.rst, it is very misleading.  Possible solutions:
1. Transfer all release information to CHANGELOG.rst and delete all Github Releases.  Note: hopefully no one adds a new Github Release again.  Is automation required to ensure that releases don't exist?
2. Publish a release with title "See CHANGELOG.rst" and body "See CHANGELOG.rst" for every new release.  Annoying and ugly, but works.

oeps/oep-0044-bp-changelogs.rst
```diff
@@ -0,0 +1,212 @@
...
+* **Who needs a changelog?**: People do. Whether consumers or developers, the end users of software are human beings who care about what's in the software. When the software changes, people want to know why and how.
+
+Recommendations
+===============
+
+All Open edX repositories should contain a CHANGELOG.rst.
+
+The recommendations for formatting changelogs have been adapted from `keepachangelog.com`_.
+
+Formatting Changelogs
```
robrap, line 66: Consider relying more heavily on keepachangelog.com default recommendations and simplifying, like in https://github.com/edx/edx-drf-extensions/commit/f5f7b50dac1c8116c8dccc1d9809a9ee568d14e7

robrap:
I am closing this PR for now.  It can get reopened when https://openedx.atlassian.net/browse/ARCHBOM-1076 gets picked up to continue this effort.

PR #143, opened 2020-04-08
OEP-45 Operating and Configuring Open edX

bderusha:
This OEP is under review from **2020-04-17 though 2020-05-04.**

The goal is to vet the proposed end state before we start talking about how to get there to make sure that the proposal properly outlines and answers major questions about how to operate and configure Open edX in this future state. 

We are keeping this review period relatively short because, based on early feedback, we don't expect it to be a particularly controversial goal.  Agreeing on this OEP quickly will allow us to move the conversation over to implementation strategies where we expect there to be more diverse needs to take into consideration.  If it turns out to be more controversial than anticipated we will extend the review period to accommodate more feedback.

 The implementation discussion will happen in [this follow-on PR](https://github.com/edx/open-edx-proposals/pull/144). While comments can be made at any time, I would ask that people refrain from any deep discussions there about how to achieve OEP-45 until we agree on what OEP-45 should be.

### DIFF
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -114,11 +114,11 @@ Additionally it is not clear which settings are required to be overridden and wh
 
 *  ``__init__.py`` - Sourcing our config from this file within the settings directory takes advantage of django defaults and means that settings will be picked up automatically without needing to specify ``--settings`` anywhere. This entry point would import ``required.py``, ``defaults.py``, and the code to override both from a config file.
 *  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default, e.g. LMS_BASE_URL which will be different per environment.
-*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+*  ``defaults.py`` - other settings which will have production-ready defaults
 
 The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
 
-``required.py`` variables will all be initialized to ``None`` and the application will not start unless they are set. This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
+``required.py`` variables must be overridden by operators.  The application will check that operators provided these values not start unless they are set. This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
 
 This new settings structure obviates the need for any other python files in the settings directory (such as ``devstack.py``, ``test.py``, etc). The values currently set in those files should be moved to a corresponding ``devstack.yml``, ``test.yml``, etc in the same settings directory. This gives developers and operators more consistency across environments since the same code paths are being executed with different values.
 
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -11,11 +11,11 @@ OEP-45: Configuring and Operating Open edX
    * - Last Modified
      - 2020-04-06
    * - Authors
-     - Bill DeRusha
+     - Bill DeRusha <bill@edx.org>
    * - Arbiter
-     - Felipe Montoya 
+     - Felipe Montoya <felipe.montoya@edunext.co>
    * - Status
-     - Under Review
+     - Provisional
    * - Type
      - Architecture
    * - Created
@@ -31,7 +31,7 @@ OEP-45: Configuring and Operating Open edX
 Abstract
 ========
 
-Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX independently deployable application (IDA) rather than providing Ansible playbooks.
+Wherever possible, the edX organization will provide and manage Docker images as the medium for packaging operating dependencies of an Open edX independently deployable application (IDA) rather than providing Ansible playbooks.
 
 The configuration for running IDAs in different environments will be done through a config file per unique [application, environment] combination. How those files are generated and managed is the responsibility of each operator.
 
@@ -43,6 +43,11 @@ Wherever possible, operating concerns will live within the codebase they are con
 * Documented settings files distinguishing required values without defaults (like secrets) from other settings which will defaults reasonable for most production installations.
 * An operations manual documenting standard operations required to run and maintain the IDA.
 * A changelog to document differences in settings, operating concerns, and application functionality from release to release.
+
+As code is moved to this new paradigm it is important to note that we strive to maintain or improve upon the state of operations without sacrificing features or visibility.  If operations code is removed from one place it must either have a home in some other public place better aligned with this OEP (EG package installation in Dockerfiles) or no longer be needed at all in this model (EG `conditional ansible commands`_ for installing packages on particular operating
+systems).
+
+.. _conditional ansible commands: https://github.com/edx/configuration/blob/ada15bec6e0269042380e340e65675640983e6ba/playbooks/roles/newrelic_infrastructure/tasks/main.yml#L44-L52
 
 Context
 =======
@@ -94,6 +99,8 @@ In order to function as documentation for operators Dockerfiles will be well-com
 
 edX will provide Docker images for IDAs that captures the latest code on the master branch as well as images representing named releases. edX will not provide these images for named releases prior to the acceptance and implementation of this OEP (Aspen through and including Juniper at time of writing).
 
+Operators will be able to use these provided images as a base for any private or custom images they need to build for their environments.
+
 
 Configuration
 *************
@@ -131,7 +138,7 @@ Since defaults are provided by the IDA, many smaller deployments should not need
  
 **Config file generation & management**
 
-Due to the varied needs and processes of different operators, how the config files are created, managed, or otherwise end up on the server is up to the operator and will depend greatly on their deployment strategy.
+Due to the varied needs and processes of different operators, how the config files are created, managed, or otherwise end up on the server is up to the operator and will depend greatly on their deployment strategy. With a consistent method for configuring IDAs it will be reasonable to have tooling to assist with migrating between releases, but the implementation of such tooling is outside the scope of this proposal.
  
 **Documentation of settings**
 
@@ -143,11 +150,23 @@ The settings found in both the ``required.py`` and ``defaults.py`` files will be
 Operations Manuals
 ******************
 
-A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that IDA. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the IDA for web traffic or as an async worker and how to manage the IDA's underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``docs`` directory (as per `OEP-19`_) for that IDA. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the IDA for web traffic or as an async worker and how to manage the IDA's underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
  
 In the same vein as not dictating how operators create and manage their IDA config files, operators will also be expected to manage how they execute the operations documented in the manual.
 
+.. _OEP-19: https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html
 .. _this commit: https://github.com/openedx-btr-wg/edx-platform/commit/18effd83f983f497ca0a1535108fa41dc50d06a2#diff-ca02329742db0a77612a18ba1260d178R1-R39
+
+
+Industry Best Practices
+***********************
+
+The recommendations above are heavily inspired by the following resources:
+
+* https://12factor.net/
+* https://code.djangoproject.com/wiki/SplitSettings
+* https://github.com/openedx-btr-wg/edx-platform/tree/btr-wg/reference-manual
+* https://github.com/openfun/richie/blob/master/src/richie/apps/core/defaults.py
 
 
 Alternatives Considered
@@ -165,6 +184,11 @@ Jumping to Kubernetes
 
 Kubernetes is an open source container orchestration platform pioneered by Google. While it often occupies the same conversation space as containers because it is a powerful way to manage them, it is a huge increase in complexity and expertise required to operate. For most installations Kubernetes is currently too much overhead/learning curve for the value. The edX organization may opt to explore deploying Docker containers this way in the future and would love to collaborate with operators who also decide to use Kubernetes to compare notes.
 
+
+Configuration via Environment Variables
+***************************************
+
+There are many django project which configure their applications by grabbing the settings value from an environment variable otherwise using a default.  While this is technically feasible, the platform relies on setting many complex data structures (lists & dicts) and to do so using ENV VARS would be quite challenging to manage and thus was declined as an option to pursue.
 
 
 Implementation Strategy
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -1,0 +1,152 @@
+##########################################
+OEP-45: Configuring and Operating Open edX 
+##########################################
+
+.. list-table::
+
+   * - OEP
+     - :doc:`OEP-0045 </oeps/oep-0045>`
+   * - Title
+     - Configuring and Operating Open edX
+   * - Last Modified
+     - 2020-04-06
+   * - Authors
+     - Bill DeRusha
+   * - Arbiter
+     - ???
+   * - Status
+     - Draft
+   * - Type
+     - Architecture
+   * - Created
+     - 2020-04-06
+   * - Review Period
+     - ???
+
+.. contents::
+   :local:
+   :depth: 3
+
+
+Abstract
+========
+
+Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination.  How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented.  How those procedures are executed is the responsibility of each operator.
+
+Wherever possible, operating concerns will live within the codebase they are concerned with.  Thus it should be expected that all applications will provide the following:
+
+* A Dockerfile from which the container image can be built.
+* A documented settings file with production-ready defaults.
+* A documented settings file with all of the configuration values that must be defined by the operator.
+* An operations manual documenting standard operations required to run and maintain the application.
+
+Context
+=======
+
+Back when edx-platform was the only code running, the Ansible playbooks in the configuration repo were manageable, though still not necessarily easy to understand.  As edX has invested in microservices, the configuration repo has grown to support the configuration of all of them.  As patterns emerged shared concerns were pulled out into unversioned shared roles making it hard to change without impacting all apps or wiring configurable toggles throughout much of the shared space.
+
+The end result is that the configuration repo has grown into a sprawling codebase that attempts to be everything for everyone trying to configure servers to run Open edX installations. There are a wide variety of roles, plays, and utility scripts with branching paths within each to support being executed against different base operating systems and many various configurable operational needs.  None of this is helped by the fact that edX.org operates at a scale that most installations do not.  With a mix of edX.org specific code, branching paths that are not regularly tested, and the high amount of required understanding around how all the pieces interact to change any of it, this approach to operating is a clear source of friction for developing and running the platform. Nothing proves this point more than the fact that the edx-platform README now explicitly warns users to leave it to the professionals.
+
+    Installing and running an Open edX instance is not simple. We strongly recommend that you use a service provider to run the software for you. 
+
+The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace.  In brief it is:
+
+* Use Dockerfiles to capture system dependencies and operate the resulting images in all environments.
+* Simplify and standardize how to configure an application.
+* Document standard operations.
+
+ 
+Containers
+**********
+
+**Why Containers**
+
+For managing microservice architectures containers are an industry standard for the following reasons: 
+
+* **Increased portability.**  Run on any host OS that supports containers.
+* **Greater efficiency.**  Ability to safely run multiple containers on the same hardware allows cost savings through efficient use of resources.
+* **Improved security.** Containers improve security by isolating applications from the host system and from each other.
+* **Improved application development.**  Allows developers to run a more production-like environment locally, preventing discrepancies between environments.
+* **Speed.** Start, create, replicate or destroy containers in seconds. Increases velocity for development, deployment, and production operations.
+* **Operational simplicity/consistency.**  By isolating application processes from the host OS it is simpler to manage and maintain the host OS.  Homogeneous administration of heterogeneous components, reduces the range of skill sets required to operate environments.
+
+
+**Clear System Dependency Documentation**
+
+For operators who don’t want to run in Docker containers the Dockerfile acts as documentation for the system dependencies they will need to recreate.  Everything an operator needs to know about how to set up their environment is all in one place rather than distributed across multiple overlapping and overriding files.
+
+Configuration
+*************
+ 
+**Django settings**
+
+Having a single artifact that runs with different configurations increases stability by improving development parity with other deployment environments.  edX applications already support configuration overrides via a yaml file for production environments, but development and test environments tend to configure the application using different code paths via a settings/devstack.py or settings/test.py file.
+
+Additionally it is not clear which settings are required to be overridden and which settings have values that may technically work but are inappropriate for production systems.  To alleviate these issue edX django applications will adopt the following settings structure:
+
+.. code-block:: text
+
+  settings
+  ├── __init__.py
+  ├── required.py
+  └── defaults.py
+
+
+*  ``__init__.py`` - Sourcing our config from this file within the settings directory takes advantage of django defaults and means that settings will be picked up automatically without needing to specify ``--settings`` anywhere.  This entry point would import ``required.py``, ``defaults.py``, and the code to override both from a config file.
+*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default.  EG LMS_BASE_URL which will be different per environment.
+*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables will all be initialized to ``NONE`` and the application will not start unless they are set.  This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
+
+This new settings structure obviates the need for any other python files in the settings directory (such as ``devstack.py``, ``test.py``, etc).  The values currently set in those files should be moved to a corresponding ``devstack.yml``, ``test.yml``, etc in the same settings directory.  This gives developers and operators more consistency across environments since the same code paths are being executed with different values.
+
+
+**Config file**
+
+Applications will be configured by a yaml file containing all of the settings variable overrides specified by the operator (including both required settings and secrets as well as default value overrides).  The file is made known to the application by an environment variable, ``<APPNAME>_CFG_PATH``, with the path to the file.  Versions of this config yaml may be provided in the application repo for certain environments such as development and test.  However, for all other environments (EG production), the file will need be managed elsewhere.
+
+Since defaults are provided by the application, many smaller deployments should not need to do much more than provide the required settings to operate.  For development environments the config will likely change the defaults to more development appropriate values. EG debug settings, log levels, email settings etc.
+ 
+**Config file generation & management**
+
+Due to the varied needs and processes of different operators, how the config files are created, managed, or otherwise end up on the server is up to the operator and will depend greatly on their deployment strategy.
+ 
+**Documentation of settings**
+
+The settings found in both the ``required.py`` and ``defaults.py`` files will be documented to describe what they are and how they should be used.  The documentation will consist of `Sphinx autodoc`_ compatible comments before each setting.  For reference that is a ``“comment with special formatting (using a #: to start the comment instead of just #)”``.  This keeps documentation close to the code as it is being written, while allowing it to be surfaced in generated docs.
+ 
+
+.. _Sphinx autodoc: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
+
+Operations Manuals
+******************
+
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that application.  See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the application for web traffic or as an async worker and how to manage the applications underlying database schema.  It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches.  Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
+ 
+In the same vein as not dictating how operators create and manage their application config files, so too will operators be expected to manage how they execute the operations documented in the manual.
+
+.. _this commit: https://github.com/openedx-btr-wg/edx-platform/commit/        18effd83f983f497ca0a1535108fa41dc50d06a2#diff-ca02329742db0a77612a18ba1260d178R1-R39
+
+
+Alternatives Considered
+=======================
+
+Refactor configuration
+**********************
+
+Refactoring the configuration repo to reduce the amount of shared code and making it easier to read and understand what is being run when you execute a particular playbook.
+
+This approach was rejected due to the sheer volume of work required to make this change in an environment that is inherently difficult to test.  Also because while it would improve ease of use, it would only provide parity in terms of functionality.  Moving to containers will also improve ease of use as well as unlocking many potential future enhancements.
+
+Jumping to Kubernetes
+*********************
+
+Kubernetes is an open source container orchestration platform pioneered by Google.  While it often occupies the same conversation space as containers because it is a powerful way to manage them, it is a huge increase in complexity and expertise required to operate.  For most installations Kubernetes is currently too much overhead/learning curve for the value.  The edX organization may opt to explore deploying Docker containers this way in the future and would love to collaborate with operators who also decide to use Kubernetes to compare notes.
+
+
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -125,7 +125,7 @@ This new settings structure obviates the need for any other python files in the 
 
 **Config file**
 
-Applications will be configured by a yaml file containing all of the settings variable overrides specified by the operator (including both required settings and secrets as well as default value overrides). The file is made known to the application by an environment variable, ``<APPNAME>_CFG_PATH``, with the path to the file. Versions of this config yaml may be provided in the application repo for certain environments such as development and test. However, for all other environments (e.g. production), the file will need be managed elsewhere.
+Applications will be configured by a yaml file containing all of the settings variable overrides specified by the operator (including both required settings and secrets as well as default value overrides). The file is made known to the application by an environment variable, ``<APPNAME>_CFG_PATH``, with the path to the file. Versions of this config yaml may be provided in the application repo for certain environments such as development and test. However, for all other environments (e.g. production), the file will need to be managed elsewhere.
 
 Since defaults are provided by the application, many smaller deployments should not need to do much more than provide the required settings to operate. For development environments the config will likely change the defaults to more development appropriate values, e.g. debug settings, log levels, email settings etc.
  
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -33,11 +33,11 @@ Abstract
 
 Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
 
-The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination.  How those files are generated and managed is the responsibility of each operator.
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination. How those files are generated and managed is the responsibility of each operator.
 
-All standard operations for running and managing any Open edX application will be documented.  How those procedures are executed is the responsibility of each operator.
+All standard operations for running and managing any Open edX application will be documented. How those procedures are executed is the responsibility of each operator.
 
-Wherever possible, operating concerns will live within the codebase they are concerned with.  Thus it should be expected that all applications will provide the following:
+Wherever possible, operating concerns will live within the codebase they are concerned with. Thus it should be expected that all applications will provide the following:
 
 * A Dockerfile from which the container image can be built.
 * A documented settings file with production-ready defaults.
@@ -47,13 +47,13 @@ Wherever possible, operating concerns will live within the codebase they are con
 Context
 =======
 
-Back when edx-platform was the only code running, the Ansible playbooks in the configuration repo were manageable, though still not necessarily easy to understand.  As edX has invested in microservices, the configuration repo has grown to support the configuration of all of them.  As patterns emerged shared concerns were pulled out into unversioned shared roles making it hard to change without impacting all apps or wiring configurable toggles throughout much of the shared space.
+Back when edx-platform was the only code running, the Ansible playbooks in the configuration repo were manageable, though still not necessarily easy to understand. As edX has invested in microservices, the configuration repo has grown to support the configuration of all of them. As patterns emerged shared concerns were pulled out into unversioned shared roles making it hard to change without impacting all apps or wiring configurable toggles throughout much of the shared space.
 
-The end result is that the configuration repo has grown into a sprawling codebase that attempts to be everything for everyone trying to configure servers to run Open edX installations. There are a wide variety of roles, plays, and utility scripts with branching paths within each to support being executed against different base operating systems and many various configurable operational needs.  None of this is helped by the fact that edX.org operates at a scale that most installations do not.  With a mix of edX.org specific code, branching paths that are not regularly tested, and the high amount of required understanding around how all the pieces interact to change any of it, this approach to operating is a clear source of friction for developing and running the platform. Nothing proves this point more than the fact that the edx-platform README now explicitly warns users to leave it to the professionals.
+The end result is that the configuration repo has grown into a sprawling codebase that attempts to be everything for everyone trying to configure servers to run Open edX installations. There are a wide variety of roles, plays, and utility scripts with branching paths within each to support being executed against different base operating systems and many various configurable operational needs. None of this is helped by the fact that edX.org operates at a scale that most installations do not. With a mix of edX.org specific code, branching paths that are not regularly tested, and the high amount of required understanding around how all the pieces interact to change any of it, this approach to operating is a clear source of friction for developing and running the platform. Nothing proves this point more than the fact that the edx-platform README now explicitly warns users to leave it to the professionals.
 
     Installing and running an Open edX instance is not simple. We strongly recommend that you use a service provider to run the software for you. 
 
-The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace.  In brief it is:
+The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace. In brief it is:
 
 * Use Dockerfiles to capture system dependencies and operate the resulting images in all environments.
 * Simplify and standardize how to configure an application.
@@ -72,21 +72,21 @@ For managing microservice architectures containers are an industry standard for 
 * **Improved security.** Containers improve security by isolating applications from the host system and from each other.
 * **Improved application development.**  Allows developers to run a more production-like environment locally, preventing discrepancies between environments.
 * **Speed.** Start, create, replicate or destroy containers in seconds. Increases velocity for development, deployment, and production operations.
-* **Operational simplicity/consistency.**  By isolating application processes from the host OS it is simpler to manage and maintain the host OS.  Homogeneous administration of heterogeneous components, reduces the range of skill sets required to operate environments.
+* **Operational simplicity/consistency.**  By isolating application processes from the host OS it is simpler to manage and maintain the host OS. Homogeneous administration of heterogeneous components, reduces the range of skill sets required to operate environments.
 
 
 **Clear System Dependency Documentation**
 
-For operators who don’t want to run in Docker containers the Dockerfile acts as documentation for the system dependencies they will need to recreate.  Everything an operator needs to know about how to set up their environment is all in one place rather than distributed across multiple overlapping and overriding files.
+For operators who don’t want to run in Docker containers the Dockerfile acts as documentation for the system dependencies they will need to recreate. Everything an operator needs to know about how to set up their environment is all in one place rather than distributed across multiple overlapping and overriding files.
 
-In order to function as documentation for operators Dockerfiles will be well-commented and use `native Dockerfile syntax`_ to describe the image.  No Ansible should be run as part of the image build. Similarly, bash scripts should be avoided. If they must be used due to a limitation of the Dockerfile commands, the scripts must live within the same codebase as the Dockerfile.
+In order to function as documentation for operators Dockerfiles will be well-commented and use `native Dockerfile syntax`_ to describe the image. No Ansible should be run as part of the image build. Similarly, bash scripts should be avoided. If they must be used due to a limitation of the Dockerfile commands, the scripts must live within the same codebase as the Dockerfile.
 
 
 .. _native Dockerfile syntax: https://docs.docker.com/engine/reference/builder/#dockerfile-reference
 
 **Docker Images**
 
-edX will provide Docker images for applications that captures the latest code on the master branch as well as images representing named releases.  edX will not provide these images for named releases prior to the acceptance and implementation of this OEP which is Juniper at time of writing.
+edX will provide Docker images for applications that captures the latest code on the master branch as well as images representing named releases. edX will not provide these images for named releases prior to the acceptance and implementation of this OEP which is Juniper at time of writing.
 
 
 Configuration
@@ -94,9 +94,9 @@ Configuration
  
 **Django settings**
 
-Having a single artifact that runs with different configurations increases stability by improving development parity with other deployment environments.  edX applications already support configuration overrides via a yaml file for production environments, but development and test environments tend to configure the application using different code paths via a settings/devstack.py or settings/test.py file.
+Having a single artifact that runs with different configurations increases stability by improving development parity with other deployment environments. edX applications already support configuration overrides via a yaml file for production environments, but development and test environments tend to configure the application using different code paths via a settings/devstack.py or settings/test.py file.
 
-Additionally it is not clear which settings are required to be overridden and which settings have values that may technically work but are inappropriate for production systems.  To alleviate these issue edX django applications will adopt the following settings structure:
+Additionally it is not clear which settings are required to be overridden and which settings have values that may technically work but are inappropriate for production systems. To alleviate these issue edX django applications will adopt the following settings structure:
 
 .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,152 @@
...
+.. code-block:: text
+
+  settings
+  ├── __init__.py
+  ├── required.py
+  └── defaults.py
+
+
+*  ``__init__.py`` - Sourcing our config from this file within the settings directory takes advantage of django defaults and means that settings will be picked up automatically without needing to specify ``--settings`` anywhere.  This entry point would import ``required.py``, ``defaults.py``, and the code to override both from a config file.
+*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default.  EG LMS_BASE_URL which will be different per environment.
```
timmc-edx, line 100:
```suggestion
*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default.  E.g. LMS_BASE_URL which will be different per environment.
```
(in at least one other place as well)

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,152 @@
...
+*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables will all be initialized to ``NONE`` and the application will not start unless they are set.  This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
```
timmc-edx, line 105: The `None` value, or the string value `"NONE"`?

bradenmacdonald:
@bderusha This sounds great, but will be a big change! Could you perhaps talk more about the rollout strategy for this?

As we've mentioned before, a partial change to this approach such as removing the config files from `configuration` but without the rest of these pieces (specifically strict `required` vs. `default` settings files, but also proper containerization and provisioning stories, and so on) [could be very difficult for those of us who run hundreds of Open edX instances](https://github.com/edx/configuration/pull/5414#issuecomment-588398918), so my hope would be to see either (1) as much of this changed as possible in a single Open edX release cycle (Koa?), or (2) to have a transition phase wherein the `configuration` repo has ansible scripts that accept all the old ansible variables and spit out the new-style configuration scripts and then deploy the apps using containers.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination.  How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented.  How those procedures are executed is the responsibility of each operator.
+
+Wherever possible, operating concerns will live within the codebase they are concerned with.  Thus it should be expected that all applications will provide the following:
+
+* A Dockerfile from which the container image can be built.
```
morenol, line 42: This looks great, what are the plans with the devstack?, currently the yaml file used in the devstack is defined in https://github.com/edx/configuration/tree/master/docker/build for each service. Do you have any plans to use other repository? Maybe that can be in the service repo itself

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,152 @@
...
+*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables will all be initialized to ``NONE`` and the application will not start unless they are set.  This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
+
+This new settings structure obviates the need for any other python files in the settings directory (such as ``devstack.py``, ``test.py``, etc).  The values currently set in those files should be moved to a corresponding ``devstack.yml``, ``test.yml``, etc in the same settings directory.  This gives developers and operators more consistency across environments since the same code paths are being executed with different values.
```
felipemontoya, line 107: Since this structure is mapped to the settings structure of the plugins for edx-platform, there could be a note about the path of upgrade for plugin maintainers. Perhaps keeping the old structure as a backwards compatibility layer for one open-release.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,152 @@
...
+
+    Installing and running an Open edX instance is not simple. We strongly recommend that you use a service provider to run the software for you. 
+
+The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace.  In brief it is:
+
+* Use Dockerfiles to capture system dependencies and operate the resulting images in all environments.
```
regisb, line 58: Should we emphasize the fact that dockerfiles should be written in a way that make them transparent and approachable? (i.e: we should not run ansible playbooks as part of building a container image)

oeps/oep-0045-arch-ops-and-config.rst
bderusha: This should be answered [later in the OEP](https://github.com/edx/open-edx-proposals/pull/143/files#diff-9d5e701bc9f15777166f8459ecffb088R107), let me know if I can provide more clarity there.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Agreed.  We should certainly have a plan for this.  I would consider that part of the implementation strategy.  I would like to capture that plan in the OEP but separate the conversations.

Let's move this topic over to this follow-on PR https://github.com/edx/open-edx-proposals/pull/144

bderusha:
@bradenmacdonald The implementation strategy for this will be extremely important.  I would like to have separate conversations about whether we think this is a good direction for the project and how we go about getting there.  As such, I have created a follow-up PR where we can have that tactical discussion and eventually merge the results back into this OEP.

https://github.com/edx/open-edx-proposals/pull/144

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Yes!  I just added that bit to the [containers section](https://github.com/edx/open-edx-proposals/pull/143/files#diff-9d5e701bc9f15777166f8459ecffb088R82) . Trying to keep the summary more brief.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,169 @@
...
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination.  How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented.  How those procedures are executed is the responsibility of each operator.
+
+Wherever possible, operating concerns will live within the codebase they are concerned with.  Thus it should be expected that all applications will provide the following:
+
+* A Dockerfile from which the container image can be built.
+* A documented settings file with production-ready defaults.
+* A documented settings file with all of the configuration values that must be defined by the operator.
+* An operations manual documenting standard operations required to run and maintain the application.
```
coryleeio, line 45: Would this include / should this include a rolling changelog for changes that will affect operators?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,169 @@
...
+
+
+Abstract
+========
+
+Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination.  How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented.  How those procedures are executed is the responsibility of each operator.
```
regisb, line 38: There are multiple double spaces in the next few paragraphs ("  ").

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,169 @@
...
+* **Speed.** Start, create, replicate or destroy containers in seconds. Increases velocity for development, deployment, and production operations.
+* **Operational simplicity/consistency.**  By isolating application processes from the host OS it is simpler to manage and maintain the host OS.  Homogeneous administration of heterogeneous components, reduces the range of skill sets required to operate environments.
+
+
+**Clear System Dependency Documentation**
+
+For operators who don’t want to run in Docker containers the Dockerfile acts as documentation for the system dependencies they will need to recreate.  Everything an operator needs to know about how to set up their environment is all in one place rather than distributed across multiple overlapping and overriding files.
+
+In order to function as documentation for operators Dockerfiles will be well-commented and use `native Dockerfile syntax`_ to describe the image.  No Ansible should be run as part of the image build. Similarly, bash scripts should be avoided. If they must be used due to a limitation of the Dockerfile commands, the scripts must live within the same codebase as the Dockerfile.
+
```
regisb, line 83: Users must be able to rebuild the production and development-ready images locally. This means that:

1. the Dockerfile should not point to private pypi/npm/git repos
2. the default values of the Docker ARGuments should not have to be modified for most users

(btw ARG values should be documented, in addition to runtime configuration settings)

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,169 @@
...
+Refactoring the configuration repo to reduce the amount of shared code and making it easier to read and understand what is being run when you execute a particular playbook.
+
+This approach was rejected due to the sheer volume of work required to make this change in an environment that is inherently difficult to test.  Also because while it would improve ease of use, it would only provide parity in terms of functionality.  Moving to containers will also improve ease of use as well as unlocking many potential future enhancements.
+
+Jumping to Kubernetes
+*********************
+
+Kubernetes is an open source container orchestration platform pioneered by Google.  While it often occupies the same conversation space as containers because it is a powerful way to manage them, it is a huge increase in complexity and expertise required to operate.  For most installations Kubernetes is currently too much overhead/learning curve for the value.  The edX organization may opt to explore deploying Docker containers this way in the future and would love to collaborate with operators who also decide to use Kubernetes to compare notes.
+
```
regisb, line 161: Could we consider using environment variables for configuration, even if it's just to dismiss them? IMHO there are simply to many complex (dicts and lists) variables to be defined in edx-platform to set them all as environment variables.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,169 @@
...
+*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables will all be initialized to ``None`` and the application will not start unless they are set.  This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
+
+This new settings structure obviates the need for any other python files in the settings directory (such as ``devstack.py``, ``test.py``, etc).  The values currently set in those files should be moved to a corresponding ``devstack.yml``, ``test.yml``, etc in the same settings directory.  This gives developers and operators more consistency across environments since the same code paths are being executed with different values.
```
regisb, line 117: Are we talking about using the Docker images for local development? I think we should: or at least, we should be able to test the Docker images locally by pointing to localhost (and not S3, for instance).

If we are to use these Docker images for local development, I'm afraid this is an overly optimistic approach for organising the settings. I don't think we will be able to reach this state where we only have defaults.py/required.py without having to resort to huge prod.yml/dev.yml files.

My point is that `required.py` and `defaults.py`, by themselves, will not be usable in either a development or production environment without very large prod.yml/dev.yml files. The reason for that is that prod.py/dev.py contain not only configuration **values**, but also configuration **logic**. 

Let's take a very simple example: in production we need to set `DEBUG = False`. This is just a fact that is true for all users. Does that mean that _all users_ should now add `DEBUG: false` to their yaml files? I don't think it should: there are quite literally hundreds of settings that need to be set separately for production and development, but for which we can easily predict sane defaults.

The alternatives are:
1. Use a single settings file that differentiates between dev and prod environments. This is the approach used by Sentry. See: https://github.com/getsentry/sentry/blob/685427a5dea1e1ad0adb1816256e28f4958f570f/src/sentry/conf/server.py#L68
2. Use separate settings files: base.py, production.py, development.py. This is advocated by [Two Scoops of Django](https://www.amazon.com/Two-Scoops-Django-Best-Practices/dp/0981467342) and it's Jacob Kaplan-Moss'  ["The One True Way"](https://www.slideshare.net/jacobian/the-best-and-worst-of-django/51).

In the case of edx-platform, I would argue for approach 2.

We would have, similar to what we currently do:

base.py:

```
# Reasonable settings
additional_settings = yaml.load(os.environ["CONFIG_PATH"]) if "CONFIG_PATH" in os.environ else {}
DEBUG = additional_settings.get("DEBUG", False)
ALLOWED_HOSTS = additional_settings.get("ALLOWED_HOSTS", ["localhost", "localhost:8000"])
...
```

development.py:

```
from .base import *

# Development logic
DEBUG = True
ALLOWED_HOSTS = ["*"]
SECRET_KEY = "v3rys3cr3t"
```

production.py:

```
```
from .base import *

# Development logic
DEBUG = False
ALLOWED_HOSTS = [additional_settings["HOST"]]
SECRET_KEY = additional_settings["SECRET_KEY"] # fail fast in case this is not defined
...
```

I understand that, by doing so, we lose the benefit of automatically loading the settings from __init__.py, but in my experience this is not a problem: settings will be automatically loaded as soon as we set the [DJANGO_SETTINGS_MODULE](https://docs.djangoproject.com/en/3.0/topics/settings/#envvar-DJANGO_SETTINGS_MODULE) environment variable (which should be set to "prod" by default in the Docker image)

Furthermore, users will still be able to customize production/development settings by creating their own settings files which will import from production.py/development.py (as we do today). This is pretty much the standard way to proceed in the django world.

File: oeps/oep-0045-arch-ops-and-config.rst (hunk 8 above)
Comment (coryleeio) [reply oeps/oep-0045-arch-ops-and-config.rst:83]: Agreed - I would strongly encourage us to ensure that 'docker build .' with no arguments produces a valid image for most users.

Comment (NIXKnight) [reply oeps/oep-0045-arch-ops-and-config.rst:83]: Agreed with @regisb

File: oeps/oep-0045-arch-ops-and-config.rst (hunk 6 above)
Comment (bderusha) [reply oeps/oep-0045-arch-ops-and-config.rst:45]: I think a changelog is a good idea.  I don't know the value of an operations specific log VS a single log to capture all relevant changes to the codebase.  Interested to hear other's thoughts here.

File: oeps/oep-0045-arch-ops-and-config.rst [hunk 11]
```diff
@@ -0,0 +1,169 @@
...
+*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default, e.g. LMS_BASE_URL which will be different per environment.
+*  ``defaults.py`` - all settings other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables will all be initialized to ``None`` and the application will not start unless they are set.  This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
```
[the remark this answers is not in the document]
bderusha: @regisb I reworded this to be less implementation specific.  This is meant to ensure that an operator at least provided a required value.  For most settings I expect that if `None` is a valid input it probably doesn't belong in the `required.py` section anyway.  But the implementation is less important than the sentiment, so I tweaked the language a bit.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default, e.g. LMS_BASE_URL which will be different per environment.
+*  ``defaults.py`` - other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables must be overridden by operators.  The application will check that operators provided these values not start unless they are set. This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
```
iloveagent57, line 121: I think there's a couple of words missing in this sentence: "The application will check that operators provided these values, *and will* not start unless they are set."

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+ 
+
+.. _Sphinx autodoc: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
+
+Operations Manuals
+******************
+
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that application. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the application for web traffic or as an async worker and how to manage the applications underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
```
iloveagent57, line 146: Nit: "how to manage the **application's** underlying..." (with an apostrophe)

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Alternatives Considered
+=======================
+
+Refactor configuration
+**********************
+
+Refactoring the configuration repo to reduce the amount of shared code and making it easier to read and understand what is being run when you execute a particular playbook.
+
+This approach was rejected due to the sheer volume of work required to make this change in an environment that is inherently difficult to test. Also because while it would improve ease of use, it would only provide parity in terms of functionality. Moving to containers will also improve ease of use as well as unlocking many potential future enhancements.
```
iloveagent57, line 161: Nit: 
"Also because while" -> "Also, while"
"Moving to containers will also improve ease of use as well as unlocking" -> "Moving to containers will also improve ease of use and unlock"

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+
+Configuration
+*************
+ 
+**Django settings**
+
+Having a single artifact that runs with different configurations increases stability by improving development parity with other deployment environments. edX applications already support configuration overrides via a yaml file for production environments, but development and test environments tend to configure the application using different code paths via a settings/devstack.py or settings/test.py file.
+
+Additionally it is not clear which settings are required to be overridden and which settings have values that may technically work but are inappropriate for production systems. To alleviate these issue edX django applications will adopt the following settings structure:
```
adamstankiewicz, line 105: nit: "To alleviate these **issues,** edX Django applications..."

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+
+**Config file**
+
+Applications will be configured by a yaml file containing all of the settings variable overrides specified by the operator (including both required settings and secrets as well as default value overrides). The file is made known to the application by an environment variable, ``<APPNAME>_CFG_PATH``, with the path to the file. Versions of this config yaml may be provided in the application repo for certain environments such as development and test. However, for all other environments (e.g. production), the file will need to be managed elsewhere.
+
+Since defaults are provided by the application, many smaller deployments should not need to do much more than provide the required settings to operate. For development environments the config will likely change the defaults to more development appropriate values, e.g. debug settings, log levels, email settings etc.
```
adamstankiewicz, line 130: nit: "..., email settings, etc."

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+ 
+
+.. _Sphinx autodoc: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
+
+Operations Manuals
+******************
+
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that application. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the application for web traffic or as an async worker and how to manage the applications underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
+ 
+In the same vein as not dictating how operators create and manage their application config files, so too will operators be expected to manage how they execute the operations documented in the manual.
```
adamstankiewicz, line 148: nit: "... manage their application config files, **operators will also be** expected to manage how they..." (the "so too will operators" reads a bit awkwardly)

oeps/oep-0045-arch-ops-and-config.rst
nedbat: Changelogs are often categorized. Perhaps "Operations" could be a category along with "Breaking changes", "Bug fixes", etc.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination. How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented. How those procedures are executed is the responsibility of each operator.
+
+Wherever possible, operating concerns will live within the codebase they are concerned with. Thus it should be expected that all applications will provide the following:
+
+* A Dockerfile from which the container image can be built.
+* A documented settings file with production-ready defaults.
```
nedbat, line 43: What does "production-ready" mean exactly? I assume we don't mean, "appropriate for edx.org production"?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+* If scripts must be used due to a limitation of the Dockerfile commands, the scripts must live within the same codebase as the Dockerfile.
+* Default ARG and CMD values should meet the needs of most users without modification.
+* Private or custom install requirements, patches, ARGs, CMD values, etc. should be included via a separately managed Dockerfile built on top of the Open edX image for that codebase.
+
+
+.. _native Dockerfile syntax: https://docs.docker.com/engine/reference/builder/#dockerfile-reference
+
+**Docker Images**
+
+edX will provide Docker images for applications that captures the latest code on the master branch as well as images representing named releases. edX will not provide these images for named releases prior to the acceptance and implementation of this OEP which is Juniper at time of writing.
```
nedbat, line 95: This is a little confusing "for named releases ... which is Juniper".  I think you are saying that Juniper will not have Docker images?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+
+Configuration
+*************
+ 
+**Django settings**
+
+Having a single artifact that runs with different configurations increases stability by improving development parity with other deployment environments. edX applications already support configuration overrides via a yaml file for production environments, but development and test environments tend to configure the application using different code paths via a settings/devstack.py or settings/test.py file.
+
+Additionally it is not clear which settings are required to be overridden and which settings have values that may technically work but are inappropriate for production systems. To alleviate these issues edX django applications will adopt the following settings structure:
```
nedbat, line 105: "Django" should be capitalized throughout.

oeps/oep-0045-arch-ops-and-config.rst
fredsmith: openedx production ready.  meaning, we would expect an openedx user to run the image on a public website and things like DDT or verbose logging or dev user accounts wouldn't be present.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: One of the goals of this OEP is to have an application that executes the same codepaths in development as in production wherever reasonable.  I believe that this is a place that could reasonably be one code path.  Let's take the dev defaults case, because it's a very valid point.  Here is how I would see that working.

required.py
```
# Variables required to run, but without reasonable production defaults
LMS_BASE_URL = None
JWT_AUTH = None
...
```

defaults.py
```
# Variable with reasonable defaults
DEBUG = false
DEFAULT_PAGE_SIZE = 25
STATIC_URL = '/static/'
...
```

__init__.py
```
import os

from .required import *
from .defaults import *

def get_env_setting(setting):
    """ Get the environment setting or return exception """
    try:
        return os.environ[setting]
    except KeyError:
        error_msg = u"Set the %s env variable" % setting
        raise ImproperlyConfigured(error_msg)

CONFIG_FILE = get_env_setting('APP_CFG')

with open(CONFIG_FILE, encoding='utf-8') as f:
    config_from_yaml = yaml.safe_load(f)
    vars().update(config_from_yaml)
```

development.yml (lives in the repo either like this or as starting yaml to copy `development.yml.example`)
```yaml
# required
LMS_BASE_URL: localhost:18000
JWT_AUTH: { ... auth ... }

# default overrides
DEBUG: true
DEFAULT_PAGE_SIZE: 1000
```

production.yml (managed securely elsewhere)
```yaml
# required
LMS_BASE_URL: www.my-prod-lms.com
JWT_AUTH: { ... prod auth ... }
```

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+*  ``defaults.py`` - other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables must be overridden by operators.  The application will check that operators provided these values, and will not start unless they are set. This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
+
+This new settings structure obviates the need for any other python files in the settings directory (such as ``devstack.py``, ``test.py``, etc). The values currently set in those files should be moved to a corresponding ``devstack.yml``, ``test.yml``, etc in the same settings directory. This gives developers and operators more consistency across environments since the same code paths are being executed with different values.
```
nedbat, line 123: Developers have found "private.py" helpful.  That will also go away?

oeps/oep-0045-arch-ops-and-config.rst
bderusha: I am all for giving developers and operators as many of the values defaulted to suit their environments as possible.  It is also likely that for testing we will need/want a `test.yml` to override certain settings.  That should also live in the repo.

IMO this method increases the likelihood that new settings end up in the right place, documented, with proper defaults. And that when people are writing and testing their apps they aren't surprised by how it works in production.

oeps/oep-0045-arch-ops-and-config.rst
nedbat: Looking at edx-platform/lms/envs/test.py, there is a little bit of executable code in it, for example to randomize keys.  How will that be handled in a test.yml world?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+**Config file**
+
+Applications will be configured by a yaml file containing all of the settings variable overrides specified by the operator (including both required settings and secrets as well as default value overrides). The file is made known to the application by an environment variable, ``<APPNAME>_CFG_PATH``, with the path to the file. Versions of this config yaml may be provided in the application repo for certain environments such as development and test. However, for all other environments (e.g. production), the file will need to be managed elsewhere.
+
+Since defaults are provided by the application, many smaller deployments should not need to do much more than provide the required settings to operate. For development environments the config will likely change the defaults to more development appropriate values, e.g. debug settings, log levels, email settings, etc.
+ 
+**Config file generation & management**
+
+Due to the varied needs and processes of different operators, how the config files are created, managed, or otherwise end up on the server is up to the operator and will depend greatly on their deployment strategy.
```
nedbat, line 134: Will we be generating test.yml?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Operations Manuals
+******************
+
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that application. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the application for web traffic or as an async worker and how to manage the application's underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
+ 
+In the same vein as not dictating how operators create and manage their application config files, operators will also be expected to manage how they execute the operations documented in the manual.
+
+.. _this commit: https://github.com/openedx-btr-wg/edx-platform/commit/        18effd83f983f497ca0a1535108fa41dc50d06a2#diff-ca02329742db0a77612a18ba1260d178R1-R39
```
nedbat, line 150: Looks like this URL broke.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: @nedbat it's a squishy word on purpose.  It means production ready as opposed to development ready.  A great example of how this would look was highlighted by Regis. 

`DEBUG = false` is a production-ready default.  `DEBUG = true` is not.

But to directly answer you, production-ready means reasonable to run on "most/many" production systems.  Anything edx.org specific would not pass this bar.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Yes.  Is this better?

`edX will not provide these images for named releases prior to the acceptance and implementation of this OEP (Aspen through and including Juniper at time of writing).`

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Yes.  However we should talk about how to replicate it's usefulness in a `yml` config world.

stvstnfrd: FWIW, I took a look back at how we handled this generally (custom logic in settings `.py` files) on the Stanford fork...

We used to have some crazier things (heck, we shadowed the entire `{lms/cms}/envs/` directory structure), but here's what we were actually using in practice, at the end:
- Contextually removing some settings (`INSTALLED_APPS`, `MIDDLEWARE_CLASSES`) [1]
- expanding wildcard paths [2]
- overriding logging config [3]

Re: `private.py` specifically:
We definitely used this and I'd worry that folks won't be able to replicate all of the capabilities of custom Python logic w/in a `.yml` file. Is there a specific reason we're looking to remove this feature, eg, "`private.py` considered harmful"?
Or is this more "getting caught up in the storm" of cleaning out the settings directory?
Since `private.py` is explicitly `git-ignored`, it shouldn't increase the number of files checked in to that directory, with minimal maintenance overhead on the implementation [4].

I'm open to alternatives, as long as we don't have to reinvent the wheel.

- [1] https://github.com/Stanford-Online/edx-platform/blob/master/openedx/stanford/cms/envs/test.py#L8-L11
- [2] https://github.com/Stanford-Online/edx-platform/blob/master/openedx/stanford/cms/envs/common.py#L67-L69
- [3] https://github.com/Stanford-Online/edx-platform/blob/master/openedx/stanford/lms/envs/devstack.py#L9-L20
- [4] https://github.com/Stanford-Online/edx-platform/blob/master/lms/envs/devstack.py#L273-L275

oeps/oep-0045-arch-ops-and-config.rst
stvstnfrd: While the deployment processes will often vary, does (or _should_) the creation vary?

If we're forcing certain values to be set by operators (and enforcing via early failure), it seems like we ought to provide a canonical way to create the files, or at least provide a blessed reference implementation.

The actual implementation would vary, but I'm imaging at least a helper of sorts, something in the spirit of:
- https://github.com/edx/configuration/blob/eec2b4ae2fdf88fd17aabfc11a308951b243f9e4/util/install/generate-passwords.sh

oeps/oep-0045-arch-ops-and-config.rst
bderusha: @nedbat I just looked through that file and I didn't see any executable code that couldn't just be a preset value.  I don't think we will be able to get away with just a straight import and override from the yaml, but I do think that we can keep any executable code required to set up the app in one code path.

If people have valid cases for executable code in the settings files that only apply to a particular environment I would love to see those cases.  I wasn't able to find any myself.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: @nedbat I think that applications should define a `test.yml` that can be used, I would not expect to need to generate them.

@stvstnfrd I don't have strong opinion here.  My thought was that it is so simple that no reference or helper would be required.  Value defined in `required.py' or `default.py` maps 1-to-1 to the value you stick in your yml file to override it.  Interested to hear more about what the helper would actually do

oeps/oep-0045-arch-ops-and-config.rst
bderusha: @stvstnfrd thanks for all those great references!  I will copy a line written in previous comment

> One of the goals of this OEP is to have an application that executes the same code paths in development as in production wherever reasonable. I believe that this is a place that could reasonably be one code path.

I would love for that to be our starting assumption and see if there is a way to achieve the needs of operators and developers without compromising current functionality (though we may need to change current processes/mindsets).  Your examples bring up a great and very common configuration challenge.  `INSTALLED_APPS` and `MIDDLEWARE_CLASSES`

Would it be reasonable for all configuration files to be able to define:
```
INSTALLED_APPS:
- 'app1'
- 'app2'
APPEND_INSTALLED_APPS:
- 'app3'
REMOVE_INSTALLED_APPS:
- 'app2'
```


`INSTALLED_APPS` would completely overwrite the value.  `APPEND` and `REMOVE` would take the corresponding list of value and perform that action.  Allowing the code path to resolve these common cases the same way across the board.

stvstnfrd: @bderusha Cool, yeah, I probably should have split apart the two lines of thought in that comment (`private.py` vs general config), as I am implying the former could be be kept around w/ negligible overhead/maintenance, but not implying as much with the latter.

For the general config stuff, I didn't mean to imply "this can't be done!"; just wanted to flag some examples that may not yet be accounted for :)

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+   :local:
+   :depth: 3
+
+
+Abstract
+========
+
+Wherever possible, The edX organization will provide and manage Docker images as the medium for packaging operating dependencies of Open edX applications rather than providing Ansible playbooks.
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination. How those files are generated and managed is the responsibility of each operator.
```
nasthagiri, line 36: For completeness, can you define "application"?  Is it equivalent to an independently deployable service?  Does it include `Django App Plugins` or feature libraries?  (I'm assuming you're referring to IDAs here.)

oeps/oep-0045-arch-ops-and-config.rst
stvstnfrd: @bderusha I guess I don't know either :) I'll probably have a better idea upon seeing potential implementations.
And no, I'm not expecting that at this point, just _slightly_ pushing back on the idea of deferring the responsibility permanently :)

My ears perk up whenever I hear, "operators can figure this out themselves", as those were often the areas that bit us most frequently :P

It's probably less of an issue for us than it will be for the community.
- I assume we'll generally make the change to `{required,default}.py` immediately in tandem with the change to the `${env}.yml` (whether that's in a private repo or what not).
- On the other hand, community operators are going to "inherit" the changes to `{required,default}.py` weeks/months/years later, and then will need to reconcile the two files, line by line (?? since it won't be a simple `diff`, since we're comparing `py` and `yml` files) or while auditing the `CHANGELOG`.
  - Of course, it won't be the "worst thing ever" :tm: , but it seems like an extra tedium to incur upon operators, that we won't feel ourselves.

Off the top of my head (again, I'm sure seeing sample implementations will help my mental model), I could imagine a script that auto-merges/reconciles the `py` files with the `yml`. Something like:

```sh
./upgrade-config --input ironwood.yml --output juniper.yml
# defaults.py and required.py are implied default inputs
```

So assuming input files like:

```py
## required.py
ADMIN_EMAIL = YAML_CONFIG['ADMIN_EMAIL']
NEW_TOGGLE_ADDED_IN_JUNIPER = YAML_CONFIG['NEW_TOGGLE_ADDED_IN_JUNIPER']
```

```py
## defaults.py
LMS_BASE = YAML_CONFIG['LMS_BASE']
ANOTHER_TOGGLE_ADDED_IN_JUNIPER = YAML_CONFIG['ANOTHER_TOGGLE_ADDED_IN_JUNIPER']
```

```yml
## ironwood.yml
ADMIN_EMAIL: 'admin@example.com'
LMS_BASE: 'localhost:18000'
OLD_TOGGLE_REMOVED_IN_JUNIPER: true
```

Might yield an output file like:

```yml
## juniper.yml
## This file was automatically generated by: `upgrade-config`

## Deprecated settings
## The following are now unsupported in this release:
# OLD_TOGGLE_REMOVED_IN_JUNIPER: true

## Required settings
## These must all be set:
ADMIN_EMAIL: 'admin@example.com'
# NEW_TOGGLE_ADDED_IN_JUNIPER: 'recommended default value'

## Default overrides
## These are optional overrides:
LMS_BASE: 'localhost:18000'
# ANOTHER_TOGGLE_ADDED_IN_JUNIPER: 'default to the value in `defaults.py`'
```

An operator would still need to tweak the output file (to update new required values), but I can imagine some process like this making it easier and more obvious to automatically see:
- which new required settings have been added
- which new optional overrides are available
- which settings have been deprecated and are no longer supported

I'm sure there would be deployments where this isn't as useful, but I'd expect it to remove busy work and mistakes for many operators.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+ 
+
+.. _Sphinx autodoc: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
+
+Operations Manuals
+******************
+
+A clear manual of operations will exist in the form of RST files in an ``operations`` directory within the ``documentation`` directory for that application. See `this commit`_ for an example provided by the Open edX Build-Test-Release working group. The operations docs will cover common operations such as how to run the application for web traffic or as an async worker and how to manage the application's underlying database schema. It will also include a list of potential maintenance tasks operators may want to leverage such as clearing sessions or applying security patches. Finally it will include the list of ad-hoc management commands operators can use to help handle edge case or one-time operations.
```
nasthagiri, line 146: Should the `documentation` directory be the `docs` directory, in keeping with [OEP-19](https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html)?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+The configuration for running applications in different environments will be done through a config file per unique [application, environment] combination. How those files are generated and managed is the responsibility of each operator.
+
+All standard operations for running and managing any Open edX application will be documented. How those procedures are executed is the responsibility of each operator.
+
+Wherever possible, operating concerns will live within the codebase they are concerned with. Thus it should be expected that all applications will provide the following:
+
+* A Dockerfile from which the container image can be built.
+* A documented settings file with production-ready defaults.
+* A documented settings file with all of the configuration values that must be defined by the operator.
```
nasthagiri, line 44: For IDAs with dependent configurable libraries and potential plugins, do we imagine eventually generating readable docs on settings from dependencies, keeping it as an exercise for the reader, or other?

nasthagiri, line 44: This is referring to the `required.py` module, right?  _(May want to just add the file's names here and above for added clarity.)_

Can we clarify this as:
A documented settings file with **remaining (or other)** configuration values that must be defined by the operator.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+Kubernetes is an open source container orchestration platform pioneered by Google. While it often occupies the same conversation space as containers because it is a powerful way to manage them, it is a huge increase in complexity and expertise required to operate. For most installations Kubernetes is currently too much overhead/learning curve for the value. The edX organization may opt to explore deploying Docker containers this way in the future and would love to collaborate with operators who also decide to use Kubernetes to compare notes.
+
+
+
+Implementation Strategy
+=======================
+
+Discussion of implentation of this OEP will happen in a `separate Pull Request`_ .
+
+.. _separate Pull Request: https://github.com/edx/open-edx-proposals/pull/144
```
nasthagiri, line 175: Consider adding references to other industry best practices or reference examples if external exploration has been done.

Here are a few I had bumped into in the past:
* https://www.agileleaf.com/blog/a-better-way-to-manage-settings-py-in-your-django-projects/
* https://www.rdegges.com/2011/the-perfect-django-settings-file/
* https://github.com/openfun/richie/blob/master/src/richie/apps/core/defaults.py

oeps/oep-0045-arch-ops-and-config.rst
idegtiarov: Good point, thanks.

oeps/oep-0045-arch-ops-and-config.rst
e0d: Would Open edX platform services be clearer?

oeps/oep-0045-arch-ops-and-config.rst
e0d: Should this include things beyond the "must be defined" set?  Meaning should there be a broader requirement for specifying all configuration options.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+* Use Dockerfiles to capture system dependencies and operate the resulting images in all environments.
+* Simplify and standardize how to configure an application.
+* Document standard operations.
+
+ 
+Containers
+**********
+
+**Why Containers**
```
e0d, line 66: Some of this content isn't universally accepted to be true, for example that containers are more secure.  I'm not sure it's required to make this case here.  Containerization feels like a different OEP to me.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+    Installing and running an Open edX instance is not simple. We strongly recommend that you use a service provider to run the software for you. 
+
+The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace. In brief it is:
+
+* Use Dockerfiles to capture system dependencies and operate the resulting images in all environments.
```
e0d, line 58: Is extending what the dockerfiles do outside the scope of this OEP?  Meaning, say I need to add a new binary dependency, an NLP library or fortran library, what's the pattern for doing that?

Should the recommendation for altering python dependencies to add new Django apps or install plugin be specified here?

oeps/oep-0045-arch-ops-and-config.rst
idegtiarov: Agree with @stvstnfrd last point sounds very reasonable.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Yes! I will make that more clear/explicit

oeps/oep-0045-arch-ops-and-config.rst
regisb: The "django way" is to have both a changelog (release notes) and a manual that indicates differences with the previous version. 

For instance, in the manual:

> Run `foobarcocktail --no-dry-run` to update the foogarizer. This feature was introduced in release Zebulon

And then in the changelog:

> # Zebulon release notes:
> 
> [Operations][Breaking change] In order to update the foogarizer the `--no-dry-run` option must be added to the `foobarcocktail` command.
>

oeps/oep-0045-arch-ops-and-config.rst
regisb: It's true that containers might not improve security across the board, but they certainly help to address *some* security problems. For instance: it's easier to limit the exposition of service ports to the outside world. Also, applications are better isolated from one another.

I feel like it's important to advocate the use of containers in this OEP, as its the primary alternative for non-edX users.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+*  ``required.py`` - all settings which are required to run and do not have a reasonable production-ready default, e.g. LMS_BASE_URL which will be different per environment.
+*  ``defaults.py`` - other settings which will have production-ready defaults
+
+The settings defined in ``required.py`` and ``defaults.py`` files are mutually exclusive, representing all application specific settings as well as installed library settings whose values either must be provided or whose defaults are not considered production-ready.
+
+``required.py`` variables must be overridden by operators.  The application will check that operators provided these values, and will not start unless they are set. This allows operators to fail fast rather than finding out about an unset value when users exercise those breaking codepaths. Application developers are encouraged to keep the list of required settings to a minimum.
```
feanil, line 121: Is it too specific to say we'll use the ImproperlyConfigured exception?

feanil, line 121: If we're we're overriding the defaults from the config file, then I think you won't actually need to do any assignment in the `required.py`  it will only have the validation checks.  From what I can tell about sphinx autodoc, it looks like it's setup to pickup assignments, so you may end up in a situation where you have to do an assignment for the docs to pick it up even though it won't be necessary for the code to work correctly.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+Due to the varied needs and processes of different operators, how the config files are created, managed, or otherwise end up on the server is up to the operator and will depend greatly on their deployment strategy.
+ 
+**Documentation of settings**
+
+The settings found in both the ``required.py`` and ``defaults.py`` files will be documented to describe what they are and how they should be used. The documentation will consist of `Sphinx autodoc`_ compatible comments before each setting. For reference that is a ``“comment with special formatting (using a #: to start the comment instead of just #)”``. This keeps documentation close to the code as it is being written, while allowing it to be surfaced in generated docs.
+ 
+
+.. _Sphinx autodoc: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
+
+Operations Manuals
```
giovannicimolin, line 143: The `edx/configuration` repo also handles setting up requirements between LMS/Studio and other IDAs, such as creating OAuth2 credentials/accouts ([oauth-client-setup](https://github.com/edx/configuration/tree/master/playbooks/roles/oauth_client_setup/)).

Since this OEP's goal is to replace `configuration`, should we add instructions in the operations manuals about what needs to be done in order to configure each IDA.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+    Installing and running an Open edX instance is not simple. We strongly recommend that you use a service provider to run the software for you. 
+
+The proposal below outlines how we can create a cleaner, more intuitive interface for operating Open edX and in doing so help both edX.org and the Open edX community at large achieve better outcome at a faster pace. In brief it is:
```
giovannicimolin, line 56: This OEP does a really good job describing how configuration should be done on each app and sets great standards for building images, but it doesn't cover the "how to put things together" like the current `configuration` repo implicitly does.

For example, when running `openedx_native.yml`, the playbooks set up, on a single instance, all the services needed to run a full Open edX deployment:
- LMS
- Studio
- Workers
- IDAs
- Other support services
- Handles setting up LMS/Studio and wiring between them and the IDAs

I know that this OEP doesn't cover `deployment` because there are a lot of different ways to do that, but the requirements for a Open edX deployment are not described anywhere (except implicitly in the `configuration` repo).
So the `configuration` repo isn't the best reference, but it's currently all we have for deploying from scratch.

Can we add a section on this OEP (or create a follow up OEP), to better document and map how the platform is supposed to work? What is each piece, and what's required for a "basic" installation if starting from scratch using the docker images?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+
+
+.. _native Dockerfile syntax: https://docs.docker.com/engine/reference/builder/#dockerfile-reference
+
+**Docker Images**
+
+edX will provide Docker images for applications that captures the latest code on the master branch as well as images representing named releases. edX will not provide these images for named releases prior to the acceptance and implementation of this OEP (Aspen through and including Juniper at time of writing).
+
+
+Configuration
```
giovannicimolin, line 98: How does this new configuration strategy handles differentiating between a container running only workers to a container actually serving the LMS/Studio pages (gunicorn).

Will each repo have a separate dockerfile for each container "type" (celery worker, gunicorn, etc) or this is going to be managed through configuration?

I'm asking this because the workers and the LMS run off the same codebase with the same requirements, with just different entrypoints.

Is it worth adding a section here about this specific point?

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+**********************
+
+Refactoring the configuration repo to reduce the amount of shared code and making it easier to read and understand what is being run when you execute a particular playbook.
+
+This approach was rejected due to the sheer volume of work required to make this change in an environment that is inherently difficult to test. Also, while it would improve ease of use, it would only provide parity in terms of functionality. Moving to containers will also improve ease of use and unlock many potential future enhancements.
+
+Jumping to Kubernetes
+*********************
+
+Kubernetes is an open source container orchestration platform pioneered by Google. While it often occupies the same conversation space as containers because it is a powerful way to manage them, it is a huge increase in complexity and expertise required to operate. For most installations Kubernetes is currently too much overhead/learning curve for the value. The edX organization may opt to explore deploying Docker containers this way in the future and would love to collaborate with operators who also decide to use Kubernetes to compare notes.
```
arbrandes, line 166: To hammer home the Blessed Deployment Project idea, the latter is where the Kubernetes discussion should happen.  While I believe the intention to build such a project should live in _this_ OEP, the specifics can be handled in a separate one.  (Spoiler alert: Kubernetes would be an excellent contender for orchestrating the Open edX applications. :)

oeps/oep-0045-arch-ops-and-config.rst
kdmccormick: Is it worth specifying how how the service will check if all of its required settings have been overriden? Or is that up to the service?

kdmccormick: It's not clear to my that `required.py` will do any validation, in which case it would exist entirely for documentation. I'm not sure how I feel about that. Could we instead just put required settings in `defaults.py`, and used Sphinx annotations to indicate that they're required.

If I'm wrong, and `required.py` will in fact do validation, then I recommend renaming `required.py` to `validate_required.py`, and clarifying that it will be imported *after* the YAML is loaded.

oeps/oep-0045-arch-ops-and-config.rst
kdmccormick: @bderusha Your `INSTALLED_APPS` example does a good job of highlighting how handling some Django settings will be more complex than just "fully override it with the value in the YAML". I think [edx-platform OPTIONAL_APPS](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L3152-L3196) also illustrates this.

I think it would be worth eventually specifying our "special" Django settings (`APPEND_INSTALLED_APPS`, `REMOVE_MIDDLEWARE`, etc.) somewhere visible, with a reference implementation of how to handle them in `__init__.py`. That being said, I think it's out of the scope of this PR, as it would require some digging and discovery to find out what those special settings should be and how we would handle them. Maybe that's something that would end up in your follow-up PR.

oeps/oep-0045-arch-ops-and-config.rst
arbrandes: PS: @bderusha, we'll be discussing the reference deployment project idea during the [next contributor's meetup](https://trello.com/c/U0H7iYze/48-contributors-meetup-may-7th-agenda), on Thursday, if you care to join us. :)

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -0,0 +1,175 @@
...
+.. list-table::
+
+   * - OEP
+     - :doc:`OEP-0045 </oeps/oep-0045>`
+   * - Title
+     - Configuring and Operating Open edX
+   * - Last Modified
+     - 2020-04-06
+   * - Authors
+     - Bill DeRusha
```
davidjoy, line 14: It'd be useful to include contact email addresses here.

oeps/oep-0045-arch-ops-and-config.rst
bderusha: I think the how is up for debate and can be moved to the implementation section of this OEP

oeps/oep-0045-arch-ops-and-config.rst
bderusha: I added a note to specify that the expectation is that custom libraries etc be managed in a separate dockerfile using the openedx image as the base

oeps/oep-0045-arch-ops-and-config.rst
bderusha: changed to IDAs

oeps/oep-0045-arch-ops-and-config.rst
bderusha: Because the implementation details of this part matter quite a bit to the final design, I think I would like to move this conversation over to the "how we get there" part.  We should be open to amending this section as we coalesce around exactly what operating pattern needs to exist to realistically land somewhere where we have 1 codepath for configuration and clear delineations for operators around what they must provide to run the app.

bderusha:
Thank you, everyone for the great discussion.  I think we are ready to start the conversation around how we can tactically move from where we are today to this bold new world.  I would like to note that there were a few important implementation details we should plan on addressing during the next phase of discussion:

* required.py and/or how we can ensure clear communications to operators about what is required and fail fast if not provided.
* guidance/paths/helpers for migrating between releases
* whether private.py would be useful for local dev or whether more complicated overrides would be better served as supported in the main settings path. 

Based on other discussions we have deemed including a reference implementation out of scope for this OEP.  That does not preclude it from existing or being maintained by edX or the community.  Just simply not something we felt was necessary when managing edx IDAs as containers.

arbrandes:
@bderusha, @felipemontoya: for the record, at OpenCraft we're still concerned about the absence of wording in this OEP regarding a reference implementation, and would welcome further discussion on it before the merge.  I created a [thread](https://discuss.openedx.org/t/oep-45-and-replacing-edx-configuration-with-another-refence-deployment-strategy/2269) for the purpose in the architecture forum.

ZurMaD:
Yes, please we need new strategy of deployment.

oeps/oep-0045-arch-ops-and-config.rst
```diff
@@ -44,6 +44,11 @@ Wherever possible, operating concerns will live within the codebase they are con
 * An operations manual documenting standard operations required to run and maintain the IDA.
 * A changelog to document differences in settings, operating concerns, and application functionality from release to release.
 
+As code is moved to this new paradigm it is important to note that we strive to maintain or improve upon the state of operations without sacrificing features or visibility.  If operations code is removed from one place it must either have a home in some other public place better aligned with this OEP (EG package installation in Dockerfiles) or no longer be needed at all in this model (EG `conditional ansible commands`_ for installing packages on particular operating
+systems).
```
arbrandes, line 48: Sounds good to me!  Thanks for this! :+1:

nasthagiri:
@felipemontoya @bderusha As I was looking to reference this OEP in another communication thread, I noticed that this OEP is yet not available on Read-the-Docs since it hasn't been merged. Are we now at a place where we can merge this particular PR in its _provisional_ status - knowing fully well that another implementation-specific PR is pending?

felipemontoya:
I think we are ready to merge this and continue with the implementation specific discussion.

PR #144, opened 2020-04-16
Follow-on Implementation Strategy for OEP-45 

bderusha:
**[OEP-45](https://github.com/edx/open-edx-proposals/pull/143) is still under review.**

The implementation plan for this OEP will be important to how successful it is. While comments can be made at any time, please refrain from any deep discussions about how to achieve OEP-45 until we agree on what OEP-45 should be.

This PR will be updated to reflect the current status of OEP-45

### DIFF
--- a/oeps/oep-0045-arch-ops-and-config.rst
+++ b/oeps/oep-0045-arch-ops-and-config.rst
@@ -164,6 +164,27 @@ Kubernetes is an open source container orchestration platform pioneered by Googl
 Implementation Strategy
 =======================
 
-Discussion of implentation of this OEP will happen in a `separate Pull Request`_ .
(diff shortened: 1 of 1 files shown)


sarina:
@bderusha is no longer at edX, so I'm closing this PR.

@kdmccormick is [OEP-45](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0045-arch-ops-and-config.html) one that we should update with the Tutor work?

edx-webhook:
Although this pull request is already closed, I've created [OSPR-6420](https://openedx.atlassian.net/browse/OSPR-6420) so that we can track it in Jira.

PR #146, opened 2020-04-28
OEP-46: Domain-Driven Design: Subdomains and Bounded Contexts

cpennington:
This serves to codify the Subdomains and Bounded Contexts that the Open edX system contains.

Note that it is currently aspirational, rather than descriptive. The current reality of the systems will likely not reflect this architecture, but we would like to move the system inline with this description (and keep this description updated as circumstances change).


davidjoy:
One thing that immediately jumps out at me is the lists of Bounded Contexts - I wonder to what extent edX product team input (and community input) would be useful in fleshing those out/vetting their contents.  How comprehensive should the lists in this OEP be?  Merely representative, or is it a definitive list of the contexts that exist?  If the latter, what supporting and generic Bounded Contexts are worth mentioning here?  The 'definitive' list of supporting/generic contexts likely has a pretty long tail, especially with respect to Open edX community instances.

I also wonder if a visualization/table/hierarchical representation would be useful for the OEP format... I found myself squinting at font sizes to figure out what level I was at, here and there.

cpennington:
@davidjoy The use cases were developed in concert with edX Product (@sstack22, in particular, I think). However, that was a year or two ago, .so it is definitely worth reviewing whether we have all our current use cases expressed, and whether the Open edX community has any more that they believe are important.

The goal is that the lists be roughly definitive, I think, although again, it may or may not be feasible to keep everything up to date as the system grows.

And yes, I am going to make a diagram at some point, I just haven't put it together yet

oeps/oep-0046-arch-ddd-domains.rst
```diff
@@ -0,0 +1,469 @@
...
+The Open edX system consists of the following bounded contexts.
+
+Terminology and Behaviors
+=========================
+Subdomain:
+  A group of related bounded contexts. A Subdomain is **Core** if it is a key differentiator for
+  Open edX, **Supporting** if it has Open edX specifics, but isn't **Core**, or **Generic** if
+  it has no Open edX business context.
+
+  There should be **no** synchronous communication between separate subdomains.
```
robrap, line 153: This is a strong statement. Do we want to add language around exceptions, or language like the manifesto uses regarding preference when making design decisions?

cpennington: I'm ok with adding exceptions, but I think it's important that we're making a pretty strong statement of preferences here about what these Subdomains actually imply for architectural decisions.

robrap: That makes sense. We may want to make a similar statement around not sharing code with business logic across subdomains, or something like that, but this might take more thought.

nasthagiri:
@cpennington FYI, I plan to review this PR tomorrow. I couldn't get to it earlier this week.

oeps/oep-0046-arch-ddd-domains.rst
```diff
@@ -0,0 +1,486 @@
...
+  - - Arbiter
+    - <Arbiter's real name and email address>
+  - - Status
+    - Draft
+  - - Type
+    - Architecture
+  - - Created
+    - 2020-04-23
+  - - Review Period
+    - 2020-04-23 - 2020-05-30
```
robrap, line 25: It is confusing whether this is in review or not.  If not, do you want to clear this field for now?  Thank you.

kdmccormick:
A couple of questions that I thought about during the study group:
1) I think (but am not sure) that edX may want to offer other university-administered degrees one day. I imagine that they'd end up managed handled through Registrar, like Master's degrees are. Is it worth naming Master's something more generic (e.g., B2U, Accredited, Degrees, etc.), or is it fine leaving it specific for now?
2) Where does Proctoring go? Maybe, like Learner Credentials, it's a supporting context in the Learning subdomain? FWIW, it's owned by the Master's & MicroMasters squad currently.
3) Program Credentials is used by all program types *except* Master's :stuck_out_tongue_closed_eyes: It might be better part of B2C.

oeps/oep-0046-arch-ddd-domains.rst
```diff
@@ -0,0 +1,327 @@
...
+Users:
+  - Course teams
+  - Instructional designers
+
+
+Bounded Contexts
+~~~~~~~~~~~~~~~~
+
+.. list-table::
+   :widths: 10 10 70
```
robrap, line 112: Add `:header-rows: 1`

oeps/oep-0046-arch-ddd-domains.rst
```diff
@@ -0,0 +1,327 @@
...
+     - Description
+   - - Course Run Authoring
+     -	supporting
+     - Course Run Authoring (supporting) allows educators to launch a Course Run. This includes the ability to override configuration settings and fields, including dates and grading policy.
+   - - Learning Content Authoring
+     - core
+     - Learning Content Authoring (core) allows educators to create and publish learning content and pathways,  which can be utilized to create course runs.
+   - - Learning Content Repository
+     - supporting
+     - Learning Content Repository (supporting) provides a shared service for storing, versioning, sharing, and discovering published learning content.  The learning content is available at varying degrees of granularity - including Pathways, Blocks and Units.
```
robrap, line 125: I'd recommend either dropping the word "shared", or helping explain why it is used here and not for other services.

oeps/oep-0046-arch-ddd-domains.rst
```diff
@@ -0,0 +1,327 @@
...
+Ubiquitous Language:
+  The set of terms and definitions that defines the entities and relationships within a Bounded Context. These terms may not have the same definitions across different Bounded Contexts.
+
+Anti-Corruption Layer:
+  A layer around the edge of a Bounded Context that converts entities and information retrieved from other Bounded Contexts in the same Subdomain into the Ubiquitous Language of this Bounded Context.
+
+Overall Behavior Flows
+======================
+
+A **learner** engages with a **Business Subdomain** to find learning content that meets their learning goals. They purchase that learning content via whatever mode is appropriate for that **Business Subdomain** (direct purchase, free enrollment, subscriptions, etc). The **Business Subdomain** requests fulfillment from the **Learner Subdomain** to allow the **learner** access to their selected learning content.
```
robrap, line 79: * I think having "Business Subdomain" (or "Business Subdomains") in the picture would help.
* If it is too much for one picture, maybe add a high-level picture without bounded contexts?

nasthagiri:
Note: This PR is planned to go through a large overhaul. Please hold off on any detail review until this PR is open for review.

sarina:
Closing this PR since Nimisha and Cale have left edX.

If anyone would like to pick up this work, please open a new PR for review.

edx-webhook:
Although this pull request is already closed, I've created [OSPR-6421](https://openedx.atlassian.net/browse/OSPR-6421) so that we can track it in Jira.

PR #149, opened 2020-05-08
OEP-47: Semantic Versioning

robrap:
Note: This OEP on Semantic Versioning was split out from https://github.com/edx/open-edx-proposals/pull/140

### DIFF
--- a/oeps/oep-0047-bp-semantic-versioning.rst
+++ b/oeps/oep-0047-bp-semantic-versioning.rst
@@ -1,0 +1,107 @@
+###########################
+OEP-47: Semantic Versioning
+###########################
+
+.. list-table::
+
+   * - OEP
+     - :doc:`OEP-0047 </oeps/oep-0047>`
+   * - Title
+     - Semantic Versioning
+   * - Last Modified
+     - 2020-05-28
+   * - Authors
+     - Robert Raposa <rraposa@edx.org>
+   * - Arbiter
+     - Régis Behmo <regis@behmo.com>
+   * - Status
+     - Accepted
+   * - Type
+     - Best Practice
+   * - Created
+     - 2020-03-19
+   * - Review Period
+     - 2020-05-11 - 2020-05-25
+
+.. contents::
+   :local:
+   :depth: 3
+
+Summary
+=======
+
+For any Open edX repository that publishes versioned packages, this OEP recommends following the `PEP 440`_ compatible sections of `Semantic Versioning v2`_.
+
+The most relevant part from the summary of the `Semantic Versioning v2`_ specification:
+
+    Given a version number MAJOR.MINOR.PATCH, increment the:
+
+    * *MAJOR* version when you make *incompatible API changes*,
+    * *MINOR* version when you *add functionality in a backwards compatible manner*, and
+    * *PATCH* version when you make *backwards compatible bug fixes*.
+
+From `PEP 440`_ (quote updated to match 2.0.0 of the spec):
+
+    The "Major.Minor.Patch" (described in this PEP as "major.minor.micro") aspects of semantic versioning (clauses 1-8 in the 2.0.0 specification) are fully compatible with the version scheme defined in this PEP, and abiding by these aspects is encouraged.
+
+Some additional tips are called out in:
+
+* `Backward Incompatible Changes`_
+* `Tooling and Automation`_
+* `Versions for Forks`_
+
+.. note::
+
+    This OEP does not apply to repositories without versions. For example, services like edx-platform, ecommerce, etc. do not have versions.
+
+.. _PEP 440: https://www.python.org/dev/peps/pep-0440/#semantic-versioning
+
+Context
+=======
+
+Many or all of the Open edX installable libraries already use versions similar to the `Semantic Versioning v2`_ specification. This OEP simply codifies this best practice, and provides additional guidance.
+
+Backward Incompatible Changes
+=============================
+
+Special attention should be given to incompatible changes, because:
+
+* Communicating incompatibilities alerts humans to possible problems when attempting to upgrade your package.
+* A MAJOR version is what is used to communicate any "Backward Incompatible Changes" (also known as "Breaking Changes").
+
+See related FAQ questions from `Semantic Versioning v2`_:
+
+* `If even the tiniest backwards incompatible changes to the public API require a major version bump, won’t I end up at version 42.0.0 very rapidly?`_
+* `What do I do if I accidentally release a backwards incompatible change as a minor version?`_
+* `What if I inadvertently alter the public API in a way that is not compliant with the version number change (i.e. the code incorrectly introduces a major breaking change in a patch release)?`_
+
+.. _If even the tiniest backwards incompatible changes to the public API require a major version bump, won’t I end up at version 42.0.0 very rapidly?: https://semver.org/#if-even-the-tiniest-backwards-incompatible-changes-to-the-public-api-require-a-major-version-bump-wont-i-end-up-at-version-4200-very-rapidly
+.. _What do I do if I accidentally release a backwards incompatible change as a minor version?: https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version
+.. _What if I inadvertently alter the public API in a way that is not compliant with the version number change (i.e. the code incorrectly introduces a major breaking change in a patch release)?:
+
+
+Tooling and Automation
+======================
+
+* A checklist item in a pull-request template can remind people of any manual steps.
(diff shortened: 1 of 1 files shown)


oeps/oep-0047-bp-semantic-versioning.rst
```diff
@@ -0,0 +1,90 @@
...
+--------------------------------------------------
+
+* Communicating incompatibilities is important because it alerts humans to possible problems when attempting to upgrade your package.
+* A MAJOR version is used to communicate any "Backward Incompatible Changes" (also known as "Breaking Changes").
+* If you think a breaking change is unlikely to affect anyone, increment the MAJOR version anyway, and simply note that fact in the changelog entry.
+
+What if I find an incompatible change after it is released as a MINOR or PATCH version?
+---------------------------------------------------------------------------------------
+
+If you later find a backward incompatible change in a MINOR or PATCH version after publishing the package, please update the changelog and communicate the inconsistency. Additionally, you could do one of the following:
```
timmc-edx, line 67: Another alternative would be to re-release as a major version and mark the offending minor/patch release as yanked. (If it is discovered after several minor or patch releases, this would involve yanking and renumbering several versions, in which case it might not be worth it.)

oeps/oep-0047-bp-semantic-versioning.rst
```diff
@@ -0,0 +1,90 @@
...
+     - 2020-05-11 - 2020-05-25
+
+.. contents::
+   :local:
+   :depth: 3
+
+Summary
+=======
+
+For any Open edX repository that publishes versioned packages, this OEP recommends following the `Semantic Versioning v2`_.
```
timmc-edx, line 33: We sometimes maintain forks of other projects. Should we include something about versioning those? (Stick with their scheme, use our own, use a `X.Y.Z+edx1` build indicator, etc.)

jmbowman: Adding a `+edx.1` suffix to the version number is now the standard way to handle this, after the Python community considered all the options: https://www.python.org/dev/peps/pep-0440/#local-version-identifiers .

oeps/oep-0047-bp-semantic-versioning.rst
timmc-edx: I think this is the part of the OEP that is likely to need the most discussion. It feels a little underspecified at the moment, but I'm not sure exactly what would need to change to improve it.

adzuci: Could you link to an example of yanking a release in a semver codebase?

I usually look for the highest minor number of a given major release and am not familiar with yanking releases, so my preference would be to prefer reverting and bumping the minor version. 

This seems very similar to the "What do I do if I accidentally release a backwards incompatible change as a minor version?" FAQ in https://semver.org/spec/v2.0.0-rc.2.html and I almost prefer to keep the explicit don't modify old releases clause:
"Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.", but if your suggested yanking process is meant to "document the offending version" then I'd be interested to know more.

oeps/oep-0047-bp-semantic-versioning.rst
```diff
@@ -0,0 +1,90 @@
...
+Some additional tips are called out in `Backward Incompatible Changes`_ and `Tooling and Automation`_.
+
+.. note::
+
+    This OEP does not apply to repositories without versions. For example, services like edx-platform, ecommerce, etc. do not have versions.
+
+Context
+=======
+
+Many or all of the Open edX installable libraries already use versions similar to the `Semantic Versioning v2`_ specification. This OEP simply codifies this best practice, and provides additional guidance.
```
adzuci, line 52: Is https://github.com/edx/edx-enterprise/releases or https://github.com/edx/edx-ora2/releases a good example of this and are there some libraries that partially follow this OEP but would need to change their release workflow to comply with it?

oeps/oep-0047-bp-semantic-versioning.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The most relevant part from the summary of the `Semantic Versioning v2`_ specification:
+
+    Given a version number MAJOR.MINOR.PATCH, increment the:
+
+    * *MAJOR* version when you make *incompatible API changes*,
+    * *MINOR* version when you *add functionality in a backwards compatible manner*, and
+    * *PATCH* version when you make *backwards compatible bug fixes*.
+
+Some additional tips are called out in `Backward Incompatible Changes`_ and `Tooling and Automation`_.
```
adzuci, line 43: semver v2 does not explicitly state if you should use pre-releases or not and I'm not sure if we currently use them in our libraries.  Is that something we want to provide guidance on?
https://www.python.org/dev/peps/pep-0440/#pre-releases

oeps/oep-0047-bp-semantic-versioning.rst
timmc-edx: Yanking a release would mean marking it as "YANKED" in the changelog, mostly—[Keep a Changelog](https://keepachangelog.com) promotes this terminology. I don't know if there's a way to suppress a release in PyPI and still make it available for use if pinned.

I wouldn't want to modify any existing release.

robrap: 1. Personally, I was trying to provide some hints without being prescriptive.  I'd rather drop this whole section than be concerned with anything controversial.  This OEP is not meant to be controversial.  
2. I feel like yanking releases can probably have its own pitfalls that I want to avoid, and I don't know that to be a common practice here, so I don't really want to add that as a recommendation.
3. My recommendations do not involve modifying any existing release.  Only adding new releases.  Is that unclear?

oeps/oep-0047-bp-semantic-versioning.rst
robrap: This isn't about changing anyone's workflow.  It is mostly about knowing when to increment X, Y or Z in X.Y.X.  Some people use it more like LARGE.MEDIUM.SMALL in terms of size of change, and that isn't following the semantic versioning.

oeps/oep-0047-bp-semantic-versioning.rst
robrap: I haven't heard of us doing this.  The intent of this OEP is not to introduce terms or process that is unfamiliar and not already used, but simply to be the "semantic" back into "semantic versioning".  I mostly want a simple OEP to clear up MAJOR vs MINOR vs PATCH.

oeps/oep-0047-bp-semantic-versioning.rst
feanil: Pulling things off of PyPI is risky and complicated, I think sticking with just the recommendation is fine for now.  We can always iterate or further improvements if we're actually running into this situation enough to warrant more clarity.

oeps/oep-0047-bp-semantic-versioning.rst
robrap: Added a section.

oeps/oep-0047-bp-semantic-versioning.rst
timmc-edx: Ignore my comment re: PyPI—I'm just not familiar with what mechanisms there are in the Python ecosystem for marking a release version as "don't use this!". If there aren't any, that's fine! I'm definitely not suggesting retroactively modifying or removing releases *from PyPI*.

The reason I bring this all up is that we recently had a situation where we accidentally introduced a breaking change in a patch or minor release: https://github.com/edx/edx-drf-extensions/releases/tag/5.0.0

I'm fine if we want to just iterate on this and leave that part unspecified for now, but the fact that we've run into it recently (and didn't already have a shared understanding of what to do!) suggests that we could benefit from figuring out what we want.

(If we do decide to deviate from [semver.org's prescription](https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version), we should at least document that here.)

robrap: My intention was to draw attention to this topic, not to go against their recommendations.  I updated the OEP to point to the relevant FAQ questions.

oeps/oep-0047-bp-semantic-versioning.rst
```diff
@@ -0,0 +1,90 @@
...
+* Other tools can be added here over time.
+
+.. _semantic-release: https://github.com/semantic-release/semantic-release
+
+References
+==========
+
+* `Semantic Versioning v2`_
+
+.. _Semantic Versioning v2: https://semver.org/spec/v2.0.0.html
```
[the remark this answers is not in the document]
robrap: I personally don't feel it is necessary to mention the alternatives, but I did scope it down and added a quote from the PEP.

PR #154, opened 2020-06-01
[ARCHBOM-1206] Modifying OEP2: Make owner key optional

jinder1s:
See for more details: https://openedx.atlassian.net/browse/ARCHBOM-1206

Originally, the intention was to openedx.yaml as the source of truth for ownership info. However, as the ownership initiative has progressed, the [ownership google sheet](https://docs.google.com/spreadsheets/d/1qpWfbPYLSaE_deaumWSEZfz91CshWd3v3B7xhOk5M4U/edit#gid=840878297) has become the source. It was decided asking everyone to switch to openedx.yaml as source of truth is not necessary in the short term and would distract from other things that need to be done for the ownership effort. 

The change in this PR is meant to align OEP-2 with current plans for ownership.

### DIFF
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -44,7 +44,7 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
 
 ``owner``: dictionary (optional)
     This key contains information about the assigned owner of this repository.
-    **Note:** Initially, the key was intended as info about ownership and about who to contact for repo related issues. Ownership info has now been moved to an edx internal location and we plan on adding contact point info to repos at some point in the future. The exact location still to be determined, possible in openedx.yaml or CODEOWNERS file.
+    **Note:** Initially, the key served two purposes: assigning ownership of the repository and as a who to contact for repo related issues. Ownership info has now been moved to an edx internal location. As for contact info, this is still WIP. We plan on adding contact point info to repos at some point in the future. The exact location is still to be determined, possible in openedx.yaml or CODEOWNERS file.
 
     ``type``: string (optional)
         The value of this key must be either ``team`` or ``repo``. It indicates which ownership model applies to this repository, and therefore which of these two keys should exist with a non-empty value.
--- a/oeps/oep-0002-bp-repo-metadata.rst
+++ b/oeps/oep-0002-bp-repo-metadata.rst
@@ -44,7 +44,7 @@ Each repo will include a file ``openedx.yaml``, with the following keys:
 
 ``owner``: dictionary (optional)
     This key contains information about the assigned owner of this repository.
-    **Note:** Initially, the key served two purposes: assigning ownership of the repository and as a who to contact for repo related issues. Ownership info has now been moved to an edx internal location. As for contact info, this is still WIP. We plan on adding contact point info to repos at some point in the future. The exact location is still to be determined, possible in openedx.yaml or CODEOWNERS file.
+    **Note:** Initially, the key served two purposes: assigning ownership of the repository and as a "who to contact" for repo related issues. Ownership info has now been moved to an edx internal location. As for contact info, this is still WIP. We plan on adding contact point to repos at some point in the future. The exact location is still to be determined, possible in openedx.yaml or CODEOWNERS file.
 
     ``type``: string (optional)
         The value of this key must be either ``team`` or ``repo``. It indicates which ownership model applies to this repository, and therefore which of these two keys should exist with a non-empty value.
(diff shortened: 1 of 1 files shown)


cpennington:
I assume we're not actually making ownership optional... in which case, maybe it makes sense to reference the official central point of ownership mapping somewhere?

nedbat:
Definitely some explanation is in order.

jinder1s:
The plan is to have Ownership spreadsheet in google drive be the source of truth for ownership. Since we no longer plan on adding any tooling to sync ownership info between spreadsheet and openedx.yaml, requiring owner key in OEP-2 no longer seems necessary. The sheet: https://docs.google.com/spreadsheets/d/1qpWfbPYLSaE_deaumWSEZfz91CshWd3v3B7xhOk5M4U/edit#gid=840878297

jinder1s:
I've updated the description of this PR with necessary context. Apologize for assuming a limited audience.

regisb:
For information, this spreadsheet is unaccessible by users outside edX.

cpennington:
Since ownership implies things about who you should go talk to if you have an issue w/ code, I feel concerned about moving that information out of the repo. I understand wanting a centralized place to go look at all the ownership info, but I'm concerned about removing that information from repositories.

nasthagiri:
Based on our recent experiences with rolling out technical ownership thus far:

1. Squads own much more than code repositories. They own apps within edx-platform. They own external services. They own cross-cutting concerns that may not map directly to code.
2. All of the above can be relatively easily maintained in a spreadsheet.
3. Downstream tooling (JIRA integrations, etc) benefit from a single place to access ownership info.
4. Squads would like a single pane of glass of all the things they own and need to manage.
5. Different squads have different preferences for team-privacy and team-communications.

Given the above learnings thus far, we are proposing to pivot away from having the source-of-truth for ownership be within the code itself.  We believe that having ownership information co-located with the code is nice-to-have, but not necessary.  It is more important to focus on the clarity of ownership assignments for the time being.  In the future, we may choose to push/synch ownership information from a central place to individual code repos - but for now, we don't see the immediate value of that effort.

nasthagiri:
We are also finding that automation eliminates the need to manually update openedx.yaml files. We have recently implemented a repo dashboard that supports automated checks for computing and assessing compliance and measuring code health. So while we are not proposing a more provocative standpoint as part of this PR, we may determine that ongoing maintenance of updating OEP status in openedx.yaml files may not be worthwhile. And if that's the case, the only valuable field left in these yaml files is `openedx-release`.

nasthagiri:
@cpennington Regarding _"who you should go talk to if you have an issue w/ code"_, this is a fair question.  I do believe this can be orthogonal to which squad owns a particular repository. We can have another field or a README convention with information on how to open issues in a repo.

Some repos have a convention of opening GitHub issues. Some others may require opening JIRA tickets or a PR for discussion.

cpennington:
> ongoing maintenance of updating OEP status in openedx.yaml files may not be worthwhile

Yeah, OEP-2 specifically mentioned that feature-specific tags should only go into `openedx.yaml` if they weren't automatically detectable. The oep2-checker (https://github.com/edx/repo-tools/tree/master/edx_repo_tools/oep2) specifically implemented automated checking for particular things across all repos, although I never got enough time to make the dashboard reliable (cc @jmbowman who just turned the daily run off after it failed for 3 years straight).

As long as we have a good way to for any community member (not just edx employees) to find out who to contact with questions/issues on a repo, I'm ok with moving the ownership into a central location (although I do have a bit of a lingering concern that that central location will end up out of sync w/ whatever `README`s/`CODEOWNERS` we put in place in the code).

If it were solely my decision, I'd probably put ownership info into the repos in standard locations (openedx.yaml files in subdirectories? CODEOWNERS as a standard location?), and then pus automation to pull all those into a central read-only spreadsheet with backlinks to provide the single pane of glass. However, in this case, I'm not the one that's going to be doing the work, so my opinions are suggestions, not blockers.

cpennington:
Do we have a plan/timeline for addressing the community need for points of communication?

cpennington:
I approved, but I also think it's probably worth writing some words about future plans in the OEP to make it clear what's replacing the owner keys.

nasthagiri:
> it's probably worth writing some words about future plans in the OEP to make it clear what's replacing the owner keys

I agree. @jinder1s Would you be able to do so?  Or should @feanil or I take a stab at it?

jinder1s:
@nasthagiri I think you or feanil should take a stab at it. You will be able to provide a better idea of the future plans than I.

nedbat:
@jinder1s Good news: there are many words in this PR discussion that can serve as source material for explanation in the OEP why this key is optional.

jinder1s:
Okay, I added a note explaining that we plan on getting back to adding contact information in the future. In writing it, I choose to be preetty concise, let me know if more details are necessary.

PR #155, opened 2020-06-10
An initial draft at what a handbook entry might look like.

feanil:


handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+Documentation can live in lot's of places, we want to be clear about which documentation lives where most of the time so that finding docs is easier.  In general, keep the documents as close to the thing that they are describing as possible.
+
+Details
+****
+
+* READMEs, HOW-TOs, API Docs and ADRs should live with the code.
+
+  * API docs should be in swagger format.
+
+  * **Everything else should be RSTs**.  Yes, you already know markdown but RST is more powerful and robust and that is what we have chosen to use for consistency across the whole platform.  This_ is a good resources for learning RST.
```
robrap, line 17: * Maybe something like?:
```
* **Everything else should be RST format**.  `Learn about using RST`_.
```
* I think the sentence "`Yes, you...whole platform.`" should either be dropped, or maybe add something like "`Learn about why RST`" that could point to the right place in the OEP.  Another option is to provide a section of this doc with links to relevant decisions.

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+
+* Process docs such as scrum notes, retros, personal notes should not live in the repo and should generally be in confluence.
+
+* Sometimes documents span repos, if a document is relevant to multiple repos that are related to each other, pick the most sensible of the repose to put the doc in.  Link all the relevant repos in the doc and link back from all the repos to the doc to make it easier to find in either direction.
```
robrap, line 21: Possible alternative (with typo fix):
```
If a document is relevant to multiple repos, pick the most sensible repo to put the doc in, and provide back links from the other repos.
```

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+
+* If a document is relevant to everyone, it should be in the OEP_ repo which is where we document things that are relevant to all of  openedx.  Things like best practices for dependency management, where we should put different kinds of documentation, etc.
```
robrap, line 23: 1. Possible alternative.  Note: The alternatives could be bullets in the overview potentially, with the details about READMEs, etc. in the rest of this doc.
```
* Documents relevant to all of Open edX should live in the OEP_ repo.
```
2. I'm not sure how to parse the second sentence and its intent.

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+
+
+All the details
```
robrap, line 26: Again, I'd add details to this doc and provide links to answer why.

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
+####
+Developer Documentation
+####
+
+Overview
+********
+
+Documentation can live in lot's of places, we want to be clear about which documentation lives where most of the time so that finding docs is easier.  In general, keep the documents as close to the thing that they are describing as possible.
```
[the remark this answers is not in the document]
antoviaque: +1 for brievety.

Although it still helps when there is a natural flow to the document's writing - a series of rules in bullet points can help building a shorter reference, but they can also make reading it pretty dry. Compare for example [this section](https://handbook.opencraft.com/en/latest/roles/#general) of the OpenCraft handbook with [this one](https://handbook.opencraft.com/en/latest/organization/)

handbook/documentation.rst
antoviaque: Also, if RST is the format for everything in the platform, aren't README, HOWTOs, etc. living in the code also using RST? Ie recommendations about format might be broader than ones about location.

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+Details
+****
+
+* READMEs, HOW-TOs, API Docs and ADRs should live with the code.
+
+  * API docs should be in swagger format.
+
+  * **Everything else should be RSTs**.  Yes, you already know markdown but RST is more powerful and robust and that is what we have chosen to use for consistency across the whole platform.  This_ is a good resources for learning RST.
+
+* Process docs such as scrum notes, retros, personal notes should not live in the repo and should generally be in confluence.
```
antoviaque, line 19: Conversely, are we recommending that Confluence should only store process docs? It could be worth defining further what "process docs" mean actually, as I assume we don't include documentation that describe the processes themselves?

handbook/documentation.rst
antoviaque: "Everyone" or "all of Open edX" might be a bit too large of a requirement, there will be broad decisions that still aren't interesting or applied to _everyone_? Maybe something like "Documents relevant to a large part of Open edX"?

feanil:
@robrap I made the changes that you and @antoviaque suggested.  I still need to add more backlining to the OEP but wanted to discuss the level of detail in this document.  My thinking was that this doc would tell you what you need to know on day 1 and if you actually need to write one of these types of documents you would go to the OEP?

I think it would make sense to link to the OEP sections but not necessarily put a lot more words here. What do you think?  Is there something specific you were envisioning in here that you think everyone should now as high-level guidance?

@davidjoy @antoviaque I'd love to know your thoughts on the way I'm thinking about this.

antoviaque:
@feanil I'm fine with either approach on this, but for things that people need to know on day 1, I tend to think the onboarding course might be the right place for that - ie something meant to be read at that time, and only once to get a high level overview. The handbook is more of a reference manual, ie the things that we refer to when we want to check the detail of how something works. While the OEP might be the diff, ie the rationale, justification and support for the discussion on a specific change, which then results in a change in the handbook, stripped of the historical context and discussions?

davidjoy:
Having just read through, it feels short and to the point, which seems appropriate for a "day one" sort of overview.  The links to more detail help people dive further if they need to.

robrap:
@feanil: Based on the conversation, I see 3 types of information being discussed:
1. Intro/Overview/Summary of reference documentation.
2. Details of reference documentation.
3. Justifications of decisions made in the reference documentation.

I'd really like to see 2. and 3. separated from each other, so when someone does need to learn the details, they don't run into the issue they have with our current OEPs.  The Documentation OEP isn't a good example of the problem here, because it is heavy on reference documentation and light on justification, so it doesn't waste too much time sorting through when you want the details.  But, that is not true for other OEPs.

My preference would be that 1. and 2. be put in one and the same document, but that 1. should be the summary at the top which makes it simple to complete and determine whether or not you want to read the rest of the doc.  That would mean that what you are calling "Overview + Details", I would make a single "Overview" section.  I would then add the details lower in the same document.

Another proposal would be to move the details to a sub-document, so your document continues to feel short and quickly digestible, but that the details are part of the handbook and separated from the justifications, as @antoviaque had mentioned.

feanil:
I like the clarity of the 3 different sections.  What would we think of OEPs having 3 parts, a quick summary doc, a detail doc and then a rational doc?  I like having the 3 docs because visually I think it would harder for people to just read the top of a doc if there was more too it but putting the details in a second doc gives users the option to jump into the details or not.  If others think this would be useful, I can try to pull out the detail for this doc and maybe create an overvew and detail file for a second item that is an oep with a lot of rationale.(Let me know if you think there are candidates.)

handbook/documentation.rst
```diff
@@ -0,0 +1,33 @@
...
+Details
+****
+
+* READMEs, HOW-TOs, API Docs and ADRs should live with the code.
+
+  * API docs should be in swagger format.
+
+* **Everything else should be RSTs**.  This_ is a good resources for learning RST.
+
+* Process docs such as scrum notes, retros, personal notes should not live in the repo and should generally be in confluence.
```
staubina, line 19: Process docs should live where the team is comfortable with them. We should avoid applying restrictions on how teams function.

staubina: Document editors in Confluence are not great and are challenging for asynchronous and team-based editing activities.

staubina:
What is the goal of a "Handbook Entry"?  Is the goal documented elsewhere?

handbook/documentation.rst
feanil: @staubina that's true which is why this is a `should` instead of a `must`.  We want to make sure there is a clear default if you aren't sure of what you should be doing.

feanil:
@staubina this is a draft PR to try to figure out what a handbook entry might be and whether it would be a useful tool. Which is to say the rational is not documented yet because we're trying to figure out the details.  In general the thinking is that OEPs mix multiple types of docs into one big doc(decision, detail, rationale).  We're trying to see if separating these three things into 3 documents makes the documentation more useful and easier to consume. ie. if you just want the agreed upon decision, you can read a few paragraphs, if you want to know more about how to do a thing you can read the details and if you want to know why we're doing it that way instead of some other way, you can read the rationale.

feanil:
Hey all, closing this for now because I haven't been able to get back to it for 3 months.  If someone want's to pick it up let me know but don't want to leave this tendril hanging around confusing people.

PR #160, opened 2020-09-14
Provide guidance for archiving forks multiple times.

feanil:
There are a lot of options for this but choosing the simplest for now
since this is a pretty corner case and it's unclear how often this will
happen let alone how often it will bite us to not have a more complex
process here.

Currently there are only 5 forks under edx-unsupported that could have
collisions, 4 of which we are unlikely to fork again.

### DIFF
--- a/oeps/oep-0014-proc-archive-repos.rst
+++ b/oeps/oep-0014-proc-archive-repos.rst
@@ -208,3 +208,8 @@ Change History
 ----------
 
 * Decide to use the new edx-unsupported org for all archived repos. Old way we were doing things is now recorded as Alternative 1: Archive In Place.
+
+2020-09-15
+----------
+
+* Updated to provide more details around archiving the same fork multiple times.
--- a/oeps/oep-0014-proc-archive-repos.rst
+++ b/oeps/oep-0014-proc-archive-repos.rst
@@ -21,11 +21,13 @@ OEP-14: Archiving edX GitHub Repositories
 | Created       | 2017-01-09                                               |
 +---------------+----------------------------------------------------------+
 | `References`  | `ORA PR Discussion`_,                                    |
-|               | `Initial Archiving Discussions`_                         |
+|               | `Initial Archiving Discussions`_,                        |
+|               | `Moving More Repos to edx-unsupported`_                  |
 +---------------+----------------------------------------------------------+
 
 .. _ORA PR Discussion: https://github.com/edx/edx-ora/pull/187
 .. _Initial Archiving Discussions: https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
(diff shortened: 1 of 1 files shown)


oeps/oep-0014-proc-archive-repos.rst
```diff
@@ -93,6 +95,11 @@ These steps should be followed for all repos within the edX organization(forks i
 
    - Archive the repository per `GitHub's archive process`_
 
+.. note::
+    Over the lifetime of Open edX, we may fork the same external open source repository multiple times.  In this case, we may need to archive the fork multiple times as we move between our fork and following upstream.  When this is necessary, delete the old copy of the fork before you move the new repo to edx-unsupported.
+
+    This may break some older version of Open edX.  Some combination of copying branches between forks, renaming branches, and changing unsupported versions of Open edX would have to be done to keep things working.  We opt not to take on this extra work by default thought may do so under extenuating circumstances.
+
```
nasthagiri, line 102: Nit: `thought` -> `though`

nasthagiri, line 102: One "extenuating circumstance" that we may want to call out is: the previous fork is still being used by the previous Open edX named release. Since we are supporting 1-previous named release, we need to ensure the previous release continues to function.  In that case, one idea is to move the previous fork to the [openedx github organization](http://github.com/openedx) and rename it for easier discovery (such as `<juniper>-fork-<repo-name>`).

feanil: I'll call that out,  just moving the old fork is not sufficient because a lot of the redirects will break but I'll add a note about it and suggest porting the branches of the old fork to the new fork.

feanil:
@nasthagiri ready for another round of review.

oeps/oep-0014-proc-archive-repos.rst
nedbat: "old for." -> "old fork."

nedbat:
Wouldn't the simpler approach be to un-archive the existing fork, and keep working in it? Then old references continue to work, I would think?

feanil:
I can recommend that but was concerned with people actually remembering to follow that process.  I'm imagining that people will have already made a new fork by the time they realize that we've done this before.  I'll make a note of it as a possibility though I don't think it should be a process people should be expected to follow since we don't have any guardrails to prevent the wrong thing from happening.

feanil:
Updated again.

nasthagiri (approved):

oeps/oep-0014-proc-archive-repos.rst
timmc-edx: Late thought: What if when we made forks, we named them after the upcoming release right away? It might be easier to remember at that moment than at archiving time, avoid name conflicts, and make it clearer which ones were safe to delete.

PR #161, opened 2020-09-16
Add a note about archiving from other orgs.

feanil:


oeps/oep-0014-proc-archive-repos.rst
```diff
@@ -93,6 +93,8 @@ These steps should be followed for all repos within the edX organization(forks i
 
    - Move the repository to the edx-unsupported organization
 
+      - If the repo is not coming from the `edx github org`_ then prefix then rename the repo to have source org name before moving. For example the ``notifier`` repo in the ``edx-solutions`` org wolud be renamed to ``edx-solutions-notifier`` before moving.
+
```
nasthagiri, line 97: `then prefix then rename the repo to have source org name before moving`
 to
`then before moving it, rename it with a prefix of the source org's name`

nasthagiri (approved):

PR #162, opened 2020-09-29
OEP-26: Fix read-the-doc links for xAPI/Caliper

nasthagiri:
Not knowing how to direct sphinx + read-the-docs to the correct document, this PR fixes broken links by updating the links to point to Github links directly.


nedbat:
It doesn't seem great to me to hardcode links to GitHub.  Are you open to making the links work properly for readthedocs?

nasthagiri:
Yes, very much open to it!
Do you have guidance on how to do so?

nasthagiri:
I'm also open to iterating on this.
I can merge the PR as is in order to fix the current broken links. 
We can then iterate later to do the more proper fix.

PR #180, opened 2021-01-04
A draft of Change Transparency

nedbat:
It's not clear this will become an OEP, but I needed someplace to write
this all down.

### DIFF
--- a/oeps/change-transparency.rst
+++ b/oeps/change-transparency.rst
@@ -2,7 +2,7 @@
 Change Transparency
 ###################
 
-TODO: This needs to be given an OEP number if it remains as an OEP/
+TODO: This needs to be given an OEP number if it remains as an OEP.
 
 Abstract
 ********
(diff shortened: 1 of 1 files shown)


antoviaque:
@nedbat Great to see this! :D Definitely +1 :+1: 

Is it related to https://github.com/edx/open-edx-proposals/issues/171 ?

nedbat:
@antoviaque yes, in the discussion on the PR itself (https://github.com/edx/edx-platform/pull/25949) we were talking about wanting to record the motivations for the information collected in the PR template.

feanil:
@nedbat do you think we still need this high-level OEP around change transparency?

sarina:
@nedbat the PR template, https://raw.githubusercontent.com/openedx/edx-platform/master/.github/pull_request_template.md, references this PR. Is that still the right thing to do?

nedbat:
@sarina You are right that we were too optimistic.... I'll update the template.

PR #181, opened 2021-01-29
OEP-0049 Django App Patterns

tuchfarber:


oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,165 @@
...
++++++++++++++
+.. Reviewer NOTE: I don't feel like I entirely understand models_api.py. My main issue is I don't understand the benefit of the additional layer versus api.py just calling ``Thing.objects.get()``.
+
+
+.. _rest_api:
+rest_api/
++++++++++
+Each app should self-contain it's related REST API, meaning it should live within the app and not in a separate "api" app where it would mingle with other apps' REST API code. Since the Python API is using ``api.py``, the REST API code (often Django Rest Framework) should live in a ``rest_api`` folder.
+
+.. Reviewer NOTE: Do we want to detail what a standard ``rest_api`` looks like? For example having a serializers.py, permissions.py, versioned folders of APIs, etc. I'm not sure how much of that is common DRF and how much is Open edX.
```
tuchfarber, line 160: Just calling this note out because it doesn't show up if you're reviewing the render

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,165 @@
...
+.. _models_api.py:
+models_api.py
++++++++++++++
+.. Reviewer NOTE: I don't feel like I entirely understand models_api.py. My main issue is I don't understand the benefit of the additional layer versus api.py just calling ``Thing.objects.get()``.
+
+
+.. _rest_api:
+rest_api/
++++++++++
+Each app should self-contain it's related REST API, meaning it should live within the app and not in a separate "api" app where it would mingle with other apps' REST API code. Since the Python API is using ``api.py``, the REST API code (often Django Rest Framework) should live in a ``rest_api`` folder.
```
ormsbee, line 158: I'm not sure if it's worthwhile to deviate from having these in `views.py`. It's where people will expect to look for views code, including DRF views. And we're not really building any other kind of view.

Also, I suspect if we really go all-in on this pattern, we'll find that REST APIs do not align 1:1 with apps and their internal APIs. For instance, with `learning_sequences`, I hope that eventually many other REST APIs query it to get course outline and sequence metadata, marked up with whatever additional information makes sense for them. It's kind of like Netflix's backend-for-frontend, where instead of having a bunch of backend services that get aggregated into different REST APIs, we get a bunch of small in-process API calls that get composed.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,165 @@
...
+        supported_programs [
+          ProgramData(
+            uuid=program.uuid,
+            title=program.title,
+            status=program.status
+          )
+          for program in all_programs
+          if program.uuid not in UNSUPPORTED_PROGRAM_UUIDS
+        ]
+        return supported_programs
```
ormsbee, line 125: What are your thoughts on pagination for this kind of request? I know the paginators in DRF have some kind of ability to accept something that's not a queryset, but I don't remember what that looked like.

oeps/oep-0049-django-app-patterns.rst
tuchfarber: I've thought about this some more and I don't know if I necessarily agree with the serializers-in-view suggestion. For example: Discovery has APIs for course, course runs, and programs. Without the a shared CourseRunSerializer, we would end up with so much duplicate code in the program serializer reimplementing the course and course run serializers. I realize it's a flexibility vs DRY argument, but I feel we have too many situations where the flexible option is infeasible. 

I really like the `services.py` and having everything call into that w/ dumb models and dumb APIs. I'm not sure how much I want to lean into a paradigm shift versus a better/consistent way to things we already do. Evolutionary vs revolutionary and all that.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+   * - Arbiter
+     - Jinder Singh
+   * - Status
+     - Draft
+   * - Type
+     - Architecture
+   * - Created
+     - 2021-01-29
+   * - Review Period
+     - TBD
```
jinder1s, line 26:
```suggestion
     - TBD
```
     - 2021-02-22 - 2021-03-10

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+   * - Title
+     - Django App Patterns
+   * - Last Modified
+     - 2021-01-29
+   * - Authors
+     - Matt Tuchfarber <mtuchfarber@edx.org>
+   * - Arbiter
+     - Jinder Singh
+   * - Status
+     - Draft
```
jinder1s, line 20:
```suggestion
     - Under Review
```

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,165 @@
...
+          )
+          for program in all_programs
+          if program.uuid not in UNSUPPORTED_PROGRAM_UUIDS
+        ]
+        return supported_programs
+
+.. _data.py:
+data.py
++++++++
+This file should include the public data structures for the app that can be passed between apps without exposing internal features. These should be used instead of sending Django model objects or querysets to apps that call the functions in ``api.py``. This file should not import anything other than stdlib modules, so that it may be imported by any other app without issue. These data objects should be simple objects with all business logic handled by ``api.py``. They may however perform simple validation, as long as it is self-contained (doesn't reach out to database, network, or any code outside of the class)
```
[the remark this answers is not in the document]
ormsbee: I think it's slightly better to just recommend `attrs` as the default, so that people don't go calling methods that they expect to be present. But honestly, the shape of the API is so similar that the number of times it's going to bite people is probably pretty small, so I don't feel strongly about it.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+   * - Created
+     - 2021-01-29
+   * - Review Period
+     - 2021-02-22 - 2021-03-10
+
+Abstract
+--------
+Proposes a common set of code patterns for Open edX Django apps.
+
+Motiviation
```
doctoryes, line 32: `Motiviation` -> `Motivation`

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+
+.. code-block:: python
+
+  default_app_config = "service_name.apps.app_name.apps.CustomAppConfig"
+
+
+.. _apps.py:
+apps.py
++++++++
+The ``apps.py`` file should contain a subclass of a Django ``AppConfig``. The AppConfig should set the app's name to it's full path (e.g. ``name = "service_name.apps.app_name"``) and should (optionally) have an overriding ``ready()`` function which initializes the app. This initialization also often includes setting up Django signals.
```
doctoryes, line 66: Long ago, the standard used to be to just import the signal handlers at the top of the file, which caused many circular import dependencies - hence this new pattern. Not sure if the text should mention that, but history and all...

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+      def ready(self):
+          """
+          Connect handlers to recalculate grades.
+          """
+          from .signals import handlers
+
+.. _api.py:
+api.py
+++++++
+This should be single point of entry for other Python code to talk to your app. This is *not* a Rest API, this is a Python API (see rest_api_). Some rules for ``api.py`` are as follows:
```
doctoryes, line 89: 👍🏼 for differentiating between Python API and REST API here. Sometimes "API" is used loosely and interchangeably and it's confusing.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+APIs should be versioned and the serializers and permissions associated with that version should be kept inside that version's folder. This prevents breakages when an API needs to be updated.
+
+An example of a common folder structure for a versioned REST API::
+
+  app_name
+  ├── rest_api
+  │   ├── v1
+  │   │   ├── permissions.py
+  │   │   ├── serializers.py
+  │   │   ├── urls.py
```
doctoryes, line 197: Is there anything useful to state about standards for `urls.py` files? (This one and the two below.)

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+  │   │   └── views.py
+  │   └── urls.py
+  ├── urls.py
+  └── views.py  # existing legacy non-REST APIs
+
+
+.. _signals:
+signals/
++++++++++
+If an app is consuming Django Signals from other apps in the service, it should include a ``signals.py`` file which includes all of it's signal handlers.
```
doctoryes, line 207: `it's` -> `its`

doctoryes (commented):
Thanks for taking the time to write out these standards! 🧁  

Please ensure that these agreed-upon standards end up being represented in this cookiecutter upon merging this OEP:

https://github.com/edx/edx-cookiecutters/tree/master/cookiecutter-django-app

Because good work is always rewarded with more work. 😉

oeps/oep-0049-django-app-patterns.rst
ormsbee: I guess I haven't been paying enough attention–I knew the `rest_api` practice existed, but I didn't realize it was that widespread. FWIW, it's not something I feel strongly about, particularly since there's a `views.py` in that directory anyhow.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,165 @@
...
+          from .signals import handlers
+
+.. _api.py:
+api.py
+++++++
+This should be single point of entry for other Python code to talk to your app. This is *not* a Rest API, this is a Python API (see rest_api_). Some rules for ``api.py`` are as follows:
+
+1. API methods defined in ``api.py`` should be well-named, self-consistent, and relevant to its own domain (without exposing technical and implementation details)
+2. An app's Django models and other internal data structures should not be exposed via its Python APIs.
+3. Ideally, tests should use only Python APIs declared in other apps' ``api.py`` files. However, if an app's API is needed *only* for testing, then test-relevant Python APIs should be defined/exported in an intentional Python module called ``api_for_tests.py``.
```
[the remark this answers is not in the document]
ormsbee: Yeah, we can always add stuff to amend this ADR when we have concrete use cases.

oeps/oep-0049-django-app-patterns.rst
ormsbee: > I realize it's a flexibility vs DRY argument, but I feel we have too many situations where the flexible option is infeasible.

I think of it less as "DRY vs. Flexible", and more as "DRY vs. Decoupled". That may be quibbling semantics, but I think of the latter as more defensively minded. Whenever we serialize to something, we're making some kind of API contract (whether we really mean to or not). So when I think of a serializer in a view, I think, "This is how we're going to serialize this data _for this version of the REST API_, but adding a field to this serializer should not change the CSV we export in this report generation task." It's not that I necessarily _want_ the flexibility to name fields differently, but more that I want a new column suddenly showing up in the middle of my CSV file because a Serializer change meant for one place rippled out to another.

Generally speaking, we haven't been as disciplined as we should be about API compatibility.

oeps/oep-0049-django-app-patterns.rst
ormsbee: Though I would point out that we're not really making any other kind of view these days.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+          supported_programs = q_supported_programs
+
+      return [
+          ProgramData(
+              uuid=program.uuid,
+              title=program.title,
+              status=program.status
+          )
+          for program in supported_programs
+      ]
```
ormsbee, line 139: This is interesting. I think I need to think about it some more, but a few reflex reactions:

1. I do think some sort of Paginator usage makes sense, though I'm not really that familiar with the options and requirements there.
2. Something like this makes sense for a DRF query or something like Django admin, and making use of Django's builtins to accomplish that would be great.
3. We'd probably need some different approach (generator?) if this was being run by a management command and wanted to iterate through a whole bunch of records. Otherwise, this would do a lot of unnecessary computation on the database side to recompute a bunch of queries instead of having one big one.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+  import attr
+
+  def ProgramStatus(Enum):
+      ACTIVE = "active"
+      RETIRED = "retired"
+
+  @attr.attrs(frozen=True)
+  class ProgramData:
+      uuid: str = attr.attrib(
+          validator=[attr.validators.instance_of(str)]
```
ormsbee, line 163: Is it too much to hope for mypy running, so we don't have to do runtime checking of this?

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,211 @@
...
+  │   │   ├── serializers.py
+  │   │   ├── urls.py
+  │   │   └── views.py
+  │   └── urls.py
+  ├── urls.py
+  └── views.py  # existing legacy non-REST APIs
+
+
+.. _signals:
+signals/
```
ormsbee, line 205: 1. Did you mean to give signals its own directory?
2. It sounds like signal handlers are supposed to go in `signals.py`. If I have a long running task that is triggered by a signal handler, should it go in in the same module, or a separate `tasks.py` file?
3. Where is the right place to define a custom signal that your app will emit? It sounds like we probably don't want it in `signals.py`, since that has celery registration side-effects. Maybe `data.py`?

oeps/oep-0049-django-app-patterns.rst
tuchfarber: Thought about 3 a lot. The best way I can think of to keep performance and app model isolation is to pass the paginator directly and lazily transform the responses into data classes. Leaving this code in a comment for now because it's it involves creating custom classes that don't exist yet. I'm curious 1) What you think of the pattern and 2) If it's worth putting in here since without the shared generic code, this would be cumbersome code for an OEP. 

```python
# Things we'd probably throw in edx-django-utils
class TransformingPage(Page):
    def __init__(self, *args, **kwargs):
        self.transformer = kwargs.get("transformer")
        del kwargs["transformer"]
        super().__init__(*args, **kwargs)

    def __getitem__(self, index):
        # Transform the data at the last possible moment so we don't evaluate the queryset
        obj = super().__getitem__(index)
        return self.transformer(obj)

class TransformingPaginator(Paginator):
    def __init__(self, *args, **kwargs):
        self.transformer = kwargs.get("transformer")
        del kwargs["transformer"]
        super().__init__(self, *args, **kwargs)

    # Suggested method to override on Paginator
    def _get_page(self, *args, **kwargs):
        return TransformingPage(*args, **kwargs, transformer=self.transform_function)

###############################
# in api.py
from django.conf.settings import UNSUPPORTED_PROGRAM_UUIDS

from edx_django_utils import TransformingPage, TransformingPaginator

from .data import ProgramData
from .models_api import get_programs as _get_programs

def get_supported_programs(page_size=None):
    """
    Gets all programs that aren't in UNSUPPORTED_PROGRAM_UUIDS settings

    Returns a Paginator of programs
    """
    DEFAULT_PAGE_SIZE = 20

    # _get_programs() returns a queryset
    q_supported_programs = _get_programs().exclude(
        uuid__in=UNSUPPORTED_PROGRAM_UUIDS
    )

    # transformer would probably be a common function elsewhere, but putting it inline for brevity
    return TransformingPaginator(
        q_supported_programs, 
        page_size or DEFAULT_PAGE_SIZE,
        transformer=(lambda program: ProgramData(
            uuid=program.uuid,
            title=program.title,
            status=program.status
        ))
    )
```

oeps/oep-0049-django-app-patterns.rst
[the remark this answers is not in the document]
tuchfarber: Removed!

oeps/oep-0049-django-app-patterns.rst
tuchfarber: :+1: I added a warning that all imports should happen in the ready function else :loop:

oeps/oep-0049-django-app-patterns.rst
tuchfarber: I think I was thinking the serializers would be versioned, but shared within the version. For example `/rest_api/v1/serializers.py` would contain all the serializers for v1 of our API. Then if you need to change or remove fields, you'd create a `/rest_api/v2` with it's own serializers and views. So it's a decouple at the version or at the versioned-view argument, and I don't feel strongly on that. 

I agree about the lack of API discipline, but `I don't want a new column suddenly showing up in the middle of my CSV` isn't version-relevant because we don't version bump on additions.

I added a reference to https://openedx.atlassian.net/wiki/spaces/AC/pages/18350757/edX+REST+API+Conventions since that what OEP-0013 points to as the source of truth. (I didn't know it existed and I don't know why we moved it from an OEP to a wiki page :shrug:)

ormsbee: > I agree about the lack of API discipline, but "I don't want a new column suddenly showing up in the middle of my CSV" isn't version-relevant because we don't version bump on additions.

I'm not sure if I'm interpreting this statement correctly, but just to clarify my own point: We've declared that adding a field to a resource when doing a GET in REST is backwards compatible. But adding that same new field to a CSV may not be backwards compatible, especially if it bumps out other fields. Many CSV readers (including Python's default one) just think of CSV files as numbered columns, so code will read through it and just use numbered indexing to get the column for that data (there is `DictReader`, but my point is just that it's not universal to use named field lookups in scripts like it would be for REST). If I'm sharing the serializer between the two, then something that's a totally allowed and benign change for the REST API results in a contract break for the CSV file.

Maybe that's a contrived example. The main place I can realistically see serializers being shared are when there are linked objects or pushing stuff to the non-yet-implemented message queue. And maybe that's actually fine. I just feel more comfortable when we have more granular control and I know that making a change to make a contract change (even a minor, non-breaking one) is only ever done in an intentional manner.

That being said, I'm honestly fine with having a shared module starting out. We can always amend this if it starts to become a problem–it just means we'd do a little copying to decouple later, and that's not that big a deal.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,223 @@
...
+README.rst
+++++++++++
+Each app should contain a README.rst to explain its use. See full details of what should go in the README.rst in OEP-0019_
+
+.. _OEP-0019: https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#readmes
+
+.. ___init__.py:
+__init__.py
++++++++++++
+The ``__init__.py`` file should either be blank if the app is a plugin, or contain a single line defining the ``default_app_config`` for the app. This ``default_app_config`` should point to the ``AppConfig`` located in ``<app_name>/apps.py``.
```
pdpinch, line 52: We've also been keeping track of the app version number in __init__.py, to support our chatops bot. How strong is this "should be blank" directive?

For example: https://github.com/mitodl/edx-git-auto-export/blob/master/git_auto_export/__init__.py

tuchfarber: Not strong. I meant that it shouldn't be where code is exported like it is in many other python apps, because that's what `api.py` is for. I'll clarify the text tomorrow!

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,221 @@
...
+
+1. API methods defined in ``api.py`` should be well-named, self-consistent, and relevant to its own domain (without exposing technical and implementation details)
+2. An app's Django models and other internal data structures should not be exposed via its Python APIs.
+
+
+Not exposing an app's data structures can be tricky because it's very easy to expose them without meaning to. Therefore there are a couple common strategies we employ.
+
+1. When importing internal app code to be used in the ``api.py`` file, prefix it with an underscore so it's clear it's for internal use only.
+
+2. Create a ``data.py`` file to house simple data objects that can be passed from your app's function to the calling app. By creating these objects, we can avoid both passing Django model objects or querysets directly and having to serialize data. Other apps may import data classes from ``data.py`` in additional to functionality from ``api.py``. See data.py_ for more details.
```
cpennington, line 97: I'm concerned about not exposing QuerySets, due to future performance concerns, but also due to extensibility. For instance, I was just discussing w/ @MichaelRoytman about the course_enrollment api in edx-platform which doesn't currently expose a QuerySet, and as a result he is unable to additionally filter those enrollments inside the database (and instead had to pull the data into memory and then filter in python). That pattern, of pulling into memory and filtering, makes pagination difficult, because to find a page of N users, you may have to filter through the entire data set in python (in the worst case). Even exposing generators just mitigates the problem by only pulling users until you fill up the filtered page, but that doesn't prevent you from needing to iterate the users in python, rather than filtering in the database.

tuchfarber: (Context: this was mainly pulled from https://github.com/edx/edx-platform/blob/master/docs/decisions/0002-inter-app-apis.rst#decisions.) 

I've also been hesitant about performance on this point. There are ways around the performance issue, such as allowing `kwargs` that get passed to the initial queryset, but that makes the hard line we're trying to draw between apps a lot more hazy. It also defeats the purpose of not exposing the internal data structure, but it'd give us a performant API that also only returns plain data classes. 

There was a recent discussion with about `models_api.py` (which I had pulled from this doc), which set it up as the "when you need to break the API contract for performance reasons" API to use. I'm not sure if having a separate file, or having kwargs (above) would be less hazy. 

cc: @nasthagiri since you wrote the referenced ADR and had that `models_api.py` discussion with me.

cpennington: Even exposing **kwargs doesn't actually help that much. For instance, you couldn't use `.annotate` or `.aggregate` if the result isn't returning a QuerySet. I wonder if there's a path forward where we advocate returning `values()` or `values_list()` querysets. Even that, though, means that clients can't choose which columns they care about, which isn't great from performance purposes.

Sigh... I wish we had a better way of being able to use the Python standard of "we're all consenting adults" to feel more comfortable exposing honest-to-goodness querysets. Like, can we just say that non-public fields on a model will be prefixed with `_` on the object (and then we can name the column whatever we want on the field definition)

ormsbee: So coincidentally, I had something like this come up in the ticket that I'm working on today, and my mind also wandered to trying to address it with `values_list`. The problem was:

* I wanted to return an iterable of course contexts that didn't have any learning_sequence outline data associated with them.
* Because I want to punt `learning_sequences` out of edx-platform and into its own repo, I didn't want to make it know about CourseOverview.
* The actual pushing of data into `learning_sequences` happens in contentstore (i.e. Studio), since `learning_sequences` likewise doesn't know about modulestore.

So I settled on a new addition to the `learning_sequences` API that was:

```python
def get_course_keys_with_outlines() -> QuerySet:
    """Queryset of ContextKeys, iterable as a flat list."""
    return LearningContext.objects.values_list('context_key', flat=True)
```

So iterating through it simply goes like you'd expect:

```python
for course_key in get_course_keys_with_outlines():
    print(course_key)
```

Yields:

```
course-v1:TNL+100+2021_01_21
course-v1:TNL+100+ABC
course-v1:TNL+LibContentTest+2021
course-v1:edX+DemoX+Demo_Course
```

But you can also use it as a qset:

```python
all_course_keys_qs = CourseOverview.objects.values_list('id', flat=True)
missing_outlines_qs = all_course_keys_qs.exclude(
    id__in=get_course_keys_with_outlines()
)
```

The query it uses is:

```
SELECT `course_overviews_courseoverview`.`id` FROM `course_overviews_courseoverview`
WHERE NOT (`course_overviews_courseoverview`.`id` IN (SELECT U0.`context_key` FROM `learning_sequences_learningcontext` U0))
```

Not that this will solve every problem, but I think there is something we can do here that is to encourage some predictable subset of the Queryset API in some disciplined way.

ormsbee: I mean, would it work if we subclassed QuerySet to make something that maps into attrs over it when you actually do the iteration? I'll experiment a little...

ormsbee: Thinking on this a little more, let's lay out some use cases for needing to send back a queryset in an API. Some things off the top of my head:

1. Queryset of simple objects like the ContextKeys in the example above, where it's an input going into a larger query.
2. Queryset of objects we want for something like paginated responses, including filtering–used in REST API views.
3. Queryset where a client has some data associated with an entity in another app. So if I'm some new grading system, I might have a 1:M relationship between a `LearningSequence` and my `UserSequenceGrade` table or some such.
4. Similar to (3), but a variant where you don't care about any of the actual extra data on the other end of your relationship, but you just need to be able to associate that data by IDs (e.g. tacking on purchase/upgrade related information to enrollments).

I do think it's fine to not have to optimize all the way down. So for instance, we should definitely avoid the whole N+1, "A new query for every row" problem, but at the same time, having one call per system on that view should be fine. So if I have an API endpoint or page that returns Enrollments, I think it's okay if the query for for `select id from student_courseenrollment...` gets run once for whatever is adding the purchase data, a second time for whatever is displaying course metadata, etc. We could make it a pattern that goes like this:

* App A has a model that has a foreign key constraint on a model in app B.
* App A's API provides a function that takes as an argument a QuerySet of app B entities (really, it just needs to be their IDs), and then returns a QuerySet that has an `attr` data class from App A's `data.py` for things attached to that. I think we could do this by using a custom manager and a subclass of QuerySet, but I don't have time to really check that tonight.

I realize that only covers M:1 and 1:1 relationships and I haven't really got a good story for 1:M and M:M...? Going to mull it over more tomorrow.

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,221 @@
...
+--------
+Proposes a common set of code patterns for Open edX Django apps.
+
+Motivation
+-----------
+As our number of Django apps continue to grow in our many services, we want to coalesce around a couple of standard design patterns to both make switching between codebases easier and to help untangle some of the links between codebases we have today. These decisions should be considered "best practices" or the default patterns, and should only be violated if the code base requires it.
+
+Decision
+--------
+All of our Django apps should have a common structure. This structure consists of a combination of default Django-style code and Open edX-style code. This document will only attempt to detail the common Open edX patterns that we would like to see everywhere, ignoring Django-default items (e.g. ``admin.py``, ``urls.py``, etc) and situation-specific items (e.g. a separate ``constants.py`` file).
```
feanil, line 38: Not a blocking comment but I wasn't able to find a good example of this but it would be nice to link to some docs that explicitly spell out the django side of our layout.  If you're new to django, you won't really know what those are and if you come here for answers, it might be nice to give them a place to go.  The closest I found was this: https://docs.djangoproject.com/en/3.1/intro/tutorial01/#creating-the-polls-app

oeps/oep-0049-django-app-patterns.rst
```diff
@@ -0,0 +1,221 @@
...
+README.rst
+++++++++++
+Each app should contain a README.rst to explain its use. See full details of what should go in the README.rst in OEP-0019_
+
+.. _OEP-0019: https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#readmes
+
+.. ___init__.py:
+__init__.py
++++++++++++
+The ``__init__.py`` file should contain a line for the ``default_app_config`` for the app, or o. This ``default_app_config`` should point to the ``AppConfig`` located in ``<app_name>/apps.py``. It may also contain small app details such as a version. However, unlike many packages, ``__init__.py`` should *not* be used to as the way to export the app's public methods. These should be exported using ``api.py`` (and thus imported as ``from path.to.app.api import public_function``). See api.py_ below.
```
feanil, line 52: `for the app or o.` ? Is this a typo?

oeps/oep-0049-django-app-patterns.rst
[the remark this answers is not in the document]
feanil: I'd prefer we pick one because that will make things consistent which in turn will speed up reading.  I would say `attrs` should be the best practice and you can fall to `dataclass` if you have a good reason(where you like it better is not a good reason).

oeps/oep-0049-django-app-patterns.rst
tuchfarber: I added a table of the things we won't have opinions on with links to their individual docs. Lemme know if this covers your concern.

oeps/oep-0049-django-app-patterns.rst
tuchfarber: Whoops, typo. Fixed!

oeps/oep-0049-django-app-patterns.rst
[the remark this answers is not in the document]
tuchfarber: The majority wins on this one. I removed the `dataclass` reference.

oeps/oep-0049-django-app-patterns.rst
tuchfarber: Based on these comments it seems less likely that we'll be able to have a hard line on `api.py` only returning data classes (which is fine). It also sounds like we might be able to iterate on a future solution where we get both abstractions and performance, but I think we'd like to see those solutions battle tested rather than doc-first. 

In the interim time, where do we think the "good enough for most people, break the rules if you have to" compromise lives? Options I see thus far:
1. Have `api.py` return data classes and have `models_api.py` return querysets
2. Leave everything in `api.py` and more generly suggest data classes over querysets
3. Leave everything in `api.py` and not suggest a preference on data return types 
4. Some better idea you have than 1-3.

ormsbee: Yeah, I don't think we can hold a hard line at the moment. I've been playing around with this a little over the weekend, and I think that we _might_ be able to have our cake and eat it as well, with something like this:

```python

# Assuming there's a TaskData class corresponding to a Task model..

def data_from_model(data_cls, model):
    model_data = {a.name: getattr(model, a.name) for a in attr.fields(data_cls)}
    return data_cls(**model_data)

class TaskDataQuerySet(models.QuerySet):
    def __iter__(self):
        for model in super().__iter__():
            yield data_from_model(TaskData, model)

class Task(models.Model):
    objects = TaskDataQuerySet.as_manager()
    # All the other fields...
```

Then you get a queryset that you can still do query manipulations on (and we can customize the manager to make for more natural queries), but when you iterate it, you get the attrs data structures instead of mutable models.

Parameterizing this stuff was trickier than I thought because QuerySets and Managers expect to be initialized in a kind of magical way deep inside the ORM for certain operations, so I think we'd have to rely on subclassing rather than passing in the class as a constructor param.

I want to try to sit down and really think about the implications of this approach, I don't think I'm going to have time for at least another week or two. :-(

tuchfarber: That looks interesting and similar to how I thought we'd do performant paginators. I'll extend the deadline so we can hash this out. I'd rather it be right than done.

ormsbee: So I think if we want that balance of performance but predictability, it might look something like:

* There's a `DataQuerySet` and `DataQueryManager` in a library we have somewhere.
* People subclass them when they're putting together new models and put class attributes like the data class, conversion specifics, list of attributes you're allowed to sort by, custom manager methods, etc.
* We attach that as something other than the default manager to the model, so that we don't break Django admin and such. Maybe call it `read_only`, if that's what it's effectively going to be?
* Managers and Querysets give us a place to make some flexible querying logic that is an approved subset of the general ORM facilities. We could even make it part of the public API, with the understanding that the search patterns explicitly on those classes are approved, and if you do anything else, you're playing with fire. We're all adults and all, so we don't need to lock things down, just strongly signal.

ormsbee: So for the last part, something might look like:

```python
tasks = api.get_tasks_for_user(user_id)

urgent_tasks = tasks.overdue().high_priority()
```

ormsbee: (Sorry, still doing this in bits and pieces where I can fit in the time...)

I think the kind of interface I posted above could work for the "there is a system out there that I want answers from" type of query, but we'd need something more for the "I want to efficiently add my own data to this" type of query.

Say I was attaching course-specific survey data to enrollments (I am picking totally at random). There's an enrollments API built in the fashion of this OEP, and I have survey objects that have a reverse relationship name of "survey". The approach I suggested above would obliterate the idea of a model relation, and I'm super-skittish about reimplementing that type of logic. But a client _should_ be able to do something like:

```python

tasks = task_api.get_tasks_for_user(user_id)
urgent_tasks = tasks.overdue().high_priority()

jira_tickets = jira_api.tickets_for_tasks(urgent_tasks)
```

Implementing the JIRA API `tickets_for_tasks` by taking a queryset, doing a `select_related`, and returning a matching qset with JIRA information corresponding to the qset that was passed in–that general pattern _should_ be a common use case, right? The JIRA app here gets to determine the reverse relationship name in their model...

tuchfarber: (FYI I extended the review date by a month so we could hash this out. New date is April 10.)

I really like the "have your cake and eat it too" design. I have a lot of hesitations on adding any of it in here though. Reasons being:
1. Whatever we decide on will be an idea and not tested in multiple places. I think this should be an `ideate -> implement -> ADR -> OEP` flow, rather than tagged on here. (Very interested in this topic though, so I'd be happy to be a part of the future work)
2. Sort of in regard to "reimplementing that type of logic", this feels like it's straddling the line between a customization and a contradiction to how Django does things. I don't think we should bend to Django's will, but as one of the larger Django projects, I think this may be a good space for cross-community collaboration to see if this is something someone's already tried, or if they have any thoughts. Honestly it feels like it'd be a rabbit hole to dive into. 

TL;DR: I want this; I want to be in on the discussion; I don't think this is where we decide it.

I'm thinking this section might become more of a "use attrs if possible, querysets are okay for performance reasons".  When we have a working use case, we can update this section to use it. 

That seem fair to you? I'll update the PR with that language if so. 

P.S. I hate stopping a brainstorm, so know that I feel bad about it. Haha.

ormsbee: Sorry I didn't reply to this. I agree that this idea is too rough to be part of this OEP, and it makes sense to experiment with it first.

tuchfarber: Thanks! I've updated the language to be less strict about not passing querysets. If you have any thoughts on it, let me know. If not, I think this was the last major question, so we'll just wait it out until April 10 :)

jinder1s (approved):
Besides the doc string correction, everything here looks good! Let's get this through to the other side.

PR #182, opened 2021-02-01
docs: OEP-0051, Conventional Commits

nedbat:
A provisional OEP for conventional commits.

Original description:
> Mostly, this just establishes the guideline, points at the spec, and
> clarifies what we will and will not do.

### DIFF
--- a/oeps/conventional-commits.rst
+++ b/oeps/conventional-commits.rst
@@ -1,0 +1,77 @@
+####################
+Conventional Commits
+####################
+
+Abstract
+********
+
+Commits should be clearly labeled for their effect on consumers of the repository.  To do this, we adopt the `Conventional Commits`_ guidelines, which detail structured commit messages that clarify the impact of each commit.
+
+This is part of our Change Transparency initiative.
+
+Spec
+****
+
+We are adopting the `Conventional Commits`_ spec, with our own adjustments.
+
+.. note:: Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are not adopting the Angular rules.
+
+Commit messages have these parts::
+
+    <type>: <subject>
+
+    <body>
+
+    <footer>
+
+Type
+====
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+* **chore**: a repetitive task such as updating requirements or translations.
+* **docs**: documentation-only changes. Documentation changes for other types of work can be included in those commits.
+* **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.
+* **fix**: bug fixes.
+* **perf**: performance improvement.
+* **refactor**: code reorganization that doesn't change behavior.
+* **revert**: undo a previous change. The previous commit message, including the type prefix, is included.
+* **style**: improve the styling of the code.
+* **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.
+
+Breaking changes include an exclamation point as part of the type::
+
+    feat!: removed the ability to author courses
+
+Ideally, commits don't mix these types together.  If they do, use the most important type in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore.
+
+Scope
+=====
+
+The `Conventional Commits`_ spec includes a parenthesized scope after the type.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes.  This could change in the future.
+
+Subject
+=======
+
+Commit message subjects should be short.  Put more information in the body of the commit message to fully explain your change.
+
+Body
+====
+
+The subject of the commit is rarely enough information to fully understand the commit.  The body can contain as much information as you like.  Be generous.  Take a moment to think about what you would want to know if someone else had authored this commit.
+
+
+Footer
+======
+
+We are not proposing any formalized footers in the commit message.
+
+
+Tooling
+*******
+
+One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  However, we are not proposing the use of these tools at the moment.
+
+.. _Conventional Commits: https://www.conventionalcommits.org
+.. _Angular commit message format: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format
--- a/oeps/oep-0051-bp-conventional-commits.rst
+++ b/oeps/oep-0051-bp-conventional-commits.rst
@@ -1,16 +1,52 @@
-####################
-Conventional Commits
-####################
+==============================
+OEP-0051: Conventional Commits
+==============================
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :doc:`OEP-0051 </oeps/oep-0051-bp-conventional-commits>`
+   * - Title
+     - Conventional Commits
+   * - Last Modified
+     - 2021-02-24
+   * - Authors
+     - Ned Batchelder
+   * - Arbiter
+     - To Be Determined
+   * - Status
+     - Provisional
+   * - Type
+     - Best Practice
+   * - Created
+     - 2021-02-01
+   * - Review Period
+     - To Be Determined
+   * - Resolution
+     - TK
+   * - References
+     - TK
+
+.. todo:
+   what type for:
+    - commit that removes a depr'd thing
+    - commit that updates a pinned package to get new features?
+
 
 Abstract
-********
+========
 
 Commits should be clearly labeled for their effect on consumers of the repository.  To do this, we adopt the `Conventional Commits`_ guidelines, which detail structured commit messages that clarify the impact of each commit.
 
 This is part of our Change Transparency initiative.
 
-Spec
-****
+Motivation
+==========
+
+
+Specfication
+============
 
 We are adopting the `Conventional Commits`_ spec, with our own adjustments.
 
@@ -25,7 +61,7 @@ Commit messages have these parts::
     <footer>
 
 Type
-====
+----
 
 We use these commit types:
 
@@ -47,31 +83,72 @@ Breaking changes include an exclamation point as part of the type::
 Ideally, commits don't mix these types together.  If they do, use the most important type in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore.
 
 Scope
-=====
+-----
 
 The `Conventional Commits`_ spec includes a parenthesized scope after the type.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes.  This could change in the future.
 
 Subject
-=======
+-------
 
 Commit message subjects should be short.  Put more information in the body of the commit message to fully explain your change.
 
 Body
-====
+----
 
 The subject of the commit is rarely enough information to fully understand the commit.  The body can contain as much information as you like.  Be generous.  Take a moment to think about what you would want to know if someone else had authored this commit.
 
 
 Footer
-======
+------
 
 We are not proposing any formalized footers in the commit message.
 
 
 Tooling
-*******
+=======
 
 One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  However, we are not proposing the use of these tools at the moment.
+
+
+Rationale
+=========
+
+The rationale adds to the specification by describing the events or
+requirements that led to the proposal, what influenced the design, and why
+particular design decisions were made. The rationale could provide evidence
+of consensus within the community and discuss important objections or
+concerns raised during discussion. It could identify any related work, 
+for example, how the feature is supported in other systems.
+
+Backward Compatibility
+======================
+
+This statement identifies whether the proposed change is backward compatible.
+An OEP that introduces backward incompatibilities must describe the
+incompatibilities, with their severity and an explanation of how you propose to
+address these incompatibilities.
+
+Reference Implementation
+========================
+
+The reference implementation must be completed before any OEP is given "Final"
+status, but it need not be completed before the OEP is "Accepted". While there is
+merit to the approach of reaching consensus on the specification and rationale
+before writing code, the principle of "rough consensus and running code" is
+still useful when it comes to resolving many discussions.
+
+Rejected Alternatives
+=====================
+
+This statement describes any alternative designs or implementations that were
+considered and rejected, and why they were not chosen.
+
+Change History
+==============
+
+A list of dated sections that describes a brief summary of each revision of the
+OEP.
+
 
 .. _Conventional Commits: https://www.conventionalcommits.org
 .. _Angular commit message format: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format
--- a/oeps/oep-0051-bp-conventional-commits.rst
+++ b/oeps/oep-0051-bp-conventional-commits.rst
@@ -28,11 +28,6 @@ OEP-0051: Conventional Commits
    * - References
      - TK
 
-.. todo:
-   what type for:
-    - commit that removes a depr'd thing
-    - commit that updates a pinned package to get new features?
-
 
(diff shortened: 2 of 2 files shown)


oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Type
+====
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+* **chore**: a repetitive task such as updating requirements or translations.
+* **docs**: documentation-only changes. Documentation changes for other types of work can be included in those commits.
+* **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.
```
timmc-edx, line 35: One thing that's not clear in the CC spec is how to label generic code "improvements" that aren't exactly _bug_ fixes, and don't introduce new or deprecate existing features, but just... improve them. Stability, security, usability, etc. Performance improvement is already called out as a separate type, but that still leaves this intermediate space ambiguous. (I've already run into this when upgrading a cryptographic signer to a newer, more trusted algorithm.) Sometimes these are user visible, sometimes not.

Would we err on the side of `feat` for these improvements? Or consider any improvement to be a `fix`? Or should there be a different type?

arbrandes: @timmc-edx, from the point of view of someone looking to use commit messages as raw material for release notes, I would very much prefer seeing such improvements under `feat`.

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+Footer
+======
+
+We are not proposing any formalized footers in the commit message.
+
+
+Tooling
+*******
+
+One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  However, we are not proposing the use of these tools at the moment.
```
nasthagiri, line 74: Arbi-BOM has plans in this quarter to add a linter to enforce conventional commits.  cc: @jmbowman

nasthagiri (approved):
Looks great!  Looking forward to using this in `edx-platform` with wider adoption in our other repos.

oeps/conventional-commits.rst
nedbat: I think this is an interesting point.  User-visible changes to features feel different than improved stability or better crypto.  I'm not sure where I would put them.  I think I would tend toward "fix", since they are making an improvement. The reason you made the change was because the existing features will work better than they did before.

oeps/conventional-commits.rst
nedbat: Seems like we should coordinate on that :)

oeps/conventional-commits.rst
nedbat: Maybe the way to explain this is:  "feat" is for changes that would get mentioned in user docs.  It's about what the software does.  "fix" is for changes to improve how the software does it.  

Or did that only make it worse?

kdmccormick: Anecdote: I've made a bajillion commits recently that amount to changing `student.models` to `common.djangoapps.student.models`, and have been wondering which commit type I shoud be using. The only visible change is that a deprecation warning is removed. It's not necessarily a bugfix, but it feels very generous to call that a `feat`.

nasthagiri: @kdmccormick Would that be a `refactor` type?

kdmccormick: Another anecdote that comes to mind are [the XModule-to-XBlock conversions](https://github.com/edx/edx-platform/pull/25965). The goal of the project is to remove our dependency on `XModule` while having no user-facing impact.

Not exactly a `fix`, but not a `feat` either.

kdmccormick: [suggestion] There could be a commit type called `quality`, which would encapuslate this grey area of code-improvements-that-aren't-quite-bugfixes-or-features category. The `style` type could be consolidated into it as well.

kdmccormick: > @kdmccormick Would that be a refactor type?

@nasthagiri , that's a good point. Maybe these grey areas are all just instances of refactoring.

nasthagiri: @timmc-edx It would be useful to have specific examples to consider.

- performance improvement -> `perf`
- security fix -> 'fix`
- usability -> `feat`
- other types of code changes -> `refactor`

nedbat: I would call a usability improvement a "fix", not a "feat".  It doesn't change what the software does, it changes how it does it.

nasthagiri: > I would call a usability improvement a "fix", not a "feat". It doesn't change what the software does, it changes how it does it.

There are multiple benefits of having all user-facing changes designated as "feat" or the like:

- It signals to PR review managers that a Product review is required. _(All other commit-types do not require product review.)_
- It signals the possibility of an a11y and UX review.
- Per @arbrandes's point above, it signals inclusiveness in release notes.

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+    <type>: <subject>
+
+    <body>
+
+    <footer>
+
+Type
+====
+
+We use these commit types:
```
nasthagiri, line 30: What do we think about adding an `api` commit type?  Are APIs that are built for Extensions designated as `feat` or do we introduce a new category for this?

oeps/conventional-commits.rst
nedbat: I'm having a hard time describing every user-facing change as a feature.  If we want fixes to be included in the release notes, then we should read the "fix" commits to include them.  How do other people feel about it?

oeps/conventional-commits.rst
nedbat: "api" is the kind of thing that would be part of the scope if we were using them.  Perhaps scopes is the solution to the "user-visible" aspect also?

oeps/conventional-commits.rst
symbolist: Here are a few tweaks that I think could satisfy some of the concerns in this thread:

* Grouping according to who the commit should interest (no one, developers, everyone). `feat` is also split into `api` (maybe there is a better name?) and `ux` for that.
* `-major` suffix means the developer making the change does the work of figuring out the significance and not whoever is compiling release notes but there are not too many different types to worry about.

Thoughts?

------

Types that should never introduce a breaking change or a UX change:
* **chore**: a repetitive task such as updating requirements or translations.
* **docs**: documentation-only changes. Documentation changes for other types of work can be included in those commits.
* **style**: improve the styling of the code.
* **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.

Types for changes that impact developers but not the UX (any of them could or could not be breaking):
* **api-major**: new APIs, significant changes to APIs or other changes that impact developers that are worth listing in release notes.
* **api**: code reorganization, smaller API changes or fixes.
* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
* **performance**: performance improvement. Significant user visible changes should be marked `ux` instead.

Type for changes that impact the UX:
* **ux-major**: a new feature or a change to UX that is significant and worth listing in the release notes.
* **ux**: a fix or a minor change to UX.

In case of reverts, if all the commits in the revert are of one type use that type. Otherwise:
* **revert**: undo a set of previous changes of different type. The previous commit message, including the type prefix, is included.

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. note:: Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are not adopting the Angular rules.
+
+Commit messages have these parts::
+
+    <type>: <subject>
+
+    <body>
+
+    <footer>
```
symbolist, line 25: Are `body` and `footer` required or optional?

oeps/conventional-commits.rst
nasthagiri: Works for me. ship it!

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
+####################
```
nasthagiri, line 1: As a rollout process question:

Can we trial this proposed set of commit-types for a period of time and get input from devs (somehow) on what's working and what's not?

That is, can we provide a place for devs to provide ongoing feedback on what they find?

- _"None of these types fit for the type of change in this commit."_
- _"Multiple types correspond to this commit and I don't see the value in separating them into multiple commits."_

This feedback will then give us input on how, if at all, we need to ammend.

oeps/conventional-commits.rst
nedbat: This is a departure from the existing conventional commit spec and common usage.  I think we want to stick with the spec as much as we can.  No one else likes the idea of using scopes to address the shortcomings of the types?

nasthagiri: What would using `scopes` look like? Would it over-complicate things, when @symbolist's proposal seems relatively simple and sufficient?

nasthagiri: As a future evolution, I'm thinking we would have folks include what area/app of the platform it affects (since that's useful to know at a glance). For example "grades", "courseware", "certificates", etc.  So that would also add another layer of specificity.

kdmccormick: Under the new proposal my example of a trivial refactoring (`student.models` -> `common.djangoapps.student.models`) becomes a change of type `api:`, which doesn't feel right.

I am leaning towards the original proposal, in which:
* `refactor` doesn't change behavior
* `fix` fixing an existing behavior  (ux, HTTP API, or otherwise)
* `feat` adds new behavior (ux, HTTP API, or otherwise)

Since the frontend replatforming, our direction is that edx-platform is getting fewer and fewer `ux` type changes, with the hope that the edx-platform UIs will eventually go away. Given that, I believe that the `feat`s of edx-platform (or any of our microservices) are naturally going to be HTTP API changes; splitting out `ux` and `api` doesn't seem necessary.

davidjoy: Catching up on the whole thread here -

It seems problematic to me to deviate significantly from the established types that many projects already use, and which are supported out of the box by tools like semantic-release.  And for what it's worth, it would also be different than what we're already doing in all our frontend repositories.

There's a lot of flexibility in the original proposal, which is based on prior art from other projects/tools.

1. Adding color to the types is what `scope` is for, and any given commit WILL have its commit message associated with it still to add even more context.  

2. If something doesn't fit exactly into fix vs. feat vs. non-version-impacting-types, a prefix can be chosen by the effect an engineer thinks it should have on the software version.  When viewed in the release notes, a mildly mis-typed commit with a good message can presumably be forgiven and/or understood anyway with little impact.

Point being, between the existing types, adding a scope, and the message itself, plus any other identifiers we want to decorate our messages with, it seems like we have _plenty_ of nuance to work with without inventing our own standard types.

symbolist: The tweaks I suggested were based on the the assumption that it is useful to be able to tell programmatically:

* If a change only impacts developers only or end users as well.
* If it a large enough change to be worth highlighting in the release notes. Given the high volume of changes in edX a list of all the fix and feature commits is still going to be very long for anyone to process.

So the first question is whether it is useful to be able to programmatically differentiate in one or both of the above? If it is not then yeah we don't need to add anything beyond the base spec and only need to decide on the conventions for the ambiguous cases that were discussed earlier in the thread.

If however it would be useful to programmatically differentiate along either of the two axis, are there any other proposals? Though it does not look like we can use scopes since the spec says:

> A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser):

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+* **chore**: a repetitive task such as updating requirements or translations.
+* **docs**: documentation-only changes. Documentation changes for other types of work can be included in those commits.
+* **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.
+* **fix**: bug fixes.
+* **perf**: performance improvement.
+* **refactor**: code reorganization that doesn't change behavior.
```
feanil, line 38:
```suggestion
* **refactor**: code reorganization that doesn't change behavior from the code consumer perspective.
```

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+Type
+====
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+* **chore**: a repetitive task such as updating requirements or translations.
+* **docs**: documentation-only changes. Documentation changes for other types of work can be included in those commits.
+* **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.
+* **fix**: bug fixes.
```
feanil, line 36:
```suggestion
* **fix**: bug fixes including changes that remove or mitigate security vulnerabilities.
```

oeps/conventional-commits.rst
```diff
@@ -0,0 +1,77 @@
...
+Body
+====
+
+The subject of the commit is rarely enough information to fully understand the commit.  The body can contain as much information as you like.  Be generous.  Take a moment to think about what you would want to know if someone else had authored this commit.
+
+
+Footer
+======
+
+We are not proposing any formalized footers in the commit message.
```
feanil, line 68:
```suggestion
We are not proposing any formalized footers in the commit message. Though the use of  a ``BREAKING CHANGE:`` footer is encouraged as a way of providing more details about breaking changes.
```

Perhaps this just muddies the waters a bit?  It doesn't need to be in there but I think given that this footer is well codified in the spec reminding people about it here would be useful.

oeps/conventional-commits.rst
feanil: I would agree with @davidjoy  and @nedbat about sticking closely to the spec.  I think especially at the beginning since we haven't seen it in action yet.  I think what @nasthagiri summarized is good enough for now and we can iterate if we find a lot of friction with this initial approach.


>     * performance improvement -> `perf`
>     * security fix -> `fix`
>     * usability -> `feat`
>     * other types of code changes -> `refactor`

oeps/conventional-commits.rst
feanil: [suggestion] Would it be useful to have a google form that was open for input regarding this spec at all times?  We could then collect issues over time and update the spec as needed.  This would build a cheap feedback loop and if we can link it from the OEP, it would hopefully be pretty easy to find.

feanil (commented):
I have some suggestions for improvements in clarity but I think this spec is good enough to land as is.  The only major change I would want to see is more clarity on how to provide feedback for future improvements or sources of friction.  Once that's clear, I think this is good from my perspective and we should re-visit it in 6-months with feedback in hand to improve it further.

oeps/conventional-commits.rst
kdmccormick: Does the exclamation point syntax (`feat!:`) covers this case well enough?

nedbat:
About a feedback form: I don't know if people will remember where it is, and use it.  Would it be enough to tell people to raise questions in the #dev Slack channel?

nedbat:
All: I've updated the document and given it the OEP-0051 number.  I'd like to merge it as a Provisional OEP, and start putting it into practice.

oeps/conventional-commits.rst
feanil: In my use, I've used the message body to talk about the change as a whole and the footer to then talk specifically about the breakage and impact.  Not needed, but given that this footer exists mentioning it here reduces a hop to know about it.

feanil (approved):
:shipit:

nedbat:
This is re-opened in #189.

nedbat:
We are looking for an arbiter to get this merged.

PR #186, opened 2021-02-09
BOM-2344 : OEP-18 and make upgrade

mzulqarnain1:
Relevant JIRA: https://openedx.atlassian.net/browse/BOM-2344


.github/workflows/requirements-upgrade.yml
```diff
@@ -0,0 +1,58 @@
+name: Upgrade Python Requirements
```
jmbowman, line 1: Do we still need this, with the new org-wide workflow template?

jmbowman (commented):
We'll need to update the Read the Docs project configuration to use the new requirements file once this is merged.

.github/workflows/requirements-upgrade.yml
mzulqarnain1: Removed.

mzulqarnain1:
> We'll need to update the Read the Docs project configuration to use the new requirements file once this is merged.

Sure. Can you guide me to where those configurations are located?

Makefile
```diff
@@ -227,3 +227,21 @@ pseudoxml:
...
+	# Make sure to compile files after any other files they include!
+	$(PIP_COMPILE) -o requirements/pip_tools.txt requirements/pip_tools.in
+	$(PIP_COMPILE) -o requirements/base.txt requirements/base.in
+	$(PIP_COMPILE) -o requirements/dev.txt requirements/dev.in
+
+.PHONY: requirements
+requirements:
+	pip install -r requirements/base.txt
+
+dev_requirements:
```
feanil, line 246: Not a blocker: Are we standardizing on underscore? I feel like most places use a dash instead.

mzulqarnain1: I've changed this to hyphen instead.

PR #189, opened 2021-02-26
OEP-0051 Conventional Commits

nedbat:
This is an OEP for Conventional Commits. This establishes the guidelines, points at the third-party public spec, and
clarifies what we will and will not do.

Our goal now is to have an OEP that we can put into practice, then iterate.

Previous review and discussion is in another pull request: #182.

### DIFF
--- a/oeps/oep-0051-bp-conventional-commits.rst
+++ b/oeps/oep-0051-bp-conventional-commits.rst
@@ -94,7 +94,7 @@ We use these commit types:
 
 Breaking changes include an exclamation point as part of the type::
 
-    feat!: removed the ability to author courses
+    feat!: remove the ability to author courses
 
 Ideally, a single commit won't mix these types together.  If a commit does, use the most important type in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore, exp.
 
@@ -159,7 +159,7 @@ We've adapted the guidelines to our own projects, and started small.
 
 Here's an example of a conventional commit, with a one-line subject, and details in the body::
 
-    build: private.txt files weren't handled properly
+    build: handle private.txt files properly
 
     The requirements/edx/private.txt file is for dev's own private package
     needs.  There are two installation mechanisms in edx-platform, and
--- a/oeps/oep-0051-bp-conventional-commits.rst
+++ b/oeps/oep-0051-bp-conventional-commits.rst
@@ -10,7 +10,7 @@ OEP-0051: Conventional Commits
    * - Title
      - Conventional Commits
    * - Last Modified
-     - 2021-02-24
+     - 2021-03-16
    * - Authors
      - Ned Batchelder
    * - Arbiter
@@ -44,14 +44,13 @@ Developers should describe the changes they make.  This information is valuable 
 
 The nearer to the change that information is captured, the easier it will be to write, and the more accurate it will be.  Changes made by developers are consumed by many people, from the pull request reviewer to the final user of the product.  The more the developer can explain the change, the more easily the change will be understood and used, and the more efficient the overall development process will be.
 
-Commit messages are the closest text to the change itself.  The more
-information-rich we can make commit messages, the more useful they will be.
+Commit messages are the closest text to the change itself.  The more information-rich we can make commit messages, the more useful they will be.
 
 
 Specification
-============
+=============
 
-We are adopting the `Conventional Commits`_ spec, with our own adjustments.
+We are adopting the `Conventional Commits`_ spec, with our own specifics.
 
 .. note::
    Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are **not** adopting the Angular rules.
@@ -67,13 +66,13 @@ Git commit messages have these parts::
 Type
 ----
 
-We use these commit types:
+We use these commit types labels:
 
 * **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
 
-* **chore**: a repetitive mechanical task such as updating requirements or translations.
+* **chore**: a repetitive mechanical task such as updating requirements or translations. See `Discussion`_ below for how "chore" can hide important information.
 
-* **docs**: documentation-only changes. Documentation changes for other types of work should  be included in those commits. This includes more than the formal docs for a repo, it also covers READMEs, ADRs, docstrings, and so on.
+* **docs**: documentation-only changes. Documentation changes for other types of work should be included in those commits. This includes more than the formal docs for a repo, it also covers any change that updates words describing the work, for example READMEs, ADRs, docstrings, annotations, comments, and so on.
 
 * **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.  Public API entry points and their behavior are part of the feature set of a product, so changes to those should be marked with "feat".
 
@@ -83,30 +82,34 @@ We use these commit types:
 
 * **refactor**: code reorganization that doesn't change behavior from the repo consumer perspective.
 
-* **revert**: undo a previous change. The previous commit message, including the type prefix, is included.
+* **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
 * **style**: improve the styling of the code.
 
 * **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.
 
-* **exp**: experimental or exploratory changes that are not meant to be permanent.  Occasionally changes have to be pushed to GitHub or even production that are intended to be short-lived. For example, debugging continuous integration, or ad-hoc debugging on live systems.
+* **temp**: temporary experimental or exploratory changes that are not meant to be permanent.  Occasionally changes have to be pushed to GitHub or even production that are intended to be short-lived. For example, debugging continuous integration, or ad-hoc debugging on live systems.
 
 
-Breaking changes include an exclamation point as part of the type::
+Breaking changes include an exclamation point as part of the type label::
 
     feat!: remove the ability to author courses
 
-Ideally, a single commit won't mix these types together.  If a commit does, use the most important type in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore, exp.
+Commits should be separated to make it easier to review them before merging and easier to understand them later.  Some types work well in a combined commit, such as a feature along with its tests and documentation.   Refactor, style, chore, and temp commits should be separate from more significant work.
+
+If a commit mixes types, use the most important type label in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore, temp.
 
 Scope
 -----
 
-The `Conventional Commits`_ spec includes a parenthesized scope after the type.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes for now.  This could change in the future.
+The `Conventional Commits`_ spec includes an optional parenthesized scope after the type label.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes for now.  This could change in the future.
 
 Subject
 -------
 
-Commit message subjects should be short enough to fit on one line.  We aren't putting a hard limit on character length, but 70 characters is a good time to turn your attention to the body of the commit message.  Put more information in the body of the commit message to fully explain your change.  Jira or GitHub issue numbers can be included at the end of the subject.
+Commit message subjects should be short enough to fit on one line.  We aren't putting a hard limit on character length, but 70 characters is a good time to turn your attention to the body of the commit message.  Put more information in the body of the commit message to fully explain your change.
+
+Don't include Jira or GitHub issue numbers in the subject.  The body is the right place for links to supporting information.  The subject is precious real estate that should be used for words.  While it is true that it takes more work to get information from the body than from the subject, we are emphasizing writing longer commit messages, and so need to be good at reading the body anyway.
 
 Body
 ----
@@ -115,13 +118,19 @@ The subject of the commit is rarely enough information to fully understand the c
 
 The more information you can put in the body of the commit message, the better. It's not unreasonable to have two paragraphs of explanation in the body.  Especially important are the reasons for the change, or other factors that won't be apparent from the code itself.
 
-Breaking changes especially should have detailed information about the implications and alternatives.
+Breaking changes especially should have detailed information about the implications and alternatives, including a ``BREAKING CHANGE`` footer.
 
+Include references or links to supporting information, such as Jira or GitHub issues.  Some Jira issues will be private to edX.  It is better to link to a private issue than no issue at all. The issue could be made public, and at least its existence is made clear, so people can ask for the information.
+
+Larger decisions should be recorded in Architectural Decision Records, as explained in `OEP-0019`__.
+
+__ https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#adrs
 
 Footer
 ------
 
-We are not requiring any formalized footers in the commit message, though the use of  a ``BREAKING CHANGE:`` footer is encouraged as a way of providing more details about breaking changes.
+Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.  No other footers are specified.
+
 
 
 Discussion
@@ -129,25 +138,21 @@ Discussion
 
 Conventional Commits asks us to categorize changes into a small number of categories.  There will naturally be cases where it isn't obvious which commit type to use.
 
-Choosing the commit type to use for a commit is important, but it's only the first step.  If you have doubts about which commit type to use, choose the highest-priority type that could apply.  Then write a detailed body explaining the full complexity of your change.
+Choosing the commit type label to use for a commit is important, but it's only the first step.  If you have doubts about which commit type to use, choose the highest-priority type that could apply.  Then write a detailed body explaining the full complexity of your change.
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.  
 
-**breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
+**Breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
 
-**DEPR**: removing deprecated code likely is a breaking change.
+**Deprecations**: deprecations happen in two steps: the announcement of the deprecation, and the eventual removal.  The first step is important for people to recognize, but is not a breaking change.  Use clear strong words in the commit subject to be sure people understand the importance.  The second step is usually a breaking change, if a component has been removed.
 
-**pinned dependencies**: updating the version of a pinned dependency seems like just a chore, but consider the repo consumer's perspective.  If an updated dependency adds a feature, then the one-line commit to update the pinned version should be marked "feat".
+**Pinned dependencies**: updating the version of a pinned dependency is "just" a chore, but can bring significant changes to the depending repo.  Conventional commits can't solve this problem.  Reading the commits for a repo won't show that a "chore" might have the effect of a "feat".
+
+**Merge commits**: Commits that git generates (such as merge commits) do not follow these guidelines.  This is not a reason to avoid those commits, though you may want to for other reasons.  The structure of conventional commits will still be useful if merge and other auto-generated commits are in the commit history.
 
 If you are interested to see other discussion about these sorts of questions, the `Conventional Commits repo issues`__ have a number of threads.
 
 __ https://github.com/conventional-commits/conventionalcommits.org/issues
-
-
-Tooling
-=======
-
-One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
 
 
 Rationale
@@ -171,8 +176,37 @@ Here's an example of a conventional commit, with a one-line subject, and details
     - `make requirements` used `private.*` which included private.in, which
       pip-sync balks at.
 
+    Fixes: BOM-2345
+
+
+Open Issues
+===========
+
+Some discussions concerning conventional commits are not yet resolved.
+
+Scope
+-----
+
+Is there a standard for scopes that could be useful?  Component names could be inferred from the files changed in the commit, and will vary from repo to repo.  Perhaps a broad description of what is affected, such as "UI" and "API"?
+
+
+Grammar
+-------
+
+Some commit guidelines are prescriptive about what grammar to use in commit subjects.  The two popular options are imperative mood (what will this commit do? "fix: handle name changes correctly") or past tense (what did this commit do? "fix: corrected the handling of name changes").
+
+
+Tooling
+-------
+
+One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
+
+
+
 Change History
 ==============
+
+2021-03-16: Updates throughout based on review feedback.
 
 2021-02-25: Converted to OEP-0051.
 
--- a/oeps/oep-0051-bp-conventional-commits.rst
+++ b/oeps/oep-0051-bp-conventional-commits.rst
@@ -28,11 +28,6 @@ OEP-0051: Conventional Commits
    * - References
      - TK
 
-.. todo:
-   what type for:
-    - commit that removes a depr'd thing
-    - commit that updates a pinned package to get new features?
-
 
 Abstract
 ========
@@ -44,15 +39,22 @@ This is part of our Change Transparency initiative.
 Motivation
 ==========
 
+Developers should describe the changes they make.  This information is valuable to many audiences: deployers, integrators, end-users, operators, and so on. Developers are in the best position to accurately describe their work and its implications.
+
+The nearer to the change that information is captured, the easier it will be to write, and the more accurate it will be.  Changes made by developers are consumed by many people, from the pull request reviewer to the final user of the product.  The more the developer can explain the change, the more easily the change will be understood and used, and the more efficient the overall development process will be.
+
+Commit messages are the closest text to the change itself.  The more
+information-rich we can make commit messages, the more useful they will be.
+
 
 Specfication
 ============
 
 We are adopting the `Conventional Commits`_ spec, with our own adjustments.
 
-.. note:: Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are not adopting the Angular rules.
+.. note:: Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are **not** adopting the Angular rules.
 
-Commit messages have these parts::
+Git commit messages have these parts::
 
     <type>: <subject>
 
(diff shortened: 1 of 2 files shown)


oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,165 @@
...
+   * - Arbiter
+     - Feanil Patel
+   * - Status
+     - Draft
+   * - Type
+     - Best Practice
+   * - Created
+     - 2021-02-01
+   * - Review Period
+     - To Be Determined
```
feanil, line 25:
```suggestion
     - Until 2021-03-15
```

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,165 @@
...
+   * - Title
+     - Conventional Commits
+   * - Last Modified
+     - 2021-02-24
+   * - Authors
+     - Ned Batchelder
+   * - Arbiter
+     - Feanil Patel
+   * - Status
+     - Draft
```
feanil, line 19:
```suggestion
     - Accepted
```

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,165 @@
...
+   * - Type
+     - Best Practice
+   * - Created
+     - 2021-02-01
+   * - Review Period
+     - To Be Determined
+   * - Resolution
+     - TK
+   * - References
+     - TK
```
feanil, line 29:
```suggestion
     - `Conventional Commits`_
```

feanil (changes requested):

oeps/oep-0051-bp-conventional-commits.rst
feanil: Link to wherever we talk about change transparency maybe?

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Scope
+-----
+
+The `Conventional Commits`_ spec includes a parenthesized scope after the type.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes for now.  This could change in the future.
+
+Subject
+-------
+
+Commit message subjects should be short.  Put more information in the body of the commit message to fully explain your change.
```
kdmccormick, line 105: It's currently common at edX to put the ticket number at the beginning of the subject line; eg `JIRA-1234 we made a change`

In spirit of keeping commit subjects short, should we recommend that ticket numbers (if included) should go in the body?

kdmccormick, line 105: What's short? <=72 chars?

I'm not saying we need a mandate here, but "short" on its own is ambiguous. We could say something fuzzy, like "short enough to fit on a single line", just so that folks reading have an idea of the statement's intent.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,167 @@
...
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+
+* **chore**: a repetitive mechanical task such as updating requirements or translations.
+
+* **docs**: documentation-only changes. Documentation changes for other types of work should  be included in those commits.
+
+* **feat**: a new feature, or a change to an existing feature. Anything that changes the set of features.
```
kdmccormick, line 77: I recall us settling in the earlier discussion that external (eg HTTP) API changes are `feat`s. Since this wasn't obvious to all of us initially, I recommend calling that out here.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,167 @@
...
+
+One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investgate tooling in the future, and are making no recommendations now.
+
+
+Rationale
+=========
+
+The existing Conventional Commits standard is a familiar and widely adopted standard.  It's a good place to start our own guidelines.  It builds on our recently adopted pull request template to help focus developers on providing helpful information about their changes.
+
+We've adapted the guidelines to our own projects, and started small.
```
kdmccormick, line 154: Copying in or linking to an example of a real convential commit to an edX repo may be helpful to the reader.

kdmccormick (commented):
I have a few suggestions, but I'm largely in favor of everything here.

oeps/oep-0051-bp-conventional-commits.rst
inventhouse: Or at the end of the subject so they still show up in `git log --oneline`, but don't get in the way as much

nedbat: Including them at the end of the subject seems good.

nedbat: I've made it:
> Commit message subjects should be short enough to fit on one line.  We aren't putting a hard limit on character length, but 70 characters is a good time to turn your attention to the body of the commit message.

oeps/oep-0051-bp-conventional-commits.rst
nedbat: I've added:
> Public API entry points and their behavior are part of the feature set of a product, so changes to those should be marked with "feat".

oeps/oep-0051-bp-conventional-commits.rst
nedbat: I've added one, let me know if you think it is a good one.

kdmccormick (approved):

oeps/oep-0051-bp-conventional-commits.rst
kdmccormick: Looks great!

nedbat:
About tense, it seems more common to require present tense.  The Angular commit guidelines say:
> use the imperative, present tense: "change" not "changed" nor "changes"

pomegranited (commented):
Glad to see this proposal! It's a great start at making the code history more comprehensible.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Subject
+-------
+
+Commit message subjects should be short enough to fit on one line.  We aren't putting a hard limit on character length, but 70 characters is a good time to turn your attention to the body of the commit message.  Put more information in the body of the commit message to fully explain your change.  Jira or GitHub issue numbers can be included at the end of the subject.
+
+Body
+----
+
+The subject of the commit is rarely enough information to fully understand the commit.  The body can contain as much information as you like.  Be generous.  Take a moment to think about what you would want to know if someone else had authored this commit.
```
swalladge, line 111: Would it be possible to modify this to highlight the importance of explaining _why_ the change was made?  Explaining _what_ is important, but only an overview is required.  One can read the code and code comments to see more about the _what_ and the _how_, but the _why_ can only really be explained well in a commit message.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,181 @@
...
+Commit messages are the closest text to the change itself.  The more
+information-rich we can make commit messages, the more useful they will be.
+
+
+Specfication
+============
+
+We are adopting the `Conventional Commits`_ spec, with our own adjustments.
+
+.. note:: Many people are familiar with the `Angular commit message format`_, which uses conventional commits and inspired the Conventional Commits spec.  We are **not** adopting the Angular rules.
```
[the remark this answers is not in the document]
nedbat: The old form works also, but your suggestion is better (and more conventional!) style in the text file, thanks.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,181 @@
...
+Conventional Commits asks us to categorize changes into a small number of categories.  There will naturally be cases where it isn't obvious which commit type to use.
+
+Choosing the commit type to use for a commit is important, but it's only the first step.  If you have doubts about which commit type to use, choose the highest-priority type that could apply.  Then write a detailed body explaining the full complexity of your change.
+
+**feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.  
+
+**breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
+
+**DEPR**: removing deprecated code likely is a breaking change.
+
```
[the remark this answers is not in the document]
nedbat: This is an interesting point.  I think a lot of the existing conventional commit guidelines are aimed at automated versioning.  For that purpose, deprecation shouldn't be "feat!", because you don't want to bump the major version when you deprecate a feature.  You want to bump it when you remove the feature.

We've been focusing our discussion primarily on people reading the commits, and you are right, they need to sit up and notice when a deprecation happens.  One option would be a new label "depr:", but I also think the word "deprecate" will stand out:
```
feat: deprecate course creation
```
I'm in favor of labeling and signifiers, but there will also still be words. If we write them well, people will read them.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,181 @@
...
+    <footer>
+
+Type
+----
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+
+* **chore**: a repetitive mechanical task such as updating requirements or translations.
```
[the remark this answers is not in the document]
nedbat: Absolutely, we'll be updating them.

oeps/oep-0051-bp-conventional-commits.rst
nedbat: That's a good point, I will add more here.

nedbat:
I've added an "exp" type for exploratory/experimental changes that are meant to be temporary.

swalladge:
@nedbat 

> About tense, it seems more common to require present tense. The Angular commit guidelines say:
> >    use the imperative, present tense: "change" not "changed" nor "changes"

Could this be included in this OEP? It will help with consistency (which is partly the goal here).

inventhouse:
> About tense, it seems more common to require present tense. The Angular commit guidelines say:
> 
> > use the imperative, present tense: "change" not "changed" nor "changes"

But they don't seem to give a rationale for that choice; do you know why?  The example in this proposal is in the past tense; furthermore, commits will be in the past a _lot_ longer than they will be in the present 😉

nedbat:
Thanks, I've corrected the tense of the examples.  I'm reluctant to dictate the grammar now. We will be starting with these guidelines, and iterating.

inventhouse:
> Thanks, I've corrected the tense of the examples. 

No, they were right before 😝   But seriously, I would like to know why some prefer present tense for commit messages.  (Although I admit to sometimes writing them "in the moment" myself, I am actively trying to be consistent with the past tense because they read so much better after the fact)

swalladge:
@inventhouse 

> But seriously, I’m interested to know why some prefer present tense for commit messages. 

Because it ends up forcing a style that is more consistent and readable. It also allows the message to be a grammatically correct sentence.  See https://chris.beams.io/posts/git-commit/#imperative for a good reference.

inventhouse:
@swalladge interesting read, but it doesn’t explain the reasoning either, just an unsupported assertion about “a properly formed Git commit subject line”; for one brief moment “if applied, this commit will...” makes sense, but for the rest of the life of the project, it doesn’t.  Most of the time, commit messages need to explain “what changed?” - past tense, the commit was applied maybe years ago.

Anyway, all this is outside the scope of this OEP; I’m sure we’ll discuss this more when there’s a proposal to standardize commit message tense 😄

stvstnfrd:
> but it doesn’t explain the reasoning either

@inventhouse I believe it does:

> Git itself uses the imperative whenever it creates a commit on your behalf.
> ...
> So when you write your commit messages in the imperative, you’re following Git’s own built-in conventions.

Now, as to _why_ git follows this convention, it's because the Linux kernel does,
both `git` and `linux` having been created by Linus Torvalds (with `git` being created to track Linux kernel source).

But why does _Linux_ do this? They explain it as [1]:

> Describe your changes in imperative mood, e.g. “make xyzzy do frotz” instead of “[This patch] makes xyzzy do frotz” or “[I] changed xyzzy to do frotz”, **as if you are giving orders to the codebase to change its behaviour.**

By using the imperative in our own commits, it maintains consistency and is easier to read, since we're already going to be "stuck" with messages in that tense.

While I won't go as far as saying this needs to be hard requirement at this time, I would recommend adopting it [2].

Disclaimer: I've been recommending that article of best practices [3] for >5 years now, since including it in my original (Stanford) OpenEdX onboarding course. I write my messages in the imperative.

- [1] https://www.kernel.org/doc/html/v4.14/process/submitting-patches.html#describe-your-changes
- [2] Also, it typically leads to shorter messages (in English, anyway) :stuck_out_tongue_winking_eye: 
- [3] https://chris.beams.io/posts/git-commit/

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+
+* **refactor**: code reorganization that doesn't change behavior from the repo consumer perspective.
+
+* **revert**: undo a previous change. The previous commit message, including the type prefix, is included.
+
+* **style**: improve the styling of the code.
+
+* **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.
+
+* **exp**: experimental or exploratory changes that are not meant to be permanent.  Occasionally changes have to be pushed to GitHub or even production that are intended to be short-lived. For example, debugging continuous integration, or ad-hoc debugging on live systems.
```
stvstnfrd, line 92: I'm not a fan of this; it's opaque to me.
- That the descrption provides two possible _exp_lanations for the _exp_ression :wink: (experimental _or_ exploratory) suggests to me that it's not as obvious as it could be.

I'd prefer `temp` or `tmp`, as these terms have broader adoption across the software landscape, and even beyond the technical domain; this would be clear to less technical readers too.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+The nearer to the change that information is captured, the easier it will be to write, and the more accurate it will be.  Changes made by developers are consumed by many people, from the pull request reviewer to the final user of the product.  The more the developer can explain the change, the more easily the change will be understood and used, and the more efficient the overall development process will be.
+
+Commit messages are the closest text to the change itself.  The more
+information-rich we can make commit messages, the more useful they will be.
+
+
+Specification
+============
+
+We are adopting the `Conventional Commits`_ spec, with our own adjustments.
```
stvstnfrd, line 54: Ideally, we would explicitly call-out the areas where we have diverged from the spec(s).

After reading both, it's not obvious to me (or someone coming into this project who already has experience w/ the original spec and/or Angular), what the immediate differences are.

inventhouse:
> > but it doesn’t explain the reasoning either
> 
> @inventhouse I believe it does:
> 
> > Git itself uses the imperative whenever it creates a commit on your behalf.
> > ...
> > So when you write your commit messages in the imperative, you’re following Git’s own built-in conventions.

As it appeared in the article, that still didn't explain the reasoning, however it does highlight an important gap in this OEP:

@nedbat Since git does not follow these conventions _at all_, we ought call out that git-generated commits should be avoided as much as possible, the most common being `merge`, and to strongly prefer fast-forward or squash merges to preserve all this fancy formatting.

@stvstnfrd You went on to explain the reasoning much better than that article did, thank you; if / when we decide to codify the tense (and I think we should), I'll go into more detail; for now it has been deemed out-of-scope for this proposal.

nedbat:
I've been thinking about the tense question (that is, the question about tenses, not that the debate here is getting heated! :) ), and I think I prefer past tense also, as evidenced by my initial examples.  Release notes will be a big reason for writing conventional commits, and they are typically written in the past tense.  I am not a fan of auto-generated release notes, but the smaller we can make the gap between commit messages and release notes, the better.

inventhouse:
@nedbat excellent point; happy to go further into my reasoning if there’s interest, I’m just trying not to hijack this OEP _too_ badly 😄

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+Type
+----
+
+We use these commit types:
+
+* **build**: anything to do with building or releasing the repo: Makefile, tox.ini, Travis, Jenkins, GitHub actions, and so on.
+
+* **chore**: a repetitive mechanical task such as updating requirements or translations.
+
+* **docs**: documentation-only changes. Documentation changes for other types of work should  be included in those commits. This includes more than the formal docs for a repo, it also covers READMEs, ADRs, docstrings, and so on.
```
pdpinch, line 76: It seems like many folks have already adopted `docs:` for the annotation doc-a-thon. Should we mention that case explicitly, or is this a one-time effort? 

```suggestion
* **docs**: documentation-only changes. Documentation changes for other types of work should  be included in those commits. This includes more than the formal docs for a repo, it also covers READMEs, ADRs, docstrings, annotations and so on.
```

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Scope
+-----
+
+The `Conventional Commits`_ spec includes a parenthesized scope after the type.  Open edX repos are large and varied, making standardization of scopes difficult.  We won't use scopes for now.  This could change in the future.
+
+Subject
+-------
+
+Commit message subjects should be short enough to fit on one line.  We aren't putting a hard limit on character length, but 70 characters is a good time to turn your attention to the body of the commit message.  Put more information in the body of the commit message to fully explain your change.  Jira or GitHub issue numbers can be included at the end of the subject.
```
pdpinch, line 109: Please only include Jira issue numbers if the Jira issue is public.

oeps/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
pdpinch: As long as the word "deprecate" appears in the commit subject, when the code is deprecated _and_ when the code is removed, I think that is clear enough.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Choosing the commit type to use for a commit is important, but it's only the first step.  If you have doubts about which commit type to use, choose the highest-priority type that could apply.  Then write a detailed body explaining the full complexity of your change.
+
+**feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.  
+
+**breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
+
+**DEPR**: removing deprecated code likely is a breaking change.
+
+**pinned dependencies**: updating the version of a pinned dependency seems like just a chore, but consider the repo consumer's perspective.  If an updated dependency adds a feature, then the one-line commit to update the pinned version should be marked "feat".
```
pdpinch, line 140: This is really important, because I see a lot of feature changes arrive in edx-platform under the commit message "Updating Python Requirements" 

But I fear it's going to be hard for the committer of a dependency update to manage. 

Consider this commit https://github.com/edx/edx-platform/pull/26589/commits/7c9513d2aed42d1689ebb3eea1c843d8f626dbe0 -- I believe the committer was intending to update astroid but they also picked up an update to the edx-sga xBlock (a bug fix in this case).  

Should the committer be responsible for tracking down the nature of each of the dependency updates? Or do they have a way to scope their commit so it doesn't go beyond the chore they intended?

oeps/oep-0051-bp-conventional-commits.rst
inventhouse: I think we should always include jira numbers if there is one, even if it isn’t public, but:
a) the commit message still needs to stand on its own; the ticket reference should be purely supplemental even if it is public
b) we should endeavor to keep as many of our jira projects and tickets open to the community as possible

OmarIthawi (approved):
I like the change. No further comments. Thank you!

oeps/oep-0051-bp-conventional-commits.rst
nasthagiri: Can we also add a reminder that the Open edX best practice is to document the _why_ of non-trivial changes in [ADRs](https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-documentation.html#adrs)?

From the OEP, ADRs answer the questions:
```
* What technical decisions were made for the component/feature in this repo/app/folder?
* Why does this component/feature do X?
```

A PR's relevant ADR(s) can be committed as separate `docs` commits with the PR itself - or in a previous PR.

oeps/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
nasthagiri: Thanks for this discussion.

Can someone clarify for me where the discussion lands on what the commit message will be when a feature is actually removed?  Would it use the `feat!` prefix or the `fix!` prefix with the removing commit?

oeps/oep-0051-bp-conventional-commits.rst
nasthagiri: I'd imagine that the "chore" of updating python dependencies could pull in newer versions of multiple repos.  It's likely that one or more of the sub-repos could have `feat`, `revert`, `docs`, etc changes in them. I don't believe there's information-value in having dependency-upgrade PRs muddy the semantics of the change by (for example) marking with the "maximum" of the commit-types. As a consumer, you lose information that these changes were brought in through upgrades.

I wonder if better tooling to visualize/summarize the commits taken from the sub-repos could help here insetad. CCing @jmbowman and @awais786 since their team is likely impacted and can provide ideas and input.

nasthagiri (commented):
Looks great overall!
I add a few comments with the intention of strengthening the proposal.

oeps/oep-0051-bp-conventional-commits.rst
inventhouse: I definitely want the ticket numbers to be visible with `git log —oneline`, _especially_ with long log bodies; I think they should be put at the end of the subject but not “counted against” the length.

inventhouse: @nedbat maybe we need to clarify that ticket/issue numbers _should_ be in the subject so they can be seen with `log —oneline`

nasthagiri: @inventhouse Can you explain why you believe it is important to have ticket numbers visible in the git log?

inventhouse: @nasthagiri Wow, there's a lot to unpack; here are a few quick thoughts...

Tickets link to the past: As long as we use tickets of some kind to track work, they will continue to be significant historical artifacts for the changes that result; they will tend to have details and clues that were rightly (at the time) summarized away or simply missed in the commit log.  We must not lose them.

Tickets link to the future: Once a commit is merged to the main branch it can never be modified, but a referenced ticket can be updated with information or links after-the-fact; we have no other way to do this, and we must not lose that.

The connection from a commit to the significant ticket or tickets can be direct and reliable if the commit includes the reference, or (in my experience) a tenuous, unreliable, mini-sleuth (or not-so-mini!)

Even if the ticket is not currently public, it may be made public in the future, but the reference can't be added post-facto.  Even if it is not made public, the reference itself leaves a clue that can be inquired about; the easier it is to see if there might be more information, the easier it is for internal engineers to help the community.

I sincerely believe the importance of including ticket references, no matter how good the commit message might be, cannot be overstated.[^1]

Separately, most of the time I find `git log` much too verbose for showing more than a commit or two; the bigger picture is completely swamped by the details and I have no sense of "where I am" after a couple screens; thus I use `--oneline` a great deal.  Ticket references are the best, shortest, way we have to enable "popping open" more information right from the space-constrained subject line.  Although the full commit message may (hopefully) be better curated, it takes more steps to open and tends to lose my place in the list of logs; if the ticket looks promising, I'll certainly read the log too.  I don't want to lose this option, and I think putting the reference at the end of the subject is a good compromise.

[^1]: Seriously, I think ***every*** commit should reference a ticket, though I'm not (presently) proposing that requirement.

oeps/oep-0051-bp-conventional-commits.rst
nedbat: I see your point.  "temp" was my first choice.  I will switch it.

oeps/oep-0051-bp-conventional-commits.rst
nedbat: I will take a stab at summarizing the differences.

oeps/oep-0051-bp-conventional-commits.rst
nedbat: Absolutely annotations count.  How about, "it also covers any change that updates words describing the work, for example READMEs, ADRs, docstrings, annotations, comments, and so on."

oeps/oep-0051-bp-conventional-commits.rst
nedbat: This I think is going to be a real challenge.  The entire motivation for conventional commits is to make it possible to easily understand the scope of a change from the commit message.  Commits like "chore: automated requirements update" will be a huge gap where significant changes can appear without indication.  I'm not sure what to do about this.

oeps/oep-0051-bp-conventional-commits.rst
```diff
@@ -0,0 +1,184 @@
...
+* **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.
+
+* **exp**: experimental or exploratory changes that are not meant to be permanent.  Occasionally changes have to be pushed to GitHub or even production that are intended to be short-lived. For example, debugging continuous integration, or ad-hoc debugging on live systems.
+
+
+Breaking changes include an exclamation point as part of the type::
+
+    feat!: remove the ability to author courses
+
+Ideally, a single commit won't mix these types together.  If a commit does, use the most important type in the commit message.  The priority order for the types is generally: revert, feat, fix, perf, docs, test, build, refactor, style, chore, exp.
```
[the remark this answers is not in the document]
nedbat: I don't know how important it is to keep every commit pure by type.  If I make a commit that updates a large comment, and nothing else, that's a "docs" change.  But that doesn't mean that all changes to comments should be in a separated "docs" commit.  How bad is it if a feature commit also includes a fix?

(This is why I want to eventually get to commits including updates to changelogs: it gives us the space to fully describe what is happening.)

oeps/oep-0051-bp-conventional-commits.rst
nedbat: (BTW: our OSPR tools looks in PR titles, not commit messages)

nedbat: I'm going to side with the "links in the body" side here.  The subject is best used for words that explain.  Supporting information, including references to Jira or GitHub issues should be in the body.  It's harder to get to, but we're asking people to write longer commit messages. We need to get better at using those messages.

oeps/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
nedbat: If I implement a feature and write docs for that feature, should they be in two separate commits?  I don't understand what value it would add to split them up like that?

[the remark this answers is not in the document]
nasthagiri: That's totally fair - especially regarding docs.

In my mind, I was thinking about keeping the following types of commits separate (as examples):

- `refactor` commits separate from `feat` commits - so it's easier for consumers to understand what functionality actually changed.
- 'style' commits separate - so consumers can easily validate those changes in a separate commit and spend more time on other substantial changes.
- 'temp' commits separate - allows easier removal of those changes in the future.

This suggestion is not a blocker for this OEP.  But a suggestion of this sort can help authors improve the presentation of their code changes and can help consumers better understand the changes.

oeps/oep-0051-bp-conventional-commits.rst
nasthagiri: I defer to @jmbowman on this.  I'm good with what is feasible for Arbi-BOM's continual-upgrade efforts.

oeps/oep-0051-bp-conventional-commits.rst
inventhouse: Oh well; I made my case and I’ve started to get some ideas for tools I can write to help work without the easy ticket references; hopefully they can help others as well.

inventhouse: (But _definitely_ include those references in the commit message!)

oeps/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
nedbat: The removing commit should use "feat!", since it is a breaking change (presumably) to a feature. "feat" should be used for any change in the feature set.

nedbat:
I've incorporated all of the discussions.  Unless there is a strong objection (or embarrassing typo), I would like to merge this as accepted in the next day.

The latest changes (copied from the commit message on 1b1edfdea09095bde09cda17e72fdc7f5e818376):

Breaking:

- Now BREAKING CHANGE footers are required, in keeping with the Conventional Commits spec.

Major:

- Forbade issue numbers in the subject line.

- Added a Discussion about how "chore" commits can hide important information.

- Clarified the type labels to use for two-phase deprecations.

- Added a longer discussion about the forces determining how to split up commits.

- "exp" is now "temp".

- Added an Open Issues section describing some decisions yet to be made, including scope, grammar, and tooling.

Minor:

- Added annotations and comments to the "doc" label.

- Added a reminder about ADRs for larger decisions.

nasthagiri (approved):
Super thrilled to see this!

Now onto adoption!  _What about Metrics? hmm..._

pomegranited:
@nedbat 

> I've incorporated all of the discussions. Unless there is a strong objection (or embarrassing typo), I would like to merge this as accepted in the next day.

No objections from me :) I had another read of the latest version and found no typos or stones unturned.

Thank you for proposing this change!

stvstnfrd (approved):
Thank you to everyone involved in this conversation!

Thanks @nedbat for leading the charge and incorporating all of the feedback!

feanil (approved):
In the famous words of Eddie.

> :+1:  - it's just been arbited.

nedbat:
@feanil it is merged.

PR #191, opened 2021-03-18
fix: use correct rst table syntax and add toc

nedbat:
The table wasn't appearing because the last cell had three columns.  Use
the correct syntax to make the last cell have a bullet list in it.


nedbat:
@feanil ?

jmbowman (approved):
LGTM.  @feanil do you have anything to add as arbiter?

feanil:
Post-Hoc :+1:   Sorry, was knee deep in a different issue.

PR #195, opened 2021-04-07
docs: Adding Decisions section to OEP-005 + feat: Adding Edit on Github button

jinder1s:
Adding "Edit on Github" button to make it easier to correct documentation
Adding decisions to OEP-0005 to increase visibility. This potentially creates an informal standard for how to add decisions to oeps.

### DIFF
--- a/oeps/conf.py
+++ b/oeps/conf.py
@@ -56,6 +56,14 @@ project = u'Open edX Proposals'
 author = edx_theme.AUTHOR
 copyright = edx_theme.COPYRIGHT
 
+html_context = {
+    "display_github": True, # Integrate GitHub
+    "github_user": "edx", # Username
+    "github_repo": "open-edx-proposals", # Repo name
+    "github_version": "master", # Version
+    "conf_py_path": "/oeps/", # Path in the checkout to the docs root
+}
+
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
--- a/oeps/oep-0005-arch-containerize-devstack.rst
+++ b/oeps/oep-0005-arch-containerize-devstack.rst
@@ -232,6 +232,19 @@ environment, and was the prototype implementation of this OEP. However, it
 will need modification to fully meet the specifications of this OEP.
 
 
+Related Decisions
+=================
+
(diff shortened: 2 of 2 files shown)


oeps/conf.py
```diff
@@ -56,6 +56,15 @@
 author = edx_theme.AUTHOR
 copyright = edx_theme.COPYRIGHT
 
+
```
robrap, line 59: Nit: One space is probably fine.

oeps/oep-0005-arch-containerize-devstack.rst
```diff
@@ -232,6 +232,21 @@ environment, and was the prototype implementation of this OEP. However, it
 will need modification to fully meet the specifications of this OEP.
 
 
+Decisions
```
robrap, line 235: "Related Decisions" or "Additional Decisions" or something like that?

oeps/oep-0005-arch-containerize-devstack.rst
```diff
@@ -232,6 +232,21 @@ environment, and was the prototype implementation of this OEP. However, it
...
+=========
+
+Multiple decisions have been made that either modify or append to this OEP.
+The hope is for these decisions to eventually make their way back to this OEP,
+but while they are not, links to decisions are listed below:
```
robrap, line 240: * Do these decisions really modify the OEP?  Would it be just as simple to add explicit links to the decisions in context in the OEP (in addition or in place of this list)?
* What about the following alternative sentence (note the intentional lack of line breaks):
```
The following related decisions modify or enhance this OEP, but have not yet been fully incorporated as updates to this OEP:
```

robrap (commented):
I meant to "Start a Review" earlier, but oh well.

oeps/oep-0005-arch-containerize-devstack.rst
jinder1s: I like your modification to OEP and will add it in next commit.

As for putting explicit links in context of OEP, I opted not to do that, cause:
1. its a-lot of work
2. This whole OEP needs to be updated and these decisions don't really do it well.
3. This changes moves us forward in terms of visibility, so I opted to just do this quickly and hopefully someone else will come back and make the whole OEP better.

robrap (approved):

PR #206, opened 2021-04-16
docs: add warning for using external user id

robrap:
Adds a warning and the start of some clarity around cases where we do not recommend using the internal LMS user_id.

A more formal OEP update should be done in the future, but this at least brings awareness to the new ADR and functionality written since the OEP was first written.

Also, in a separate commit, I made a variety of small fixes to remove warnings and errors during the build.


oeps/oep-0005/decisions/0001-location-of-decentralized-docs.rst
```diff
@@ -24,7 +24,8 @@ Decision
 Place DD documentation(general decisions, general how-tos) in edx/open-edx-proposals/oeps/oep-0005. Eventually, these decisions will be used to update OEP-5.
 
 .. note::
-Any documentation specific to a particular repo should still go in that repo regardless of its relation to DD.
+
+    Any documentation specific to a particular repo should still go in that repo regardless of its relation to DD.
```
jinder1s, line 28: I'm not the biggest fan of using the DD acronym. It seems like an unnecessary layer and Decentralized docs is niche enough of a subject that most people won't know about it all the time. It seems okay in this doc, cause the acronym DD is defined at top. 
I'm mostly giving my opinion here, please feel free to use your judgement(a non-blocking comment)

jinder1s (approved):

oeps/oep-0005/decisions/0001-location-of-decentralized-docs.rst
robrap: I 100% agree with you, but I was just fixing RST warnings here. Feel free to add a PR if you want to update this throughout and I'd approve.

oeps/oep-0032-arch-unique-identifier-for-users.rst
```diff
@@ -33,8 +33,12 @@ We will use the *LMS user_id* to uniquely identify a user in the Open edX system
 
 The *LMS user_id* should be used for all internal events. Additionally, the *LMS user_id* can be used for public events, REST APIs, and JavaScript APIs.
 
-.. _Django auth_user: https://docs.djangoproject.com/en/2.0/topics/auth/default/#user-objects
+.. warning::
+
+    A new external user id should still be used with third-parties under certain circumstances. This is especially important when the user id is sent along with other `Personally Identifiable Information (PII)`_. See the `ADR for the edx-platform external_user_ids Django app`_ for more details.
```
nasthagiri, line 38: [clarification] What does "certain circumstances" mean? Are we not recommending the external-user-id for all external uses?

nasthagiri, line 38: [clarification] What does "when the user id is sent along with other PII" mean?  Doesn't this OEP recommend never sending PII with the user_id?

robrap: I don't feel I have all the answers, or I would help make a more complete update. My intent was to add a temporary improvement to raise awareness that there were additional solutions and situation to consider, which some people were not aware of, because it wasn't listed in the OEP. At a minimum, this might trigger a question and conversation, rather than people presuming incorrectly that they should use the LMS user_id for "certain situations".

I think the LMS user_id is used with Caliper and xAPI. I would consider that external, but I'm not sure what external versus internal means, which is part of what would need to be worked out for a more complete update.

Additionally, I realized that we haven't yet started superseding OEPs. I'm also not clear yet if we think we will be contradicting the original OEP or adding clarity to cases that weren't considered. I know this contradicts what I _thought_ we meant by the OEP.

robrap: Would it be more clear if it read "...a user id..." in place of "...the user id..."? The ADR states that there was a case where PII needed to be sent to a third-party, and sending the LMS user_id was not an option because then the LMS user_id would no longer be anonymous.

Again, I agree that this OEP should get a better update, but I didn't feel prepared for that, and I felt that this warning was better than what existed, because the only questions I've heard about the OEP were from people who happened to know about the ADR and were confused.

I'd be happy to help with a more formal update, but I think that would take some time and a decision to prioritize this.

PR #212, opened 2021-04-30
feat: In OEP-18 recommend putting constraints in `base.in` in libraries

timmc-edx:
Constraints should be expressed in the base.in file whenever it's
important that a relying package also use that constraint. Otherwise,
there's no way to ensure that an IDA's requirements files are compiled
in a way that respects the known incompatibilities in a library.

Also add some more encouragement to impose minimum version constraints in
libraries when developers notice newly added features are being used.

This change is inspired by an observation of an incident where
edx-proctoring was upgraded to compatibility with PyJWT 2.0.0's breaking
changes but edx-platform was limited to PyJWT<2.0.0, leading to errors in
production. This situation could have been prevented only if:

1. A minimum version constraint on PyJWT was added in edx-proctoring, and
2. The constraint was added in base.in rather than constraints.txt.

Both of these were previously in contravention of this OEP.

----

Still needed:

- [ ] Changelog entry
- [ ] Find a new Arbiter
- [ ] Broader distribution for discussion

### DIFF
--- a/oeps/oep-0018-bp-python-dependencies.rst
+++ b/oeps/oep-0018-bp-python-dependencies.rst
@@ -186,7 +186,7 @@ dependencies in order to take advantage of the latest bug fixes, performance
 improvements, and security fixes, we sometimes need to impose some constraints
 on the version to be used. These constraints can either be placed directly on
 the existing dependency in the ``*.in`` file or as a new entry in
-``requirements/constraints.txt``:
+``requirements/constraints.txt``, according to the following rules:
 
 * If it is a constraint on a dependency listed in **a library's base.in**,
   the constraint should be added to the existing dependency line.
@@ -234,17 +234,22 @@ Some guidelines to keep in mind when adding constraints:
   constraint has been imposed.  If there is an issue (either in Jira or an
   upstream issue tracker) for resolving the problem, a link to it should be
   included in the comment.
-* Minimum versions are generally not necessary for IDAs since ``pip-compile``
-  always tries to use the latest compatible version in the generated
-  requirements files; if ``make upgrade`` produces a downgrade in the compiled
-  dependencies, it should be carefully investigated before committing.
-  For library packages, on the other hand, dependency resolution is performed
-  outside of the library's codebase, and specifying minimum version constraints
-  in ``base.in`` when the code depends on newer features is the only way to
-  ensure that a version conflict is detected when the IDA's dependency
-  resolution occurs.
+* Minimum constraints are required in some circumstances, and should be avoided
+  in others.
 
-These constraints should be periodically reviewed to determine if some of them
+  * For IDAs, minimum versions are generally not necessary since ``pip-compile``
+    always tries to use the latest compatible version in the generated
+    requirements files; if ``make upgrade`` produces a downgrade in the compiled
+    dependencies, it should be carefully investigated before committing.
+  * For library packages, minimum versions should be specified when a committer
+    recognizes that their changes depend on a particular version of a package,
+    particularly when it is a relatively new version. With libraries, dependency
+    resolution is performed in the context of a relying package, and specifying
+    minimum version constraints in ``base.in`` is the only way to
+    ensure that a version conflict is detected when the IDA's dependency
+    resolution occurs.
+
+Maximum or exact-version constraints should be periodically reviewed to determine if some of them
 are no longer required.
 
 Generate Exact Dependency Specifications
--- a/oeps/oep-0018-bp-python-dependencies.rst
+++ b/oeps/oep-0018-bp-python-dependencies.rst
@@ -235,7 +235,7 @@ Some guidelines to keep in mind when adding constraints:
   upstream issue tracker) for resolving the problem, a link to it should be
   included in the comment.
 * Minimum constraints are required in some circumstances, and should be avoided
-  in others.
+  in others. (They generally will not need a comment.) Guidelines for when to use:
 
   * For IDAs, minimum versions are generally not necessary since ``pip-compile``
     always tries to use the latest compatible version in the generated
--- a/oeps/oep-0018-bp-python-dependencies.rst
+++ b/oeps/oep-0018-bp-python-dependencies.rst
@@ -184,18 +184,44 @@ Imposing Constraints on Dependencies
 Although we usually want to use the latest available version of our
 dependencies in order to take advantage of the latest bug fixes, performance
 improvements, and security fixes, we sometimes need to impose some constraints
-on the version to be used.  These should be collected in
-``requirements/constraints.txt`` so they can be imposed uniformly across all
-the repository's requirements files; this is done via a ``-c constraints.txt``
-line just under the summary comment of each ``*.in`` file in the
-``requirements`` directory.  Some guidelines to keep in mind when populating
-this file:
+on the version to be used. These constraints can either be placed directly on
+the existing dependency in the ``*.in`` file or as a new entry in
+``requirements/constraints.txt``:
+
+* If it is a constraint on a dependency listed in **a library's base.in**,
+  the constraint should be added to the existing dependency line.
+
+  * This allows constraints to be advertised via ``setup.py`` and taken
+    into account in dependency resolution of relying packages.
+  * If a package advertises an extra feature, constraints should be similarly
+    placed in the extra's corresponding ``*.in`` file.
+
+* All other constraints should instead be added to
+  ``requirements/constraints.txt``.
+
+  * The ``*.in`` files must include the line ``-c constraints.txt`` just below
+    their summary comment. This allows constraints to be shared uniformly
+    across multiple files. (The ``-c`` flag keeps the constraints from being
+    interpreted as dependencies.)
+  * Most constraints could be listed either here or in ``*.in`` files directly,
+    but the constraints file was historically preferred and so continuing to
+    prefer it reduces churn. The only case where the constraints file is
+    specifically required is when multiple ``*.in`` files declare the same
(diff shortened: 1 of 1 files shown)


oeps/oep-0018-bp-python-dependencies.rst
```diff
@@ -208,14 +234,18 @@ this file:
   constraint has been imposed.  If there is an issue (either in Jira or an
   upstream issue tracker) for resolving the problem, a link to it should be
   included in the comment.
-* Minimum versions should generally not be included here; ``pip-compile``
+* Minimum versions are generally not necessary for IDAs since ``pip-compile``
```
robrap, line 237: I wonder if this section could have a different intro with sub-bullets. For example:
```
* Minimum constraints are required in some circumstances, and should be avoided in others. See the following rules for details:

  * For IDAs, ...
  * For library packages, ... 
```

oeps/oep-0018-bp-python-dependencies.rst
```diff
@@ -184,18 +184,44 @@ Imposing Constraints on Dependencies
...
+    their summary comment. This allows constraints to be shared uniformly
+    across multiple files. (The ``-c`` flag keeps the constraints from being
+    interpreted as dependencies.)
+  * Most constraints could be listed either here or in ``*.in`` files directly,
+    but the constraints file was historically preferred and so continuing to
+    prefer it reduces churn. The only case where the constraints file is
+    specifically required is when multiple ``*.in`` files declare the same
+    dependency, but in such a situation it is generally preferred to rearrange
+    things so that one ``*.in`` pulls in the other's ``*.txt`` file  using a
+    ``-r`` line, and only list the dependency in the latter.
```
robrap, line 212: I think I would be more clear that in general we want maximum constraints in the constraints file, unless it must be in a file like base.in.  I think this makes it simpler to go to a single location to understand the list of constraints and remove them as needed. I think this bullet adds additional complexity, to an already complex list of bullets, and doesn't provide enough benefit.

Maybe the doc could get an updated with either my reasoning of why we want to keep these together, or, if there is a reason why we don't want to, help provide that background instead.

oeps/oep-0018-bp-python-dependencies.rst
```diff
@@ -208,14 +234,23 @@ this file:
...
-This file should be periodically reviewed to determine if some of the
-constraints are no longer required.
+* Minimum constraints are required in some circumstances, and should be avoided
+  in others.
+
+  * For IDAs, minimum versions are generally not necessary since ``pip-compile``
+    always tries to use the latest compatible version in the generated
+    requirements files; if ``make upgrade`` produces a downgrade in the compiled
+    dependencies, it should be carefully investigated before committing.
+  * For library packages, minimum versions should be specified when a committer
```
timmc-edx, line 244: I wonder if min versions even deserve a comment. Mostly they should just get ratcheted up, and if someone wants to know why it's specified, it's likely because of a recent change which broke a relying package (and therefore should be easy to find in git history.)

robrap: I agree that the comment is useful for max constraints, and doesn’t seem necessary for minimum constraints that will never Ned to be removed.

oeps/oep-0018-bp-python-dependencies.rst
```diff
@@ -234,17 +234,22 @@ Some guidelines to keep in mind when adding constraints:
...
-  dependencies, it should be carefully investigated before committing.
-  For library packages, on the other hand, dependency resolution is performed
-  outside of the library's codebase, and specifying minimum version constraints
-  in ``base.in`` when the code depends on newer features is the only way to
-  ensure that a version conflict is detected when the IDA's dependency
-  resolution occurs.
-
-These constraints should be periodically reviewed to determine if some of them
+* Minimum constraints are required in some circumstances, and should be avoided
+  in others.
```
robrap, line 238: FYI: Not sure if you prefer consistency with the rest of the doc, but there should be no hard wraps.

oeps/oep-0018-bp-python-dependencies.rst
timmc-edx: I actually don't have a strong preference for where "other" constraints should go. (Other: Singularly-specified, and one or more of non-minimum, non-base, or IDA.) I think that most of our max constraints actually will be in base.in, which removes a lot of the usefulness of constraints.txt as a centralized place to find constraints. But it's also easy to collect all the constraints with a single shell command, even if they're spread out among multiple `.in` files: `grep -Po '^[^#;]+[>=]=[^# ]+' requirements/*.in`

It might actually be simpler to switch to preferring any new constraints to be in-place, and say that constraints.txt is _only_ for new constraints that need to be referenced from multiple places.

oeps/oep-0018-bp-python-dependencies.rst
timmc-edx: Yeah, I was keeping it for consistency. We could do a pass over the repo at some point to unwrap lines.

oeps/oep-0018-bp-python-dependencies.rst
jmbowman: Would we be better off just making the `load_requirements` function in `setup.py` smarter, so that it looks in `constraints.txt` for an override of each package in the `*.in` file being processed?  This would avoid adding a bunch of new logic to the OEP guidelines for people to follow, and keep all our version restrictions in one easily scanned place.  If possible, I'd rather have the computers do extra work instead of developers.

robrap: I like @jmbowman's proposal.

oeps/oep-0018-bp-python-dependencies.rst
```diff
@@ -184,18 +184,44 @@ Imposing Constraints on Dependencies
...
 * Version constraints should only be used to exclude dependency versions which
   are known (or strongly suspected) to not work in at least one context.
 * Constraints on indirect dependencies (used by dependencies but not directly
   by the code in the repository itself) can be added if needed to enforce a
-  compatible version.
+  compatible version. If these indirect dependencies are relevant to the base
+  dependencies in a library, then the dependency should be made explicit in
+  ``base.in`` along with a constraint and a comment explaining that it is only
+  listed for the constraint, and should be removed entirely once the constraint
+  is no longer needed.
```
robrap, line 224: Should we use the term "maximum or pinned constraint" here, to differentiate from min constraints?

oeps/oep-0018-bp-python-dependencies.rst
timmc-edx: I think I'd like the idea of `load_requirements` being smarter if we only had one repo, but as it is we end up with different versions of that logic across over 60 repos. That logic would also be nontrivial—you'd have to determine if each constraint is applicable to base.in or not by comparing it to base.txt, and possibly also deal with URL-based dependencies. It would be nice to package that up, but there's a bootstrapping problem.

Alternatively, `make upgrade` could perform the extraction and dump it into a place that load_requirements can simply read straight in, and there isn't a bootstrapping problem at that time.

But how much benefit do we actually get from having all the constraints in constraints.txt, versus base.in + constraints.txt?

oeps/oep-0018-bp-python-dependencies.rst
timmc-edx: Yes, I think better to express it here than as an exception later.

oeps/oep-0018-bp-python-dependencies.rst
jmbowman: I was thinking of just one addition to the existing logic: before returning the list currently generated by `load_requirements`, see if there's any line in `constraints.txt` that starts with the list entry followed by a character that isn't valid in a package name or URL.  If so, replace the list entry with that line.  The logic shouldn't be that complex, and it could be written to only need to parse `constraints.txt` once.  I can think of one or two pathological corner cases (like when you have different constraints for the same package in `base.in` and `constraints.txt`) where it could produce less than optimal results, but they'd still be better than what we currently have.

And yeah, it's annoying to have to make the logic change in lots of different `setup.py` files.  But I'm concerned that adding complexity on where to put constraints will just result in it being done differently everywhere as developers forget the logic or keep following the old advice.

timmc-edx: Hmm, I just realized that we could use semgrep or pylint to automate upgrading the setup.py code. Something like this, with the code-cleanup running and `semgrep -f rules.yml --autofix setup.py`:

```yaml
rules:
- id: reqs-constrained
  languages:
  - python
  message: |
    Use a newer load_requirements function that knows how to combine base.in
    with constraints.txt.
  pattern: |
    def load_requirements(...):
      ...
    def is_requirement(line):
      ...
  fix: |
    def load_requirements(constraints_path, *requirements_paths):
        etc
  severity: WARNING
```

Since this is a pretty plausible path forward for replacing code blocks quickly, maybe I'll do an experiment to see how bad it is to implement a smarter load_requirements. (I'd still have some suggested changes for this OEP around min constraints, though.)

timmc-edx: (Note: Filed ARCHBOM-1772 for tracking.)

sarina:
@timmc-edx do you plan on continuing with this OEP?

timmc-edx:
No, and thanks for poking me. @rgraber ended up implementing the alternative approach and [deployed it](https://github.com/openedx/edx-cookiecutters/blob/f49b8beb757ae250747dea191e57e6ce037d9686/scripts/update_setup_py.sh) to all of our packages by modifying their setup.py files.

PR #233, opened 2021-08-16
OEP-52 (DRAFT): docs:  add OEP for event bus

feanil:
More details in the OEP but make a case for a specific technology to use
for the message bus and provide context as to why it was chosen over
others.


oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+Cons
+^^^^
+
+* Can lose acked data, even if RAM backed up with an append-only file (AOF).
+* Requires homegrown schema management.
+
+Abstract Message Bus Class
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+@feanil - What is this???
```
feanil, line 217: This is the idea of building an abstraction for the messaging tool we use so that the edx-platform can remain agnostic and just define an interface that can be implemented with either Kafka or Pulsar.  This allows people to have flexibility in choosing their backends.  The cost here is that we would need to build and maintain the abstraction and it would prevent us from using unique features available with whatever backend we chose.

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+     - 2021-08-16
+   * - Review Period
+     - TBD
+
+Overview
+--------
+
+* Adding an event bus to the Open edX platform allows asynchronous event messaging across services, which enables a number of improvements aligned with our architectural goals.
+
+* We are trialing Kafka to implement an Open edX event bus. This decision will be updated as we commit or change direction.
```
regisb, line 34: Whatever the event bus ends up being, please take into account the deployment costs. In my experience, Kafka is a great piece of software that is very difficult to deploy reliably. For reference, here is the complete list of configuration parameters: https://kafka.apache.org/documentation/#configuration Note that Zookeeper needs to be configured separately.
I would strongly suggest a benchmark of Redis Streams: https://redis.io/topics/streams-intro My personal experience with it has been extremely positive. I think it should cover all of the classic use cases.

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+
+Pros
+^^^^
+
+* Already part of Open edX platform
+
+Cons
+^^^^
+
+* Can lose acked data, even if RAM backed up with an append-only file (AOF).
```
regisb, line 211: In practice, exactly-once processing is very difficult to achieve with all brokers, including Kafka. This argument should be balanced with the difficulty of deploying an exactly-once data processing broker.

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+* Requires homegrown schema management.
```
regisb, line 212: Isn't it also the case of Kafka? This is the reason why stream producers use an encoding format such as Protocol Buffers or Avro.

blarghmatey:
My personal bias is for Pulsar being the selected option to provide the pub/sub mechanism. As to it being battle tested I would argue that it has passed that bar having been run at massive scale at Yahoo. There are also a large and growing number of options for managed hosting from companies such as [StreamNative](https://streamnative.io/), [DataStax](https://www.datastax.com/products/astra-streaming), and [Pandio](https://pandio.com/apache-pulsar-as-a-service/). It also has the benefit of supporting multiple methods of messaging and queuing beyond just distributed pub/sub, including AMQP. 

From the scalability perspective, Kafka has the challenge of requiring the number of consumers of a given topic to be known at the time that the topic is created, constraining the ability to scale up or down. There are a number of fundamental architectural constraints built into Kafka that make it challenging to operate and constrain its ability to scale up and down, including the tight coupling of compute and storage, meaning that in order to scale processing capacity it is necessary to also add storage nodes and vice versa. Pulsar is designed in a way to allow processing capacity to be allocated independent of storage capacity making it more conducive to a cloud-oriented operating model. Additionally, Pulsar has native multi-tenancy which can be used to support segregation of workloads among Open edX processes, as well as among distinct installations of the full Open edX platform.

As to the fact that the Kafka ecosystem is more widely adopted, the Pulsar community has been gaining ground on that front. There are also a number of adapters available to layer the Kafka wire protocol on top of Pulsar, and compatibility layers to let the Kafka Connect plugins operate on a Pulsar cluster.

I am happy to engage in further discussion on this topic and I am glad to see that distributed messaging is being added to the Open edX platform architecture.

I will leave a few references for interested parties who would like to learn more about Pulsar:
- [Data Engineering Podcast episode with Pulsar creators](https://www.dataengineeringpodcast.com/pulsar-fast-and-scalable-messaging-with-rajan-dhabalia-and-matteo-merli-episode-17/)
- [Data Engineering Podcast episode with StreamNative founder](https://www.dataengineeringpodcast.com/streamnative-pulsar-streaming-data-episode-132/)
- [Data Engineering Podcast episode about BookKeeper (Pulsar's storage layer)](https://www.dataengineeringpodcast.com/bookkeeper-fast-distributed-storage-episode-193/)
- [Data Engineering Podcast episode about DataStax managed Pulsar](https://www.dataengineeringpodcast.com/datastax-astra-streaming-managed-pulsar-episode-207/)

oeps/oep-0052-arch-event-bus-technology.rst
blarghmatey: Not entirely. Both Pulsar (natively) and Kafka (as a Confluent created add-on) have support for schema enforcement in the write path.

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+
+Messaging Features
+~~~~~~~~~~~~~~~~~~
+
+Kafka is a distributed streaming platform. Kafka's implementation maps nicely to the pub/sub pattern. However, some native features of a message broker like RabbitMQ are not built-in.
+
+There is a useful `blog article comparing Kafka and RabbitMQ`_ by Eran Stiller. The article compares the technologies as pub/sub implementations across the following dimensions (winner in parentheses):
+
+* Message Ordering (Kafka)
+* Message Routing/Filtering (RabbitMQ)
```
ormsbee, line 93: What is the routing/filtering story of how we would use Kafka (or any of these)? How do people decide the granularity of topics?

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+  * Feed into xAPI/Caliper capabilities.
+  * New services and features can be built fully de-coupled from the core application.
+
+Rejected Alternatives
+---------------------
+
+Apache Pulsar
+~~~~~~~~~~~~~
+
+Although rejected to start, `Apache Pulsar`_ remains an option if solving with Kafka turns out to be overly burdensome for developers or operators.
```
robrap, line 180: @blarghmatey: Thanks for your comment: https://github.com/edx/open-edx-proposals/pull/233#issuecomment-913605682. I’m going to move discussion of Pulsar to this section, so we can have a threaded discussion.
1. I think some of what you had written about is already discussed here, but I can adjust accordingly.
2. The one detail I almost missed from your post was about multi-tenancy.  I appreciate you bringing it up.  This may require more research. Typically, there is a way to solve most issues with Kafka, but I’m not sure how ugly or elegant this would be.

robrap (commented):
@regisb @blarghmatey @ormsbee: Thank you for your thoughts. I’ve added some responses.

oeps/oep-0052-arch-event-bus-technology.rst
ormsbee: > I think a topic basically covers a single message type, but maybe you are already assuming that as the coarsest granularity and wondering when and if we should have additionally filtered topics? Or maybe you are asking something entirely different?

Caveat: I don't have operational experience in designing and maintaining this kind of system over time, so a lot of this is conjecture based on limited reading and thinking–I don't know how important these concerns are in practice.

My question goes in both directions. In the long term, I'm interested in finer-grained topics as they relate to permissions. We've had multiple requests where university teams have developed tools that they want to use for a certain subset of courses or across an org. But I suspect this can be solved by having a set of repeaters of some sort that will echo things out to different topics elsewhere, and that we can worry about that at a later time.

In the short/medium term, I'm interested in coarser-grained topics, and I'm not assuming a topic is dedicated to one event type. My intuition (again, not grounded in experience) is that it would be advantageous to combine certain events onto the same topic, depending on the expected clients and ordering constraints. If I have a service that is analyzing student interactions with course content, it's probable that I'm interested in a wide range of different events. Furthermore, it's likely that the ordering will really matter to me: `seq_goto` immediately followed by `seq_next` is very different from `seq_next` followed by `seq_goto`–especially if for some reason there's an issue with a particular topic/partition and things fall behind.

oeps/oep-0052-arch-event-bus-technology.rst
ormsbee: Right, but a counterpoint to haystack is celery/kombu, which gave us an easy path from RabbitMQ to Redis, which has been operationally much simpler. Another is our own internal abstraction of `MongoConnection`, which is enabling recent work to remove the MongoDB dependency from the split modulestore. I think it mostly boils down to how large the interface we have to support actually is, and (as @feanil said) what unique features we lose. It seems like Pulsar and Redis Streams both drew heavy inspiration from Kafka, so I'm optimistic that it's a manageable layer. We'd probably want something like that just to be able to swap it out for a mockable interface when running tests anyhow.

robrap: Agreed. Not much more to write on this topic at this time. We need to make some headway on some other fronts, and then we'll have a little more idea about what would need to be abstracted and how and if that could work. However, in addition to earlier mentioned possible downsides, this would also take additional resources to prove out that the abstraction really could work with Redis Streams. Maybe we could line up wider community resources for that type of effort?

oeps/oep-0052-arch-event-bus-technology.rst
robrap: Great questions that I'm not going to address quite yet. I think we will have lots of ongoing best practices to discover. The main question now is whether we think the technology choice will make or break our ability to do what we need to here. If you are trying to make a point like that, can you clarify the feature you think we require? If not, maybe we should continue this particular thread in some parallel location where we work on discovery of more of our best practices. (Note: sometimes it's just not clear yet whether or not this will affect the technology choice, and maybe you feel this needs to be part of the POC to discover that.)

ormsbee: > The main question now is whether we think the technology choice will make or break our ability to do what we need to here. If you are trying to make a point like that, can you clarify the feature you think we require?

I wasn't trying to make a particular point. I'm just trying to get a better understanding, because from an arch/app point of view, routing and filtering is one of the biggest blindspots I have right now. I get how it would likely work in RabbitMQ, with its more sophisticated routing/matching abilities. But RabbitMQ isn't in serious contention here, and I think these other systems require better up-front planning in that sense. 

Just to be clear:

* I am not advocating for RabbitMQ. I was around for the various operational issues we've had with it in the past.
* I agree that it's possible to get the kind of routing functionality I'm thinking about with the event stream-like systems being seriously considered here. Clearly other organizations have done it. I just don't understand _how_ yet.

> Note: sometimes it's just not clear yet whether or not this will affect the technology choice, and maybe you feel this needs to be part of the POC to discover that.

It seems like the choices on the table are all roughly equivalent in routing functionality (excepting Pulsar's built-in support for multi-tenancy), so I don't think it's necessary to resolve when choosing between them with a POC. I think that it will have a significant effect on how we namespace our events and the interface that we surface for subscribing to them, and that we should allocate some time to hash it out when building the first real thing on the system that's chosen. I don't want to get into a situation where we build some really powerful interface and spend the following five years in operational pain because it runs counter to how the data is actually arranged underneath.

That being said, I think that this OEP should probably touch on the access/permissions options (I think I saw that on a previous doc?). TNL has gotten interest from different course teams and vendors over the years that would love to get event stream data of some sort for their integrations. Having the ability to give an outside user limited permissions to read just one or two particular topics (which we could filter/copy the appropriate events on to) would be really nice.

nasthagiri: On the note of exposing event-bus APIs to external vendors and course teams, I recommend the team consider where event-bus APIs fall in our [API Trichotomy](https://openedx.atlassian.net/wiki/spaces/AC/pages/790036554/API+Trichotomy+Proposal).

From a Boundaries POV, there is value in introducing abstraction/translation layers when crossing between API consumers. For example, xAPI/Caliper Event APIs are intended to be (standards-based) APIs for external consumption. Hence, the support model, documentation, governance, versioning, etc for xAPI/Caliper Events would be different than what we would need for other types of Event APIs.

Similarly, consider where these event-bus APIs fall in that trichotomy spectrum. With a holistic event-driven architecture, a set of events (eg Django Signals) will trigger other events (eg Event-Bus), which in turn will trigger other events (eg downstream notifications).

For those interested, the recent [Flow Architectures book](https://www.goodreads.com/book/show/51771839-flow-architectures) provides a future vision of eventing workflows across organizational boundaries.

oeps/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,264 @@
...
+
+The Open edX platform will benefit from having a message bus. We believe `Apache Kafka`_ is a good choice for the event bus and will begin trialing it to solve some specific platform problems addressed by the `publish-subscribe messaging pattern`_.
+
+.. _Apache Kafka: https://kafka.apache.org/
+.. _publish-subscribe messaging pattern: https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern
+
+Why (trial) Kafka?
+~~~~~~~~~~~~~~~~~~
+
+Kafka has been around for a long time. See `Thoughtworks's technology radar introduced Kafka`_ as "Assess" in 2015, and "Trial" in 2016. It never moved up to "Adopt", and also never moved down to "Hold". Read `Thoughtwork's Kafka decoder page`_ to lear more about its benefits and trade-offs, and how it is used.
```
inventhouse, line 71: Typo: "...to learN more about its benefits and trade-offs..."

oeps/oep-0052-arch-event-bus-technology.rst
ormsbee: I agree that event-bus APIs need to be given the same level of thought as our other APIs in terms of where they fall in terms of exposure and support. FWIW, I don't mean to imply that we would allow third parties to drink from the same firehose of events as the built-in services. Only that it would be good if the bus itself were capable of more granular permissions, so that we could send some more controlled set of events to endpoints that third parties could integrate with.

tuchfarber:
Another 2 cents post:
[Anecdotal] Having done a minor version of this for a hackathon, working with Pulsar was extremely developer friendly. I was able to spin up an instance and implement publishers and subscribers with the client libraries rather quickly.

[Actual concern] Though this OEP isn't a Kafka vs Pulsar (vs others) doc, that seems to be where most of the concern is at this point. Reading between the lines, the Kafka decision feels more like:
> Kafka is the right decision *right now*. Pulsar (or another) would be preferred, but it needs a bit more time and preferably an AWS managed solution (to appease edx.org)

Under the assumption that that is true, I would definitely lean towards this potential abstraction layer. Out of the gate it would make the community happier and it would prepare us for a future where we swap out from the stable-but-cumbersome Kafka.

robrap:
Thanks @tuchfarber (and many others). 
1. Although we were starting with Kafka, I was also trying to list some of the concerns we need to investigate, regardless of technology.
2. We may be switching to Pulsar to try some of its unique features. I’m more concerned with completing a POC to see how it all hangs together.
3. As we proceed with our POC, I think we’ll have a better idea of how and if an abstraction layer might work.

This early draft of an OEP was less about making a quick decision, but more about eliciting feedback so concerns like multi-tenancy and a potential abstraction layer can ultimately be addressed.

oeps/oep-0052-arch-event-bus-technology.rst
regisb: > In general, the hosting and deployment story for each Open edX deployment makes this OEP that much more complex. In the case of Amazon’s ElastiCache for Redis (which we use), AOF is not an option for multi-AZ, and they state that multi-AZ has better resiliency. However, it’s not a great story for disaster recovery, which means some homegrown solution of backing up and being able to restore all events. Instead, I think we are going to need a solution with persistence.

The fact that ElastiCache for Redis does not come with persistence has come up before (for the replacement of Rabbitmq) and it baffles me. It means that on AWS Redis can only be used as a cache layer, which drastically limits its capabilities. Thus, Redis should not be deployed with ElastiCache. But why limit yourself to AWS-offered options? Redis Enterprise offers AWS-hosted solutions that come with persistence out of the box: https://redis.com/redis-enterprise-cloud/pricing/ I am sure there are other Redis providers out there that offer a better storage story.

> Another con that is missing here is that because Redis is in-memory, it is limited by RAM. I also don’t think it has a good compaction story. This seems like a hard limit that again would be very costly to work around, like trying to manage and balance across multiple Redis clusters, or any other custom solution we might come up with.

Do we have an estimate for the amount of data (/user/day) that we would need to store? I know that deploying a full blown event-driven architecture requires storing all events forever, but surely this is intractable for most Open edX system administrators -- even when stored on disk. In order to dismiss RAM-based storage, we need an estimate for the amount of stored data.

oeps/oep-0052-arch-event-bus-technology.rst
regisb: The features brought by native schema management seem minimal compared to bringing our own data serialization format to the table. Also, the Pulsar Python client supports only a subset of the schema features: https://docs.google.com/spreadsheets/d/1iTzn6QnOpTYK8QQHxXRb6a25nvuTNGPPJDdUQHGBUS4/edit#gid=1784579914

oeps/oep-0052-arch-event-bus-technology.rst
regisb: I like where this conversation is going. To summarize the arguments:

1. We need to define what features will be needed from our event bus to figure out whether an abstraction layer can be built.
2. The selected event bus will have to fit in a devstack.
3. We need a good deployment story for most Open edX hosting platforms, and not just edX.org.

oeps/oep-0052-arch-event-bus-technology.rst
feanil: Looking at both Pulsar and Kafka, this is possible and we can easily create access control rules to give users access to specific topics as read-only or read-write.  So we should be able to fulfill this future use case with either technology.

oeps/oep-0052-arch-event-bus-technology.rst
feanil: An update from myself and @robrap as we work more with this, we are starting to see that the API may not be as complex as we initially thought and an abstraction layer may be more viable.  No promises yet but , it's something we'll be looking at more as we build a working version of this without the abstraction layer.

@regisb thanks for the summary, I think it captures the current conversation very well.  One thought I had is reasonably quantifying "good" in the 3rd point.  I think this is perhaps a good conversation independent of this particular technology conversation so I've re-framed/[asked that question in the forums](https://discuss.openedx.org/t/a-good-deployment-story-for-most-open-edx-hosting-platforms/5915).

oeps/architectural-decisions/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,277 @@
...
+
+.. _Apache Pulsar: https://pulsar.apache.org/
+.. _Pulsar: https://pulsar.apache.org/
+.. _Kafka: https://kafka.apache.org/
+.. _publish-subscribe messaging pattern: https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern
+
+Why (trial) Pulsar?
+~~~~~~~~~~~~~~~~~~~
+
+Pulsar and Kafka were both in very close running in our initial assessment.  In writing code to work with kafka and examanining its client interface, we saw the lack of a synchronous interface for writing data to be problemating.  It meant we would have to write more complex code in order to gauranteed that messages were persisted and make use of callbacks to respond to errors.  Pulsar on the other hand provides both an async and synchronous message sending interface.  The synchronous interface makes it a lot easier to write code when you need to ensure that database comitting and message sending happen atomically.
```
robrap, line 75:
```suggestion
Pulsar and Kafka were both in very close running in our initial assessment.  In writing code to work with Kafka and examining its client interface, we saw the lack of a synchronous interface for writing data to be problematic.  It meant we would have to write more complex code in order to guarantee that messages were persisted and make use of callbacks to respond to errors.  Pulsar on the other hand provides both an async and synchronous message sending interface.  The synchronous interface makes it a lot easier to write code when you need to ensure that database committing and message sending happen atomically.
```

oeps/architectural-decisions/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,277 @@
...
+
+In trialing Kafka we saw that some of our initial assumptions about the ease of operating Kafka in AWS due to the availablity of MSK(Managed Streamking Kafka) were not as adventageous as we expected.  While MSK solves the issue of deploying the broker, Kafka has enough add-ons(schema mangement, partition-management tool) that we don't actually reduce operational burden by having Kafka as a managed service compared to Pulsar.
```
robrap, line 77:
```suggestion
In trialing Kafka we saw that some of our initial assumptions about the ease of operating Kafka in AWS due to the availability of MSK (Managed Streaming Kafka) were not as advantageous as we expected.  While MSK solves the issue of deploying the broker, Kafka has enough add-ons (schema management, partition-management tool) that we don't actually reduce operational burden by having Kafka as a managed service compared to Pulsar.
```

oeps/architectural-decisions/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,277 @@
...
+* Ease of data retention capabilities.
+* Additional built-in pub/sub features (built-in, according to docs).
+* 
+
+Cons
+^^^^
+
+* Requires 3rd party hosting or larger upfront investment if self-hosted (kubernetes).
+* Less mature (but growing) community, little documentation, and few answers.
+* Python built-in schema management is hard to work with for complex use cases.
```
robrap, line 206:
```suggestion
* Python built-in schema management is buggy and hard to work with for complex use cases.
```

oeps/architectural-decisions/oep-0052-arch-event-bus-technology.rst
```diff
@@ -0,0 +1,277 @@
...
+     - 2021-08-16
+   * - Review Period
+     - TBD
+
+Overview
+--------
+
+* Adding an event bus to the Open edX platform allows asynchronous event messaging across services, which enables a number of improvements aligned with our architectural goals.
+
+* We are trialing Kafka to implement an Open edX event bus. This decision will be updated as we commit or change direction.
```
robrap, line 34:
```suggestion
* We are trialing Pulsar (and Kafka) to implement an Open edX event bus. This decision will be updated as we commit or change direction.
```

robrap (commented):
@feanil: Mostly fixing typos, but added some minor comments.

oeps/architectural-decisions/oep-0041-arch-async-server-event-messaging.rst
```diff
@@ -433,8 +437,8 @@ outages over the years, particularly when course-discovery becomes overloaded
 with edx-platform generated requests.
 
 
-Eliminate Expensive, Batch Synchronization
-------------------------------------------
+Eliminate Expensive, Delayed, Batch Synchronization
```
robrap, line 440: Hopefully some of these changes will land in a separate PR: https://github.com/openedx/open-edx-proposals/pull/321/files

robrap:
Hello everyone. This PR is being closed in favor of this newer PR: https://github.com/openedx/open-edx-proposals/pull/324. See the updated OEP's "Change History" for more details of what has been updated. You can also determine if you have early comments that have not yet been addressed, and we can continue those discussions. Thank you!

FYI: @ormsbee @feanil @regisb

robrap:
Oops. See https://github.com/openedx/open-edx-proposals/pull/233#issuecomment-1083584254 for closing comment.

PR #246, opened 2021-10-01
doc: Update OEP-1

feanil:
Add a new "Obsolete" status that can be assigned to OEPs that are no
longer relevant but don't have a new OEP supersedeing them.
OEP-7(Migrating to Python 3) is a good example of an OEP that would
benefit from this status existing.


oeps/oep-0001.rst
```diff
@@ -232,6 +246,15 @@ OEPs can also be superseded by a different OEP, rendering the original
 obsolete. In that case, the OEP's status should be changed to "Replaced"
 and updated with a link to its superseding OEP.
 
+Obsolete
```
robrap, line 249: I'm not really opposed to this, but I wonder if it is a problem simply because we don't have good index-like docs to what is relevant, so someone might still feel compelled to review all OEPs.

feanil: A follow-up to this that I would like to do is to organize the OEPs  a bit better in the index by what kind of OEPs they are but as a part of that I wanted to move the obsolete ones out of the way.

nedbat (approved):
Fix the change history, and it's good to go. Thanks for the svg flowchart! (and for the promotion to arbiter...!)

PR #252, opened 2021-10-21
docs: Update deprecation OEP with suggestions on release cycle alignment

timmc-edx:


oeps/processes/oep-0021-proc-deprecation.rst
```diff
@@ -174,6 +174,20 @@ considers the timing of the next `Open edX named release`_.
...
+afterwards. In this example, a deployment using named releases would
+have a number of months to adjust their Maple deployment to stop
+depending on the deprecated feature before Nutmeg fully removes
+it. This approach would be most appropriate for features that can be
+left in place for an extended period before removal and where a
+transition to an alternative would require a moderate to large amount
+of effort. For more trivial deprecations, it may be appropriate to
+simply deprecate and remove within the same release cycle.
+
+Either way, deprecations and removals should be included in the named release's release notes.
```
timmc-edx, line 189: Note to self, feedback from Diana: "if we want to require adding deprecations into the release notes, we should put that line in the Removing/Removed section" -- plan on making this change.

felipemontoya:
I like this PR and the sentiment behind it. :+1:

oeps/processes/oep-0021-proc-deprecation.rst
pdpinch: As the release notes writer for the past two named releases, I definitely tried to include deprecation information. For the most part, I picked it up from reading the named release page in the Confluence wiki, and by skimming through DEPR tickets. I'd like to think some details could be extracted from commit messages as well. 

What do you all have in mind?

timmc-edx: I did a bit of a rewrite and included a note in the Accepted section. Does this seem about right?

feanil (approved):

idegtiarov:
I like this PR as well. 👍🏻

oeps/processes/oep-0021-proc-deprecation.rst
```diff
@@ -195,6 +195,20 @@ considers the timing of the next `Open edX named release`_.
   Consider when the next Named Release is cut; if it is very soon, you may wish to delay final
   removal until after the cut date.
 
+Consider choosing deprecation and removal dates that allow for a full
+release cycle for transition planning. For example, a deprecation
+proposal could be accepted while Maple is being finalized, then
+implement the removal some time after Maple is released (so that the
```
sarina, line 201: Nit: The removal can actually come as soon as Maple is cut - at that point, all changes will be in Nutmeg.

sarina (approved):

oeps/processes/oep-0021-proc-deprecation.rst
timmc-edx: This is true, but I'm a little concerned that removals that happen shortly after branch cut may interfere with fixes that need to be backported. Maybe that's too rare to be concerned about, though.

pdpinch:
You can see my draft of the deprecations section of the nutmeg release notes at https://edx--2056.org.readthedocs.build/projects/open-edx-release-notes/en/2056/nutmeg.html#deprecations

If you have comments or questions about the nutmeg release notes specifically, please comment on the PR https://github.com/openedx/edx-documentation/pull/2056

oeps/processes/oep-0021-proc-deprecation.rst
sarina: Yeah. I don't feel entirely strongly either way. I think sometimes it can be really annoying to have to hold up artificially another two months, and practically, people are merging breaking things after a given release is cut. This also has the benefit of more time on master and thus more tested before the next release. But! This is why I said "nit" - I would defer to you.

sarina: Maybe a compromise could be something like,

```
implement the removal some time after Maple is released (so that the
removal itself will land in Nutmeg). Note that while removals technically
can come as soon as a named release is _cut_, it is discouraged as the
removal may interfere with fixes that need to be backported. Any deployment...
```

sarina:
@timmc-edx this has two approvals and I think is good to go. I added one suggestion to our back and forth from earlier, but I really don't feel super strongly about it. As author and someone who participates in DEPR, I defer to you. Let's get this merged!

oeps/processes/oep-0021-proc-deprecation.rst
timmc-edx: I updated the language a bit -- more or less what you said. I think it's ready for merge if you do. :-) (Would you be acting as Arbiter here? I'm not totally clear on the OEP-update process, even after having looked at OEP-1.)

sarina: I find OEP-1 written like OEPs will hardly ever change because a change implies a new OEP.

My personal interpretation is a reasonably-sized change to an OEP should have an arbiter, but smaller things (ie, clarifying points or updating the OEP to reflect how we actually operate) don't need one - just good review.

To me this seems like it falls in the latter bucket - it's a small change that mostly just clarifies how we commonly operate. If you think it's larger, the "correct" way to do things would be to first ask the original arbiter and if s/he isn't available, find a new one.

timmc-edx: Makes sense!

sarina:
Thanks for remembering the changelog! Here is an extra 👍🏻 from me!

PR #254, opened 2021-11-01
docs: add squash conventional commit prefix (OEP-51)

robrap:
Add a new "squash" conventional commit prefix as an alternative
to omitting the prefix to designate you intend to squash, or using
another prefix like "temp" or "fix", that wasn't intended for this
purpose.

### DIFF
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -211,7 +211,7 @@ Tooling
 
 One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
 
-
+As we trial conformance checking, one open question is whether or not "squash" commits should pass or fail commit linting. See an initial discussion of `linting squash commits in this PR comment thread<https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683>`__.
 
 Change History
 ==============
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -216,7 +216,7 @@ As we trial conformance checking, one open question is whether or not "squash" c
 Change History
 ==============
 
-2021-11-01: Squash commits.
+2021-11-01: Add a new ``squash`` commit type.
 
 2021-09-08: Scopes are optional, but unstandardized.
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -211,7 +211,7 @@ Tooling
 
 One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
 
-As we trial conformance checking, one open question is whether or not "squash" commits should pass or fail commit linting. See an initial discussion of `linting squash commits in this PR comment thread<https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683>`__.
+As we trial conformance checking, one open question is whether or not "squash" commits should pass or fail commit linting. See an initial discussion of `linting squash commits in this PR comment thread <https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683>`__.
 
 Change History
 ==============
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -88,6 +88,8 @@ We use these commit types labels:
 
 * **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
+* **squash**: temporary changes that are intended to be squashed before merged.  Use with caution, because a linter for conventional commits can no longer remind you to squash by failing on missing prefixes.
+
 * **style**: improve the styling of the code.
 
 * **test**: test-only changes. Adds missing tests or corrects existing tests. Tests accompanying other types of changes go into those commits.
@@ -147,6 +149,8 @@ Conventional Commits asks us to categorize changes into a small number of catego
 Choosing the commit type label to use for a commit is important, but it's only the first step.  If you have doubts about which commit type to use, choose the highest-priority type that could apply.  Then write a detailed body explaining the full complexity of your change.
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.
+
+**squash vs temp vs unknown or missing prefix**: commits that are not intended to be merged could use no prefix or an unknown prefix to communicate to a linter that a PR is not ready to merge. This is nice, but it also forces reviewers to more closely track why a PR is failing throughout the review process. An alternative is to categorize a commit that is *not* intended to be squashed before merge by using "squash". Do *not* use "temp" or "fix" for this use case, since these categories convey a different meaning once merged, and are intended to be merged. Note that if you use "squash" commits, rather than omitting the prefix, you will no longer get reminded to squash through a failed PR.
 
 **Breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
 
@@ -211,6 +215,8 @@ One of the advantages of formalized commit messages is using them as input to to
 
 Change History
 ==============
+
+2021-11-01: Squash commits.
 
 2021-09-08: Scopes are optional, but unstandardized.
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -150,7 +150,7 @@ Choosing the commit type label to use for a commit is important, but it's only t
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.
 
-**squash vs temp or fix**: commits that are intended to be squashed before merge should use the "squash" prefix. Although "temp" or "fix" may feel appropriate, they are not meant to be used in this way and convey something meaningful after a PR merges, where "squash" commits are not intended to be merged.
+**squash vs temp or fix**: use "squash" when the *commit* is temporary, and you intend to squash the commit into a larger commit with a different semantic label. Use "temp" when the *change* is temporary, but you intend to merge the commit. Again, use "squash" over "fix" when the *commit* is temporary and is not intended to be merged.
 
 **Breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -88,7 +88,7 @@ We use these commit types labels:
 
 * **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
-* **squash**: temporary changes that are intended to be squashed before merged.
+* **squash**: temporary changes that are intended to be squashed before merging.
 
 * **style**: improve the styling of the code.
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -88,7 +88,7 @@ We use these commit types labels:
 
 * **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
-* **squash**: temporary changes that are intended to be squashed before merged.  Use with caution, because a linter for conventional commits can no longer remind you to squash by failing on missing prefixes.
+* **squash**: temporary changes that are intended to be squashed before merged.
 
 * **style**: improve the styling of the code.
 
@@ -150,7 +150,7 @@ Choosing the commit type label to use for a commit is important, but it's only t
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.
 
-**squash vs temp vs unknown or missing prefix**: commits that are not intended to be merged could use no prefix or an unknown prefix to communicate to a linter that a PR is not ready to merge. This is nice, but it also forces reviewers to more closely track why a PR is failing throughout the review process. An alternative is to categorize a commit that is *not* intended to be squashed before merge by using "squash". Do *not* use "temp" or "fix" for this use case, since these categories convey a different meaning once merged, and are intended to be merged. Note that if you use "squash" commits, rather than omitting the prefix, you will no longer get reminded to squash through a failed PR.
+**squash vs temp or fix**: commits that are intended to be squashed before merge should use the "squash" prefix. Although "temp" or "fix" may feel appropriate, they are not meant to be used in this way and convey something meaningful after a PR merges, where "squash" commits are not intended to be merged.
 
 **Breaking changes to features**: changing how a feature works is not a breaking change.  For example, users are sent to a new experience instead of the old experience. This is not a breaking change.  It should get a "feat" label, but not a "feat!" label.
 
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -24,11 +24,11 @@ jinja2==3.0.2
     # via sphinx
 markupsafe==2.0.1
     # via jinja2
-packaging==21.0
+packaging==21.2
     # via sphinx
 pygments==2.10.0
     # via sphinx
-pyparsing==3.0.1
+pyparsing==2.4.7
     # via packaging
 pytz==2021.3
     # via babel
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -46,7 +46,7 @@ markupsafe==2.0.1
     # via
     #   -r requirements/base.txt
     #   jinja2
-packaging==21.0
+packaging==21.2
     # via
     #   -r requirements/base.txt
     #   sphinx
@@ -54,7 +54,7 @@ pygments==2.10.0
     # via
     #   -r requirements/base.txt
     #   sphinx
-pyparsing==3.0.1
+pyparsing==2.4.7
     # via
     #   -r requirements/base.txt
     #   packaging


oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -148,6 +150,8 @@ Choosing the commit type label to use for a commit is important, but it's only t
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.
 
+**squash vs temp vs unknown or missing prefix**: commits that are not intended to be merged could use no prefix or an unknown prefix to communicate to a linter that a PR is not ready to merge. This is nice, but it also forces reviewers to more closely track why a PR is failing throughout the review process. An alternative is to categorize a commit that is *not* intended to be squashed before merge by using "squash". Do *not* use "temp" or "fix" for this use case, since these categories convey a different meaning once merged, and are intended to be merged. Note that if you use "squash" commits, rather than omitting the prefix, you will no longer get reminded to squash through a failed PR.
```
inventhouse, line 153: The purpose of a pre-merge check is to catch things that should not be merged to the main branch; this is true of code linting, unit tests, and now commit linting.  Making commitlint "turn a blind eye" to commits that _expressly_ should not be merged defeats its purpose.  It's much better for a reviewer to keep track of why a PR is failing checks than to track whether is is passing checks that it shouldn't; pre-merge checks should _only_ pass when the PR is, in fact, ready to merge.

Put it like this:
- You have a test case you want to be sure to cover before you merge so you're going to commit a stub placeholder version now and fill it in before you merge; is it better to make that stub pass so the unit test check stays green and it's easier to see if other tests break or should you make it fail so the unit test check ensures that placeholder will not get merged?  I assert that the latter is much better practice.
- You have a commit that you want to squash before you merge; is it better to use a commit type that will pass the check so your PR stays green or to use _anything else_ so the check ensures that commits you don't intend to merge don't end up on the main branch?  Again, I assert the latter is much better practice.

I think a `squash:` type is a great suggestion and we should use it.  However, as its express purpose is to mark commits that are not intended to be merged, commitlint _must_ flag such commits to reviewers since the PR, as it stands, is not quite ready to merge; the reviewer can, of course, respond "approved assuming the `squash:` commits are squashed.", just as we sometimes do when a PR needs minor changes, is failing a unit test, or any other pre-merge check.

inventhouse (changes requested):

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: [Note: The following was written assuming commitlint becomes required at some point.]

**Option 1: commitlint fails for `squash:` commits, "Squash and merge" is _disabled_:**

Pros:

- `squash:` commits can't be merged.
- Master can't be broken with a bad message from "Squash and merge".

Cons:

- Authors must squash locally to get a green build.
  - This adds another round of running tests, which is slow on many cases.
  - Re-running tests puts pressure on authors to squash early, but squashing early wastes reviewer time if they want to see diffs for "minor" final changes.
  - Github's popular "Squash and merge" option would need to be disabled for commitlint to become mandatory. 
  - Minor annoyance to both author and reviewers (in my opinion) that any PR that requires squashing must be red throughout the review process because squashing should be the final step. 
- Even if we add `squash:` to the OEP, there would be no linting to ensure consistency on this particular option.

**Option 2: commitlint fails for `squash:` commits, "Squash and merge" remains _enabled_:**

Pros:

- Developers can continue to use Github's "Squash and merge".
  - Tests don't need to be re-run just to squash. 
  - Authors are not incentivized to squash early, enabling a more efficient reviewer experience.
  - Reviewers and authors can get to a green build earlier in the process that covers most cases. 
- A green commitlint signifies all commits conform to the OEP and provide meaning.

Cons:

- It remains possible that `squash:` commits could be merged.
  - Although not ideal, the commit prefix probably still provides enough meaning when this does happen.
- Master could be broken with a bad message using "Squash and merge"
  - Again, although not ideal, this is similar to not forcing every PR to be up-to-date with master before merging. In the case of not requiring a rebase off master, we trade-off developer efficiency for the common case against ensuring that master can't be broken with a PR that would have failed unit tests had it been re-based ahead of time.

I prefer Option 2, but there are pros and cons to both options.

inventhouse: When or whether we start requiring commitlint for merging and whether we would then continue to allow squash-and-merge if we do are separate from whether commitlint should flag commits explicitly marked as do-not-merge or should pass in spite of them.

Do we really consider `squash:` commits to be "good enough to merge"?

As a reviewer, I want all outstanding issues that can be automatically caught to be spelled out by the pre-merge checks.  Often, it is fine to review a PR before every issue has been resolved, but if there's lint issue, or a unit test that needs to be finished or updated before merging, I want those to be flagged; if there are commits whose only meaning is "this should be squashed away", I want those to be flagged too.  It's fine to say "approved pending...", and in this case, it's fine for the author to resolve these commits with squash-and-merge.

As a reviewer, I want all-checks-passing to be as strong a statement about the PR's readiness to be merged as possible, I would _much_ rather parse failing checks than have to second-guess the actual readiness of a passing check.

**THE** big "con" for option 2 as outlined above is that all-checks-passing would not indicate whether a PR is actually ready to merge (unless we decide that `squash:` commits _are_ good enough to merge); "getting to a green build sooner" is not a "pro" if the PR still has outstanding issues.

But, again, we do not currently require commitlint to pass for merging, and changing that is a much bigger discussion with much wider implications.  I am advocating for...

**Option 3: commitlint fails for `squash:` commits, "Squash and merge" remains enabled for the time being:**

Pros:
- Developers are made aware that they shouldn't use other types for commits that are not intended to merge to the main branch.
- Commits that are not truly meaningful are much less likely to be merged to the main branch.
- Developers can continue to use Github's "Squash and merge".
- Tests don't need to be re-run just to squash.
- Authors can choose whether to interactively rebase their changes into commits that capture complete changes or leave them to squash later if that would be easier to review.
- Reviewers and authors can see exactly what issues should be resolved before merging.
- All-checks-passing makes a strong statement about the PR's readiness to merge.

Cons:
- Reviewers may be more frequently reviewing PRs before all checks pass.
- It remains possible that `squash:` commits could be merged.
    - Although not ideal, the commit prefix probably still provides enough meaning when this does happen.
- Master could end up with a bad message edited using "Squash and merge".
    - Again, although not ideal, this is similar to not forcing every PR to be up-to-date with master before merging. In the case of not requiring a rebase off master, we trade-off developer efficiency for the common case against ensuring that master can't be broken with a PR that would have failed unit tests had it been re-based ahead of time.

robrap: @inventhouse: Agreed. I was also about to write a comment stating that I'd update this PR to remove language around how linting will be performed, and move that discussions elsewhere.

This debate can be taken up on the linter before or after it becomes required. Also, it would be nice to have a global decision, but the decision could be different for different repos.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -88,6 +88,8 @@ We use these commit types labels:
 
 * **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
+* **squash**: temporary changes that are intended to be squashed before merged.
```
inventhouse, line 91: I would say "before merging" or "before they are merged"

oeps/best-practices/oep-0051-bp-conventional-commits.rst
inventhouse: I think this is an important discussion, but I agree that this OEP is about the format and types, and specifically not about the tooling, marking this as resolved.

inventhouse (approved):

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -148,6 +150,8 @@ Choosing the commit type label to use for a commit is important, but it's only t
 
 **feat vs fix**: some user-visible changes to features could be classified as "feat" or "fix".  Choose "feat" if the change adds to the set of features.  Choose "fix" if the change affects how a feature behaves.  Yes, this is still subjective.
 
+**squash vs temp or fix**: commits that are intended to be squashed before merge should use the "squash" prefix. Although "temp" or "fix" may feel appropriate, they are not meant to be used in this way and convey something meaningful after a PR merges, where "squash" commits are not intended to be merged.
```
inventhouse, line 153: You might also draw the distinction that `temp:` is intended to mean that the _change_ is temporary, whereas `squash:` means that the _commit_ is temporary but the changes it contains likely are not and will be incorporated into a larger commit with a different semantic type.

robrap: I like this language and updated, but made additional adjustments.

inventhouse: I like it!  Resolved.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -207,11 +211,13 @@ Tooling
 
 One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
 
-
+As we trial conformance checking, one open question is whether or not "squash" commits should pass or fail commit linting. See an initial discussion of `linting squash commits in this PR comment thread<https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683>`__.
```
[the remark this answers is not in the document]
ghassanmas: I think it needed to be added to here, the "squash".
 https://github.com/edx/edx-lint/blob/3584abab13b3d5d1f93301119da3aded81e51bb6/commitlint.config.js#L62-L67 
 Hence commitlint test is failing now and since @robrap  is using it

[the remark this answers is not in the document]
ghassanmas: May be to change here as well: https://github.com/edx/edx-lint/blob/3584abab13b3d5d1f93301119da3aded81e51bb6/edx_lint/files/commitlint.config.js#L5-L10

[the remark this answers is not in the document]
inventhouse: Our pre-merge check validates that a PR is ready to merge to the main branch, as such (I contend in the linked thread) no tooling change is needed as the current commitlint check does not accept this prefix and flags it as not-for-merging unless/until it is squashed away.

If we were using a pre-commit check, that would need to allow `squash`, but the pre-merge check still (I argue) should not.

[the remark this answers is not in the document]
timmc-edx: If commitlint were made to be a blocking check, then I would need to squash in order to merge -- but then that would also invalidate all the other checks and I'd have to wait another 20-30 minutes for all of those to run before I could actually merge.

If commitlint is not blocking... then I can just squash-merge and fix the commit title as I do it.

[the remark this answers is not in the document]
inventhouse: Moving commitlint to a blocking check is a much bigger discussion with much bigger consequences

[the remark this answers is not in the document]
robrap: We've intentionally split the tooling question from this PR. Please see https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683 if you wish to add to the debate on how the new "squash" label should should interact with linting.
1. If you have new pros/cons for either option, feel free to add a comment on the other thread.
2. If you simply want to vote for the option you prefer, feel free to add a comment on the other thread.

**Note:** Before this change, I personally was using "temp" and "fix" instead of "squash", and thought these were the correct labels to use. With this change, I would stop doing that and I think the intention around my commits would be more clear. I still have questions about tooling, but I think either way we go, this feels like a step in the right direction.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: @inventhouse: This thread is no longer marked as resolved so it is easier for others to see and contribute to. Others clearly have thoughts on this topic, even though we are hoping to merge this PR without having to make a final decision on this separate topic (i.e. whether to pass or fail linting on "squash" commits).

oeps/best-practices/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
timmc-edx: Sorry, yes, I got turned around—deferring the question of linting behavior sounds good. Having a special "blessed illegal value" seems fine. :-)

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -88,6 +88,8 @@ We use these commit types labels:
 
 * **revert**: undo a previous change. The previous commit message, including the type label, is included.
 
+* **squash**: temporary changes that are intended to be squashed before merging.
```
timmc-edx, line 91: I might suggest not having this as part of the list, but as an addendum below the list, so that it's not seen as a "regular permitted value" by someone reading the list quickly.

robrap: You've hit on the ugly truth of pretending that the linting question is separate, when we in fact are actively linting. I agree that it would be best to clarify expectations. I could add something like:
```
  * **Note**: At this time, "squash" commits are a convention, but will still fail the build. See the open questions section if you believe these should pass the build.
```
In the future, we'd remove this note altogether if we chose to make it pass, or we could update the note to:
```
  * **Note**: Although "squash" commits are a convention that convey meaning, they will still fail builds to ensure these commits are not merged.
```

robrap: Actually, here is a potential update:
```
* **squash**: temporary changes that are intended to be squashed before merging. This label is a convention that conveys meaning, but it will still fail the build. See the tooling open questions section for the pass/fail debate.
```

Note: I liked using "temp" because it would pass the build, and I'm not sure that people won't continue to do so. I can't promise I won't while this is still under debate.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
inventhouse: Might it be possible to have commitlint _warn_ (but not fail) for `squash`?  (This would, of course, still leave unanswered the question of “what does a required commitlint check even mean if the message can be completely rewritten after the check)

oeps/best-practices/oep-0051-bp-conventional-commits.rst
inventhouse: I added a comment about warning instead of failing to the linting thread above

inventhouse: (As for using other tags just to make the build “green”, I still say that’s just like adding `pylint disable` instead of actually fixing the warning, or `assert True #FIXME` to a unit test)

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: 1. Warning sounds more appropriate than error.
2. The required check still ensures that the author knows about and is generally using conventional commits. If it is rewritten afterwards, it is likely to use an existing comment that already conformed. Could some intentionally cause a problem? Sure, but this seems like a very low risk.
3. It is important to note that the pass/fail decision is easily reversible. The only issue is managing the change in expectations.

inventhouse: 2) “could someone _make a mistake_” is the more likely issue, and, again, how is a non-binding “required” check actually different than than the current state?

robrap: The goal is to get people to write  conventional commits. Ned provides data about the growth of their usage, but we don’t have full compliance. I bet making it a non-binding “required” check would spike this data and give us what we are looking for. Also, because it would minimally disturb the typical developer workflow, we could more quickly make it required without upsetting developers, and more quickly get compliance on using conventional commits.

@nedbat: I’d love to hear where you stand in this given your goal to get people using conventional commits.

pshiu: Jumping in late, but it sounds like this PR is tackling the problem of devs taking the path of least resistance[^1] by using inappropriate commit prefixes.

The proposed solution is to (a) Create a "blessed illegal" commit prefix, as Tim well put it, and (b) exercise one of Options 1–3 to make commitlint and our GitHub merge strategies create a better path of least resistance.

Very controversial opinion[^2]:

**Option 4**: Make `temp` harder to use.

This can be done by requiring the word "until" in the commit message and having very good commitlint console output for the dev.

As a dev, I prefer being taught the right way than having to keep track of one more exception in my head.[^3]

For `fix`: Odds are, it's probably being used correctly; it just also happens to be not intended for merging.

And that way, with Option 4, we don't need the `squash` merge prefix.

[^1]: This is a good thing! Developers should not be spending thinking time taking paths of non-least resistance. Plus, it provides architects the challenge of crafting the path of least resistance so the commit prefixes are useful.

[^2]: This is also a _very_ weak opinion. I'd be happy to go either way as a dev.

[^3]: Fair, having to use "until" only for `temp` is an exception as well, but it is an exception which is crafted to help the dev adhere to a larger principle, rather than create an exception to the principle.

pshiu (approved):

oeps/best-practices/oep-0051-bp-conventional-commits.rst
inventhouse: @pshiu `squash` has a different meaning than `temp`:

> **squash vs temp or fix**: use "squash" when the *commit* is temporary, and you intend to squash the commit into a larger commit with a different semantic label. Use "temp" when the *change* is temporary, but you intend to merge the commit. Again, use "squash" over "fix" when the *commit* is temporary and is not intended to be merged.

Though I like the idea of expiration dates on `temp`, that'd be a different proposal for a different problem

inventhouse: Looks like it would be pretty straightforward to create a custom rule to warn for `squash` commits: [Commitlint Local Plugins](https://commitlint.js.org/#/reference-plugins?id=local-plugins)

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -207,11 +211,13 @@ Tooling
 
 One of the advantages of formalized commit messages is using them as input to tooling and conformance checkers.  We will investigate tooling in the future, and are making no recommendations now.
 
-
+As we trial conformance checking, one open question is whether or not "squash" commits should pass or fail commit linting. See an initial discussion of `linting squash commits in this PR comment thread<https://github.com/edx/open-edx-proposals/pull/254#discussion_r741051683>`__.
 
 Change History
 ==============
 
+2021-11-01: Squash commits.
```
feanil, line 219:
```suggestion
2021-11-01: Add a new ``squash`` commit type.
```

oeps/best-practices/oep-0051-bp-conventional-commits.rst
feanil, line 91: This seems un-necessary.  Git already has a notion of indicating that a commit is meant to be squashed into another(`squash! <title of commit to squash into>`), and that commit title already fails our lint testing so it achieves both the communicating of intent(better than the squash type in my opinion since the title can be used to map back to where it was meant to be squashed) and fails tests like the new squash type would. 

The value of making this change seems to be that "temp feels wrong" when we want to indicate other information?  Perhaps what we need is to say that normal git notation around fixups and squashes are allowed in PRs but we should aim to resolve them before putting commits onto our main branches.

robrap: 1. Can you point to docs for "squash!"? I'm not familiar with it.
2. Your comment "This seems un-necessary" seems at odds with your help with how you would do this. Maybe you are saying it is unnecessary to invent a new method of doing this, and we should instead document that we want people to follow this git convention?
3. My original intent was to have a convention that both conveys meaning **and** got me a green build. I'll admit that without the promise of a green build, this doesn't seem worth pursuing to me. And as noted in my pros/cons, I think there are trade-offs with pass/fail on this, and don't think failing is the only valid solution.
4. I'll admit that I was happily using "temp" and "fix" for these commits, and I didn't think I was doing anything wrong, and I was happy to have a green build as I wanted. I now see that if you read the definition of "temp", then you would know that was not its intent. For "fix", these temporary commits often are fixing something on the PR. Even "refactor" would work for some of these commits. Note that if I fix or refactor something in a follow-up PR, these are exactly the labels I would use, and the question is whether those are only valid labels in a follow-up PR, and not in an in-progress PR.

feanil: 1. https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---squashltcommitgt
2. Yes, I think it's un-necessary to invent a new way.
3/4. I think the goal of these labels is to help make the commits clear.  It's up to the rest of the words in the commit message to support this initial word and you can always be more clear in those words.  I think it's fine to use fix/refactor in the same PR as the thing that they are fixing.  Commits don't know about PRs and are just referencing things that happened earlier in the commit history.

inventhouse: Certainly a fair take.  On the other hand, I like:
- Keeping and working with the CC format habits
- Using a title directly related to this (sub-) change, rather than the original commit's title
- I think this will make it easier to write a custom commitlint rule that can warn for these without blocking if we decide to make the commitlint check "required" but still allow squash-and-merge, which could make that transition more palatable.

robrap: Thanks @feanil and @inventhouse. I agree with @inventhouse's response. 

For those developers that use `autosquash`, I'm not sure if the tooling (commit linting, auto change logs, etc.) could also handle that well, and maybe that could be taken up as an additional alternative path for those that want to use `autosquash` **and** want a green build. If you are ok with a failed build, you can really use whatever you want.

nedbat:
Sorry, i just merged a change that moved the Tooling section...

oeps/best-practices/oep-0051-bp-conventional-commits.rst
nedbat: I was also realizing this morning that git already has customs and tooling for detailed squashing (N->M squashes).  @robrap/@inventhouse I don't know what style of commit messages you are envisioning here.  How much detail is in the "squash:" messages?

inventhouse: @nedbat honestly probably not a lot, usually just `squash:` and a title.  Also, I'm likely to experiment with the git tooling that was referenced above.

My main interest in this is around flagging commits that are not intended to be merged as-written and especially helping folks use the tooling to track that; I'd be quite happy with just a mention that using git's tooling and formatting or one's own convention on development branches is a good way to do this with the linting we do today and discourage folks from using `temp`, etc. for changes that are not meant to be temporary or commits that are not meant to be merged.

If/when we discuss ratcheting-up the commitlint check to be required, we can discuss / investigate custom rules for non-blocking warnings and such if we like.

robrap: @nedbat: At this time, as someone who already uses conventional commits, I find the failures annoying and will likely not look at them because they are just crying wolf about the squash commits I already know about. If I happen to mis-type or misremember one, and there is a real error, I probably will miss it.

I am open to any formatting for squash commits that would quiet the linter. If we aren’t quieting the linter, none of this really matters much to me. 🙂

inventhouse: With what I said above, I would be interested in pursuing custom warning rules to help developers track the commits they do not intend to merge as-written _while also_ catching genuine errors and reducing the incentive to mis-use accepted type tags.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: @inventhouse: I'm not sure if we'd need a custom rule. Wondering if something like this would work: https://github.com/edx/edx-lint/pull/215/files

inventhouse: @robrap I don't expect you're allowed to have two rule entries with the same key; I think it would have to be a different rule name.

robrap: @inventhouse: I'm guessing you are right, but it also shows the pattern for how we could use the custom rule. I wasn't clear on that at first.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: https://github.com/edx/edx-lint/pull/215 would make "squash" a warning, and not an error, which you can see in action on that PR. 

Are there any disagreements with this direction?

FYI: @nedbat @inventhouse

oeps/best-practices/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
robrap: After further discussion:
1. It was too difficult to separate these concerns, because after reading about "squash" commits, it would be nice to provide engineers and expectation if this really would still fail their builds, which would be odd as compared to all the other types.
2. It may be palatable to everyone to make "squash" commits warn rather than error, but not fail the build. See https://github.com/edx/open-edx-proposals/pull/254/files#r743981426 for discussion of that change in action.

requirements/base.txt
```diff
@@ -24,11 +24,11 @@ jinja2==3.0.2
     # via sphinx
 markupsafe==2.0.1
     # via jinja2
-packaging==21.0
+packaging==21.2
     # via sphinx
 pygments==2.10.0
     # via sphinx
-pyparsing==3.0.1
```
robrap, line 31: I'm not sure what is up with this downgrade, so I'll try to fix.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
robrap: It looks like @timmc-edx also demonstrated that commitlint is already configured to ignore `squash!` (i.e. it passes **and** provides no warnings). 

I propose we just document behavior of `squash!` and allow it and move on. This simplifies docs and custom code.

I would drop the separate `squash:` type and custom `edx-lint` code because we would just have more stuff to maintain.

Thoughts?

oeps/best-practices/oep-0051-bp-conventional-commits.rst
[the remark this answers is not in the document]
robrap: Added more details to above comment, so switch threads to stay current. :)

robrap:
I'm closing in favor of documenting existing ignore rules in https://github.com/edx/open-edx-proposals/pull/258.

PR #284, opened 2022-01-13
[OEP-1] Community-based Decision Making for Arbiter Selection

sarina:
In the original doc, the process for getting an Arbitrator is to have one assigned by an internal edX team (I believe, at least - the wiki page linked to in the OEP is not publicly available). Moving forward, no one community member should have the ultimate authority in community decision making, not even tCRIL; I don't want tCRIL to just step in and take edX's place. Instead, I'd much rather a solution that allowed the whole community to participate in decision making, and this PR proposes a way for that to happen.

Additionally, I think it's reasonable to codify the fact that many of us are organically finding Arbiters due to us having domain expertise and knowing who in the community is out there working on the same or similar problems and would be good to arbitrate. Therefore I've written that part in, as a nod to existing practice. If we should not be doing this, I'll remove that point but I would like to add in some reasoning as to why that process should be followed.

docs: Clarify process of choosing Arbiters
* Codifying that choosing an Arbiter, in practice, is done by the OEP Author(s)
* Remove authority for assiting with arbitration and the overall process from
  the edX internal architecture team to the Open edX community architecture group

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -92,9 +92,8 @@ forums, and attempts to build community consensus around the idea.
 Arbiter
 -------
 
-Each OEP also has an Arbiter (as described in `Step 1. Request an Arbiter`_).
-The Arbiter will be chosen by the `edX Architecture Team`_. An Author of an OEP
-cannot concurrently be the Arbiter of that OEP.
+Each OEP also has an Arbiter (as described in `Step 1. Find an Arbiter`_).
+An Author of an OEP cannot concurrently be the Arbiter of that OEP.
 
 The Arbiter will be the person making the final decision on whether the OEP
 should be Accepted, and as such, the Arbiter should be knowledgeable about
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0001.rst
```diff
@@ -133,14 +139,18 @@ OEP Workflow
...
+When writing an OEP, you may already have an idea of an Arbiter in mind. If so,
+reach out to that person and ask them; they should have the domain expertise
+needed to be an effective Arbiter and the time to do so. It is best practice for
+the Arbiter to be from a different team or group than the author.
 
-.. _`arch-team@edx.org`: mailto:arch-team@edx.org
+If you're not sure who would make a good Arbiter, you should post in the
+`Architecture Group`_ category on the Discourse forum or the ``#architecture``
+channel in the `Open edX Slack`_. Once found, this Arbiter will be
+recorded in the "Arbiter" header on the OEP.
```
feanil, line 153: Should we include anything about what to do if you disagree with an arbiter choice?  Having some text to guide people if they have concerns about a specific arbiter, might be useful as a way of mediate issues.

An initial stab at what it might look like:
"If you have concerns about an abiter that has been chosen for a particular OEP, please share your concerns with the author first and see if you can resolve your concerns directly.  If you continue to have concerns, please share them in slack or discourse..."

feanil (approved):
Looks good without addressing my suggestion.  Feel free to update or merge as is, happy to discuss further if you want.

oeps/processes/oep-0001.rst
sarina: Thanks. I added this, and also a link to our code of conduct (where you can get in touch w/ specific people if you have major concerns)

PR #285, opened 2022-01-13
[OEP-14] Change internal edX procedures to community-based ones

sarina:
Moving forward, we should make sure that OEPs contain procedures that are relevant to the whole community. When specific community members have practices that are relevant just to them, they should find a way to document those extra steps for their team.

### DIFF
--- a/oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
+++ b/oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
@@ -7,7 +7,7 @@ OEP-14: Archiving edX GitHub Repositories
 +---------------+----------------------------------------------------------+
 | Title         | Archiving edX GitHub Repositories                        |
 +---------------+----------------------------------------------------------+
-| Last Modified | 2022-01-10                                               |
+| Last Modified | 2022-01-26                                               |
 +---------------+----------------------------------------------------------+
 | Author        | Christina Roberts <christina@edx.org>                    |
 |               | Feanil Patel <feanil@edx.org>                            |
@@ -21,27 +21,25 @@ OEP-14: Archiving edX GitHub Repositories
 | Created       | 2017-01-09                                               |
 +---------------+----------------------------------------------------------+
 | `References`  | `ORA PR Discussion`_,                                    |
-|               | `Initial Archiving Discussions`_,                        |
 |               | `Moving More Repos to edx-unsupported`_                  |
 +---------------+----------------------------------------------------------+
 
 .. _ORA PR Discussion: https://github.com/edx/edx-ora/pull/187
-.. _Initial Archiving Discussions: https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
 .. _Moving More Repos to edx-unsupported: https://openedx.atlassian.net/browse/ARCHBOM-1481
 
 Abstract
 ========
 
-The `edx organization`_ contains a large number of repositories, most of which are active and maintained, but some of which are now obsolete. To clarify the status of repositories, a process for archiving a repository is defined below.
+The `openedx organization`_ contains a large number of repositories, most of which are active and maintained, but some of which are now obsolete. To clarify the status of repositories, a process for archiving a repository is defined below.
 
-.. _edx organization: https://github.com/edx
+.. _openedx organization: https://github.com/openedx
 
 Motivation
 ==========
 
 Recently openedx.yaml files were added to edX repositories per `OEP-2`_. In the course of deciding owners for those repositories, there was an `ORA PR Discussion`_ about how best to handle deprecated or obsolete repositories. In particular, do obsolete repositories need owners, and how can repositories be clearly marked as present for archive purposes only?
 
-This discussion resurfaced related to edX's usage of Gemnasium to report the usage of third-party libraries that have known security issues. All repositories under the edX organization were being monitored, but this added noise when trying to understand the number of third-party library updates required for actively maintained repositories.
+This discussion resurfaced related to edX's usage of Gemnasium to report the usage of third-party libraries that have known security issues. All repositories under the Open edX organization were being monitored, but this added noise when trying to understand the number of third-party library updates required for actively maintained repositories.
 
 .. _OEP-2: https://open-edx-proposals.readthedocs.io/en/latest/oep-0002.html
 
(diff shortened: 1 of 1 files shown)


oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -27,31 +27,34 @@ OEP-14: Archiving edX GitHub Repositories
 
 .. _ORA PR Discussion: https://github.com/edx/edx-ora/pull/187
 .. _Initial Archiving Discussions: https://openedx.atlassian.net/wiki/display/IT/Proposed+Github+Deprecation+Process
+
+**Code Review: This link is private. Could someone move it to the COMM space or suggest a better alternative?**
```
sarina, line 31: Code reviewers: Looking for suggestions here

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -65,27 +68,30 @@ First, if the repository is public, and a part of Open edX releases, follow thes
 
 2. If someone does wish to take ownership of the repository, email the internal edX developers mailing list to see if there are any objections. If there are no objections after 2 work days, post to `Open edX Deprecation Announcements`_ that the transfer will take place.
 
-3. Create an IT help ticket for the repository to be transferred to the new organization.
+**Code Review: How could we make the above internal-to-edX step something public? I think publicizing on the deprecation announcement forum and on Slack (I suggest #general and #depr-slash-n-burn) should be enough; organizations are welcome to do more, but each org's processes shouldn't be reflected in an OEP.**
+
+3. Create a new `tCRIL engineering task`_ for the repository to be transferred to the new organization.
```
sarina, line 73: @kdmccormick I think this is right, what do you think?

kdmccormick: @sarina Maybe `GitHub Request on the tCRIL board` (that way gets auto-assigned and a `github-request` label) instead of a general tCRIL task, but otherwise looks solid.

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -27,31 +27,34 @@ OEP-14: Archiving edX GitHub Repositories
...
+
+**Code Review: This link is private. Could someone move it to the COMM space or suggest a better alternative?**
+
 .. _Moving More Repos to edx-unsupported: https://openedx.atlassian.net/browse/ARCHBOM-1481
 
 Abstract
 ========
 
-The `edx organization`_ contains a large number of repositories, most of which are active and maintained, but some of which are now obsolete. To clarify the status of repositories, a process for archiving a repository is defined below.
+The `openedx organization`_ contains a large number of repositories, most of which are active and maintained, but some of which are now obsolete. To clarify the status of repositories, a process for archiving a repository is defined below.
```
nedbat, line 38: Can we write this OEP so that it can be applied to many GitHub organizations?  I'd hate to see this apply only to openedx, and then have edX either have no process, or have to copy this document to have a process.

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -27,31 +27,34 @@ OEP-14: Archiving edX GitHub Repositories
...
 
 
 Specification
 =============
 
-When a repository under the `edx organization`_ will no longer be maintained by edx Inc. because it is no longer in use by the latest version of the Open edX platform, the following steps should be followed.
+When a repository under the `openedx organization`_ will no longer be maintained by the Open edX community because it is no longer in use by the latest version of the Open edX platform, the following steps should be followed.
 
-This process is not for repositories that are currently still in use by either the latest version of the Open edX platform or by edX Inc. the company.  If a repository is in use, but planned to be removed, it should go through the `deprecation process`_ and when that is completed it can be archived as described by the process below.
+This process is not for repositories that are currently still in use by either the latest version of the Open edX platform.  If a repository is in use, but planned to be removed, it should go through the `deprecation process`_ and when that is completed it can be archived as described by the process below.
```
nedbat, line 57: I understand why you removed "or by edX", but we need to keep that idea in this OEP.  How about, "still in use by either the latest release or the master branches of the Open edX platform"?

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
dianakhuang: I think edX/internal deprecations tend to be handled via our Jira boards. The DEPR process for better or worse is primarily a means for announcing deprecations to the wider community.

sarina: @nedbat I would be comfortable adding a line like, "we encourage others to follow these guidelines", but I'm not sure how I feel about an OEP specifically calling out a community member to say, hey this is your guideline too. Why can't edX just use this guideline? Maybe there's an internal wiki page that says "Hey we follow this oep" and points here.

nedbat: I'm not suggesting to call out a community member.  I want edX to use this guideline.  I'm just suggesting using broader language than "The openedx organization".  How about "The Open edX project, including the openedx GitHub organization, contains a large number ..."

sarina: I feel that's saying that every Open edX project must use this process...?

kdmccormick: In my mind "the code of the Open edX project" is equivalent to the code in the `openedx` GitHub organization. There are certainly Open edX plugins, tooling, config overrides, etc other GitHub orgs, but I don't think they need to be governed by Open edX's own project policies. Like @sarina said, if edX wants to opt-in to this OEP for their GitHub org, that's awesome, but I wouldn't want it to seem like we're trying to tell all Open-edX-associated organizations how to manage their repositories.

There are definitely some Open edX project policies that could span multiple GitHub orgs - the Code of Conduct comes to mind. But that's a social policy, whereas this is a very tactical OEP about how exactly to archive a repository.

nedbat: OK, we can adopt the OEP by reference.

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
nedbat: I've requested to move the page (https://openedx.atlassian.net/servicedesk/customer/portal/1/ITSUPPORT-17691 though you won't be able to see that ticket).  It has four proposals for how to deprecate repos, maybe we don't need the history?

sarina: Yeah, since I can't see the page I'm not sure what the intention was of linking it. Perhaps it was simply having the history of considered and rejected alternatives?

feanil (approved):

feanil:
I think this is good to land, we can always make more improvements and updates in the future.

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
ormsbee: I think having history is nice in general, but it doesn't have to block this PR from merging.

ormsbee (approved):

PR #275, opened 2021-12-23
OEP-54: Core Contributor Program

sarina:
This document defines the Core Contributor (CC) Program, a program that grants rights to individuals
in the Open edX community that allow them to actively participate in defining and deciding the
direction of the platform.

PR #292, opened 2022-02-01
OEP-54: Procedures for removing Core Contributors

sarina:
In https://github.com/openedx/open-edx-proposals/pull/275 it was pointed out right at the end of review that there was no caveat for removing Core Contributors. This presents a set of criteria for how that may happen.

### DIFF
--- a/oeps/processes/oep-0054-core-contributors.rst
+++ b/oeps/processes/oep-0054-core-contributors.rst
@@ -9,13 +9,13 @@ OEP-54: Core Contributors
    * - Title
      - Core Contributor Program
    * - Last Modified
-     - 2022-01-26
+     - 2022-02-04
    * - Authors
      - Sarina Canelake <sarina@tcril.org>
    * - Arbiter
      - Xavier Antoviaque <xavier@opencraft.com>
    * - Status
-     - Under Review
+     - Accepted
    * - Type
      - Process
    * - Created
@@ -52,10 +52,9 @@ The Open edX Core Contributor Program grants specific types of access rights to
 Open edX community members, which enables them to further the development and
 adoption of the Open edX platform. CCs earn these rights by sustained, active
 participation in the Open edX community and by following the “3 Cs”: Commitment,
-Conduct, Caliber. Core Contributors follow the `Code of Conduct
-<https://open.edx.org/code-of-conduct/>`_, are respectful of other community
-members, are polite and welcoming to all in the community, and have shown the
-ability to deliver high quality work over time.
+Conduct, Caliber. Core Contributors follow the `Code of Conduct`_, are
+respectful of other community members, are polite and welcoming to all in the
+community, and have shown the ability to deliver high quality work over time.
 
 Core Contributor Roles
 ----------------------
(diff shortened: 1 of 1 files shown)


sarina:
Requesting review from any/all who reviewed the initial PR: @felipemontoya @pomegranited @nedbat @davidjoy @pdpinch @antoviaque

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -267,6 +265,37 @@ discussion category
...
+* A CC is not upholding the standards of the community
+
+Trigger events for removing a CC would be:
+
+* A CC informs the Program Administrators that they can no longer continue in
+  the role
+* A CC has been inactive and unreachable for more than 6 months
+* A CC hasn't met project quality standards consistently despite feedback
+* A CC has consistently shown poor judgement, such as merging PRs without
+  approval
```
pomegranited, line 287: CCs can approve PRs and merge them without outside approval, so maybe reword this?

```suggestion
* A CC has consistently shown poor judgement, such as merging PRs without due consideration
```

pomegranited (approved):
Minor recommendation, but the rest LGTM! Thanks for adding this @sarina .

👍 
- [x] I tested this by reading the rendered version.
- [x] I checked for formatting, clarity, and correctness.

oeps/processes/oep-0054-core-contributors.rst
sarina: Ah, great catch. My intention was something along the lines of "merging PRs without addressing outstanding concerns" - your wording works for me.

feanil (approved):

.gitignore
```diff
@@ -7,6 +7,7 @@ _build
 .project
 .pycharm_helpers/
 .pydevproject
+.vscode/
```
nedbat, line 10: If you put this in a global gitignore locally, you won't have to add it to individual repos.

sarina: Ah ok, sounds good, I only use this editor for oeps but i will remove this, sorry

nedbat (approved):

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -267,6 +265,37 @@ discussion category
...
+
+Trigger events for removing a CC would be:
+
+* A CC informs the Program Administrators that they can no longer continue in
+  the role
+* A CC has been inactive and unreachable for more than 6 months
+* A CC hasn't met project quality standards consistently despite feedback
+* A CC has consistently shown poor judgement, such as merging PRs without
+  due consideration or addressing outstanding concerns
+* A CC has violated the project `Code of Conduct`_
```
antoviaque, line 288: @sarina I'm coming a bit late for the review here it seems -- but what about the cases where core contributors are somewhat active, but don't fulfill the commitments made? With just a vague "activity" requirement, don't we run the risk of ending up with many core contributors who don't do much?

sarina: We discussed this at tCRIL and also I did some reading on healthy communities. Having a core contributor who contributes a few patches a year, or only does review, or reviews translations only once per month or quarter... well, that work _is_ valuable. And just because they may be very busy right now does not mean they will be in the future.

We do say below,
> Program administrators will be responsible for determining when a CC is inactive/unreachable, by doing a check at least once per quarter.

I think as part of doing the check we will begin to understand if this concern is warranted or not. As is my prerogative, I tend to not want to overly-solve for situations until we encounter the situation - with context - and figure out where to go from there.

antoviaque: @sarina OK, we can see how that goes. But imho we are in a special case of a community, where a lot of the parties deciding on where/how much work is contributed are companies -- not the core contributors themselves. And the goal with this is to help the core contributors getting that time allocated by their bosses, by providing a concrete incentive for the company itself (the claim of having X core contributors). Some companies will play fair regardless, but imho for others if they can get the outcome without having to prioritize it over other company priorities, they won't - and we'll have core contributors saying "we have been busy at our company, so I haven't been able to spend on this".

antoviaque: Submitted at https://github.com/openedx/open-edx-proposals/pull/304

PR #293, opened 2022-02-02
docs: clarify new CC member voting for OEP-54

ormsbee:
Our first Core Contributor nomination under OEP-54 guidelines was:

  https://discuss.openedx.org/t/core-contributor-nomination-maria-grimaldi/6548

The clarifications in this commit are a result of feedback and
observations during this nomination process.

### DIFF
--- a/oeps/processes/oep-0054-core-contributors.rst
+++ b/oeps/processes/oep-0054-core-contributors.rst
@@ -187,6 +187,20 @@ The process for nomination is as follows:
 
 #. During the comment period, other CCs should weigh in on the thread. As much
    as possible, posts should conclude with a definitive "yes" or "no" vote.
+   "Liking" the nomination post does not count as a vote.
+
+   #. Any Core Contributor may vote on any CC nomination, regardless of their
+      contributor role.
+
+   #. In the case of Core Contributor nominations that involve granting commit
+      access to repos, anyone who has commit access to any of the repos in
+      question may also vote.
+
+   #. All community members are welcome to provide comments on the thread.
+
+   #. CoC violations or other behaviors not upholding the "Conduct" quality of
+      the 3 Cs ought to be raised by anyone, whether in the Discourse thread or
+      to the Program Admins.
 
    #. If anyone has concerns they feel cannot be raised publicly, they should
       `Contact the Program Administrators`_ to determine how to proceed. The
@@ -214,9 +228,11 @@ The process for nomination is as follows:
 #. At the end of the comment period, a candidate is approved if there are at
    least 5 (five) affirmative "yes" votes, and zero "no" votes.
 
-   #. For candidates coming from an organization, at least one affirmative vote
-      must come from a CC outside of the candidate's org. This is to ensure that
-      core contributors are evaluated on their ability to communicate with the
+   #. Candidates coming from an organization must have support from someone
+      outside their organization. This can be done by having a sponsor from
+      outside their organization, or by receiving at least one "yes" vote from
+      someone outside their organization. This is to ensure that core
+      contributors are evaluated on their ability to communicate with the
       community, outside of their org.
 
    #. For roles with few existing members, it is encouraged to ask CCs of other
@@ -320,6 +336,11 @@ the ``#core-contributors`` Slack room in the `Open edX Slack
 
 Change History
 ==============
+
+2022-02-16
+----------
+
+Added clarifiations around the voting process for new CC members.
 
 2022-02-11
 ----------


oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -189,6 +189,10 @@ The process for nomination is as follows:
 
 #. During the comment period, other CCs should weigh in on the thread. As much
    as possible, posts should conclude with a definitive "yes" or "no" vote.
+   "Liking" the nomination post does not count as a vote.
+
+   #. Anyone with commit access to the relevant repos may vote. This may change
+      as the Core Contributor program evolves.
```
sarina, line 195: Hmm. I don't _fully_ agree that you have to have commit access to the repo. If you're a CC on other repos, but have say pulled into or reviewed the candidate's work previously, I feel you should be able to give a vote with explanation.

I absolutely _do not_ agree that you need commit rights to post a No vote, especially in cases of CoC violations or otherwise not upholding community values.

Finally, this doc is very explicitly written to be agnostic to CC roles. So I'd think about how you could rephrase. 

Combining all that feedback, maybe something like:

```
In the case of Core Contributor roles that involve granting commit access to repos, anyone who has commit access to the relevant repos may vote; additionally, all core contributors and indeed any community member are welcome to provide comments on the thread.
```

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -216,10 +220,10 @@ The process for nomination is as follows:
 #. At the end of the comment period, a candidate is approved if there are at
    least 5 (five) affirmative "yes" votes, and zero "no" votes.
 
-   #. For candidates coming from an organization, at least one affirmative vote
-      must come from a CC outside of the candidate's org. This is to ensure that
-      core contributors are evaluated on their ability to communicate with the
-      community, outside of their org.
+   #. For candidates coming from an organization, either the nomination or at
```
sarina, line 223: This doc uses the word "sponsor" to refer to the person who puts forth the nomination. Maybe...

...either the sponsor, or at least one of the affirmative votes, must come from...

ormsbee: Ah, yeah, definitely don't want to unnecessarily introduce new terminology. Will fix.

oeps/processes/oep-0054-core-contributors.rst
ormsbee: FWIW, I'm actually open to a wide range of possibilities here, so I ask the following only for clarification (and not because I'm trying to argue against it). Is this an accurate summary of your view?

1. Any member of the community is free to make comments in the thread, to either voice support or concerns.
2. Any core contributor to Open edX may vote, regardless of whether they are formally a part of the CC program today, or are de facto contributors (e.g. edX employees with merge rights).
3. It is up to the core contributors themselves to judge whether or not they are qualified to vote, based on the nature of their interactions with the nominee. If the link is not obvious, the CC should provide some explanation with their vote.

> I absolutely do not agree that you need commit rights to post a No vote, especially in cases of CoC violations or otherwise not upholding community values.

In this case, you'd still want them to be a core contributor of some sort to have a "no" vote, right? I'm interpreting this to mean, "I think it would be good for CCs to be able to vote 'yes' on a candidate even if those CCs aren't working on those particular repos (or work on entirely different things), but I might be convinced otherwise. In contrast, I believe _very strongly_ that any CC should be able to vote 'no' based on their community interactions with this person." I ask this clarifying question because I'm not sure whether you intended to make sure the "no" vote could be open to the full range of CCs, or whether you intended to open it up to the full community at large.

sarina: Great question, and summary overall that crystalizes my thinking. I think anyone in the community should be able to voice concerns (ex, a new dev who's not a CC who's had particularly sour interactions with the nominee). I think the sponsor should take all concerns seriously and follow up on allegations. To me, if I nominated someone and got feedback like that, I would consider it a "no" vote (although possibly pending asking for proof that this occurred). Or I might expect the sponsor to raise allegations like this to the program admins or via the [code of conduct escalation path](https://openedx.org/code-of-conduct/). I wouldn't approve the nominee until concerns were addressed sufficiently - perhaps it's a misunderstanding, perhaps it's a serious CoC violation, or somewhere in between.

Kind of a rambly paragraph. But I would hope that a non-CC raising a serious concern would be seen as a de facto No vote. I think also a No vote is a start of a conversation, and it's quite possible that concerns could be withdrawn via conversation.

sarina: Maybe it's something like, "only CC's get No votes, but CoC violations or other behaviors not upholding the "Conduct" quality of the 3 Cs ought to be raised by anyone, whether in the Discourse thread or to the Program Admins"

How does that feel?

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -189,6 +189,16 @@ The process for nomination is as follows:
 
 #. During the comment period, other CCs should weigh in on the thread. As much
    as possible, posts should conclude with a definitive "yes" or "no" vote.
+   "Liking" the nomination post does not count as a vote.
+
+   #. In the case of Core Contributor roles that involve granting commit access
+      to repos, anyone who has commit access to the relevant repos may vote;
```
sarina, line 195: In some cases this could be less than 5 votes. In some cases, there may only be people from the same organization. I can't think of anything better to say; I think we should monitor the situation and if _this_ arises, we can deal with it.

sarina (approved):

oeps/processes/oep-0054-core-contributors.rst
kdmccormick: What if access to multiple repos is being granted? Do all voters need access to all such repos?

And do we give non-commit-related access via the CC program that is worth mentioning here?

sarina: @kdmccormick 

> What if access to multiple repos is being granted? Do all voters need access to all such repos?

I read this as "at least one of" the repos. Right now, it would be extremely restrictive to say the subset has to match entirely. I'm not sure exactly what to address... do we need to then further specify that there must be not only five overall votes, but at least one affirmative vote for each repo being granted access to?

> And do we give non-commit-related access via the CC program that is worth mentioning here?

I'm not sure what you're referring to. In any case this bullet starts with, `In the case of Core Contributor roles that involve granting commit access`, and I think if other special cases in other roles arise, we can continue to amend this OEP.

kdmccormick: Sorry, I was reading this on a phone and responded hastily. I also hadn't read the good discussion you and Dave had [up here](https://github.com/openedx/open-edx-proposals/pull/293#discussion_r797956840). Disregard my comment, and let me try again in another thread.

kdmccormick (changes requested):
With the addition of:

> anyone who has commit access to the relevant repos may vote

I am confused as to who can vote when. From discussions on the PR it seems like it's:
* any  CC can Yes-vote, but we'd like if folks with repo access comprised some of those Yes votes.
* any CC can No-vote

But from the proposed text of the OEP, I read:
* only CCs with access to (some? all?) proposed repos can vote Yes or No.

I'm happy to suggest more specific edits, but as it stands I don't know what set of rules we want to land on.

oeps/processes/oep-0054-core-contributors.rst
antoviaque: Restricting the vote to the people who have access to the repo sends a signal to the other core contributors that their opinion is very secondary. It wouldn't encourage core contributors to look at applicants outside of their own responsibilities, and would imho greatly reduce the number of reviews new applicants would get. And with the current formulation, it even unclear whether a "no" from a core contributor who doesn't have access would count...

I think we need to separate the concerns here - I would suggest:
* All core contributors can vote, regardless of commit rights: a "yes" counts against the 5 minimal votes, and a "no" is a veto vote, regardless of which core contributor posted it
* BUT to get access to a specific repository, at least one of the (positive) votes should come from a core contributor having access to that repository. If there are multiple repositories involved, these votes can come from multiple core contributors having different access.

kdmccormick: Hey @antoviaque - I am going to resolve this and then copy your comment to the PR itself so that we are all talking in one thread.

kdmccormick:
@antoviaque 's comment, copied in from another thread:
> Restricting the vote to the people who have access to the repo sends a signal to the other core contributors that their opinion is very secondary. It wouldn't encourage core contributors to look at applicants outside of their own responsibilities, and would imho greatly reduce the number of reviews new applicants would get. And with the current formulation, it even unclear whether a "no" from a core contributor who doesn't have access would count...
> I think we need to separate the concerns here - I would suggest:
> * All core contributors can vote, regardless of commit rights: a "yes" counts against the 5 minimal votes, and a "no" is a veto vote, regardless of which core contributor posted it
> * BUT to get access to a specific repository, at least one of the (positive) votes should come from a core contributor having access to that repository. If there are multiple repositories involved, these votes can come from multiple core contributors having different access.

kdmccormick:
> Restricting the vote to the people who have access to the repo sends a signal to the other core contributors that their opinion is very secondary. 

@antoviaque I agree, and based on [this thread](https://github.com/openedx/open-edx-proposals/pull/293#discussion_r797956840) I think we're all aligned with you, even though though the current wording doesn't make that clear.

> BUT to get access to a specific repository, at least one of the (positive) votes should come from a core contributor having access to that repository. If there are multiple repositories involved, these votes can come from multiple core contributors having different access.

@antoviaque  I could see a situation where, after two weeks:
* a candidate gets five Yes votes and zero No votes, but
* doesn't have Yes votes from folks with access to their proposed repositories.

In that case, how would we proceed?

>  It may make sense to say the maintainer of the repo itself should give a final affirmative vote on access to that repo, although we haven't yet codified "maintainer".

@sarina I think this is a good idea, but it'd be hard to codify here without the Maintainer OEP landed. We could stick with some variant of what we're discussing on this PR (repo access), and then consider re-framing it around maintainership once all repos have maintainers.

sarina:
> @sarina I think this is a good idea, but it'd be hard to codify here without the Maintainer OEP landed. We could stick with some variant of what we're discussing on this PR (repo access), and then consider re-framing it around maintainership once all repos have maintainers.

Yes, sorry I didn't specify here that I was mostly lamenting that that hadn't landed. I like considering updating this once OEP-55 has landed.

felipemontoya (approved):
I like this clarification. The sponsor counting towards affirmative votes from other org was raised in my mind during the process as well.

antoviaque:
@kdmccormick 

>  I could see a situation where, after two weeks:
>
> *    a candidate gets five Yes votes and zero No votes, but
> *    they don't have a Yes vote from a CC with access to one or more of their proposed repositories.
>
> In that case, how would we proceed?

In that case imho they can't get the access, because they are missing one of the requirements -- a bit like someone who would get 4 "yes" votes, they would be missing one, and would need to get it before they can get the status/access? That would probably be something we could then resolve by pinging the core contributors from that repo, which could take the final decision there -- which is similar to what was just discussed about maintainers just above?

kdmccormick:
@antoviaque Would they still become core contributors?

sarina:
> People who get to vote: All CCs and all committers for any affected repos.

This reads to me as, if I nominate someone for `repo A`, only those who can commit to `repo A` can vote at all. I don't love this. Some repos have a very small number of committers, and chasing them down may be hard. Many repos only have committers from 2U. What is objectionable about:

People who get to vote: Any CC or committer can vote. Additionally, at least one vote must come from a person with commit access to the repo being nominated

kdmccormick:
I think @ormsbee 's suggestion was meant to be *more* liberal than the status quo in terms of who can vote. I read it as:
> People who get to vote: (All CCs) and (all committers for any affected repos).

where the "all committers for any affected repos" clause is only necessary until tCRIL and 2U are integrated into the CC program, after which point everyone with commit access will be a CC.

In contrast, @sarina 's suggestion (which seems equivalent to @antoviaque 's suggestion), is slightly more conservative than the status quo:

> People who get to vote: Any CC or committer can vote. Additionally, at least one vote must come from a person with commit access to the repo being nominated

Phrasing aside, I think these are both reasonable conceptually. Sarina & Xavier's suggestion requires active affirmation from someone with commit access; Dave's suggestion requires just passive affirmation from them (absence of a "No" vote).

As far as I can tell, nobody is suggesting that we restrict the general voting pool to anything smaller than "All CCs", so I think we need to just pick from between Dave's suggestion or Sarina/Xavier's suggestion.

ormsbee:
> This reads to me as, if I nominate someone for repo A, only those who can commit to repo A can vote at all. I don't love this. Some repos have a very small number of committers, and chasing them down may be hard. Many repos only have committers from 2U.

That was absolutely not the meaning I intended, so I'm glad you brought it up because I seem to perpetually muddle my communication on this no matter how hard I try to phrase it. When I said "All CCs and all committers for any affected repos", I meant that the union of those groups can vote, not the intersection:

* _Any core contributor can vote._ Period. **All of them.** The translation reviewers, the UX designers, the Project Managers, the Forum Moderators, Code Contributors for any repo–everyone who is or who should be listed on the [Core Contributors](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3156344833/Current+Core+Contributors+to+the+Open+edX+Project) page or any place we eventually decide to move authoritative list of Core Contributors to the Open edX Project–can vote.
* _In addition to the above_, any committer for any repository that the CC candidate would be granted commit/merge rights to can also vote, regardless of whether they are or are not listed as a Core Contributor.

_I specifically want to make it so that we're not blocked on chasing people down for permission to add CCs._ I want to widen the scope of people and rely on the ability to cast "no" votes to prevent unwanted outcomes, rather than defaulting to "no" by letting people stop nominations via silence.

> What is objectionable about:
> People who get to vote: Any CC or committer can vote. Additionally, at least one vote must come from a person with commit access to the repo being nominated
> 
> If I don't have commit access to repo A but I can commit to repo B, surely I may have reason to voice a yes vote: perhaps the nominee has committed to repo B and I can vouch for their conduct/quality. Perhaps I am capable of looking at the nominee's PRs on repo A and assessing the conduct/quality.

Disclaimer Preface: I think that would work, and if I you feel strongly about it, I'll adopt the language verbatim and merge this today.

My understanding of the delta between what we're proposing is:

1. Your proposal would require at least one "yes" vote from someone with commit access to the specific repo in question. Mine would not–people could be voted in without positive input from any existing committers to that repo. Both proposals allow any one existing committer to an existing repo to say "no" and veto the nomination, however.
2. Your proposal would allow non-CCs who have commit rights to unaffected repos (things the CC-candidate would not be given commit/merge access to) to vote. In my proposal, non-CCs could only vote if they were committers for one of the affected repos.

So our proposals are both a more more and a bit less permissive in who gets a say.

The reasons why I didn't want to require a "yes" vote from an existing committer to a repo:

* I felt that committers still had input since they have a veto (as everyone does), and that was sufficient to protect their interests. We shouldn't be afraid of saying no.
* I wanted more of a "default to yes" mindset and I didn't want things to get slowed down or stalled entirely by a team that feels too busy to respond, or to have to chase those people down.
* I thought the edge cases would get complicated. What happens when there are three repos? Or one of those repos is being maintained by one person who's not responding properly. Or one of those is an abandoned repo that has no official committers at all? There are common sense solutions for all of those, but I figured it was easier to just relax that rule entirely rather than codify what to do in these situations.

The reasons why I didn't want to let non-CC committers to other repos have a vote:

* Being a CC implies a long term commitment and involvement to the project as a whole, that being a committer to one unrelated repo does not necessarily entail.
* There should be a straightforward path to becoming a CC.
* I thought that it would simplify the process if there were disagreements on technical grounds. Say someone is nominated for four repos, but gets a "no" vote by a maintainer of only one of them due to architectural disagreements. I think it should be reasonable to move forward with the three other repos in that case.

------

But again, I think what you proposed will work, and I'm only putting this up here by way of explanation of my position. This can always be amended later if it's a problem. Please let me know your thoughts.

Thank you.

sarina:
OK @ormsbee you've convinced me by the "default to Yes" argument. I'm overindexing on the specific concerns of and disagreements with 2U from my past.

> Say someone is nominated for four repos, but gets a "no" vote by a maintainer of only one of them due to architectural disagreements. I think it should be reasonable to move forward with the three other repos in that case.

I do agree with this.

sarina (approved):

ormsbee:
@sarina: I made the changes, and tried to remove a few commas to make the meaning clearer. Please let me know if it reads correctly to you and I'll merge.

Thank you.

sarina:
This is good 2 go from my perspective! You'll need to chase @kdmccormick down to mark his requested changes as done.

kdmccormick (approved):
Could use some more unit test coverage, please revise.

Jk, looks great 💯

ormsbee:
Thanks folks!

PR #297, opened 2022-02-11
OEP-1 Codify the Change History section for all OEPs

sarina:
Change history as a part of the OEP was added to OEP-1 in https://github.com/openedx/open-edx-proposals/pull/19 but this seems to have been removed from OEP-1 at some point. However, a good proportion of OEPs do have this section, and broadly I find these sections extremely useful. I propose we add them back in as part of the OEP template, and we make them even **more** useful by also adding links to PRs so readers can quickly access the full discussion / context around why a change was made.

 See https://github.com/openedx/tcril-engineering/issues/132

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -489,9 +489,25 @@ OEPs may include auxiliary files such as diagrams. Such files must be added to
 an oep-XXXX/ directory, where "XXXX" is the OEP number. Include original diagrams
 alongside image files, to make it easy for others to update the OEP in the future.
 
+Change History
+--------------
+
+For every change (including the initial document creation), include an entry in
+a "Change History" section modeled off the one below. A Change History entry
+should include three parts: the date of the change, a very brief summary of
+changes made, and a link to the pull request where the discussion and approval
+took place. The changes should be ordered such that the most recent change is
+at the top of the list.
 
 Change History
 ==============
+
+2022-02-11
+----------
+
+* Codify the "Change History" section, which most OEPs already use
+* Specify that entries should link to the discussion PR.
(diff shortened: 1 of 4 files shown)


davidjoy (approved):
Seems pretty sensible to me; it's nice to have easy access to all that context and not have to rely on squinting at Github's commit history.

oeps/processes/oep-0001.rst
```diff
@@ -489,32 +489,51 @@ OEPs may include auxiliary files such as diagrams. Such files must be added to
...
 alongside image files, to make it easy for others to update the OEP in the future.
 
+Change History
+--------------
+
+For every change (including the initial document creation), include an entry in
+a "Change History" section modeled off the one below. A Change History entry
+should include three parts: the date of the change, a very brief summary of
+changes made, and a link to the pull request where the discussion and approval
+took place.
```
feanil, line 499:
```suggestion
took place.  The changes should be ordered such that the most recent change is at the top of the list.
```
Happy for it to go the other direction as well but right now, we don't prescribe this and it has resulted in some confusion.  And some change histories that have been partially out of order as some people add changes to the top while others add them to the bottom.

feanil (approved):
One suggestion to add some implementation detail but otherwise this makes sense to me.

oeps/processes/oep-0001.rst
nedbat: Definitely let's pick a direction! :)   Most recent at the top is good.

nedbat (approved):

felipemontoya (approved):

pdpinch (approved):
This is great, thanks.

PR #298, opened 2022-02-12
OEP-1 Add tl;dr section to OEP-1

sarina:
The What

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -126,8 +126,6 @@ Finally, the Arbiter is responsible for the decision to transfer an OEP if the
 original Authors have become unresponsive (as described in `Transferring OEP
 Ownership`_).
 
-.. _Architecture Group:
-
 Architecture Group
 ------------------
 
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -47,6 +47,20 @@ OEP-1: OEP Purpose and Guidelines
   :local:
   :depth: 1
 
+Overview
+========
+
+* An Open edX Proposal specifies a best practice, architectural decision, or
+  community process that the Open edX communty has agreed should be adopted by
+  the project.
+* OEPs are not used to dictate small decisions made in every day feature work.
+  See `OEP-19 <https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0019-bp-developer-documentation.html?highlight=documentation#agile-documentation-types>`_
+  for more detail.
+* This document specifies both how to write an OEP and how to get to a consensus
+  throughout the community that the OEP should be accepted.
+* OEPs are written by one or more `Authors`_ and guided through the process by
+  an `Arbiter`_.
+
 What is an OEP?
 ===============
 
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0001.rst
```diff
@@ -47,6 +47,20 @@ OEP-1: OEP Purpose and Guidelines
   :local:
   :depth: 1
 
+TL;DR
```
kdmccormick, line 50: What would you think about calling this an "Abstract", "Overview", or "Synopsis" instead?

"TL;DR" is kinda informal and also a bit more cryptic to newcomers or nonnative English speakers.

oeps/processes/oep-0001.rst
```diff
@@ -47,6 +47,20 @@ OEP-1: OEP Purpose and Guidelines
...
+=====
+
+* An Open edX Proposal specifies a best practice, architectural decision, or
+  community process that the Open edX communty has agreed should be adopted by
+  the project.
+* OEPs are written about broad architectural changes and community-wide
+  processes and practices. They are not used to dictate small decisions made in
+  every day feature work.
```
kdmccormick, line 58: Although I agree with this, the OEP doesn't actually say it, so I don't think it should be in the TL;DR.

kdmccormick (commented):
Couple comments, but looks great otherwise.

oeps/processes/oep-0001.rst
sarina: I was trying to summarize the ideas in the following sentences:

>  An OEP is a document that details a specific technology decision being made by the Open edX community, in the form of a best practice, architecture design, or process adjustment. An OEP should provide the use cases and rationales that surround that choice. 

> Small enhancements or patches often don’t need an OEP and can be injected into the Open edX development workflow with a patch submission.

I may have injected too much of the how-it-currently-works into that statement. Is there a better way of quickly indicating something like, "OEPs are big and there are easier ways to effect smaller changes"?

kdmccormick: > Small enhancements or patches often don’t need an OEP and can be injected into the Open edX development workflow with a patch submission.

I read this as a funny way of saying that "you don't need an OEP to accompany every single PR." Which, on second thought, does sort of capture what your TL;DR bullet is saying. Aside: it would be nice if the OEP also said "for medium-size, repo-specific decisions you can write an ADR instead".

Anyway, I think the bullet point is good to leave in, although to really lean into short-and-snappiness I recommend phrasing it as simply:
> * OEPs are *not* used to dictate small decisions made in every day feature work.

sarina: I like that @kdmccormick . What do you think about:

```
OEPs are not used to dictate small decisions made in every day feature work. See [OEP--19](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0019-bp-developer-documentation.html?highlight=documentation#agile-documentation-types) for more detail.
```

kdmccormick: 👍🏻

nedbat (approved):

kdmccormick (approved):

PR #302, opened 2022-02-13
[OEP-1] Switch author-ownership model to a community-stewardship one

sarina:
## problem

In the first sections of OEP-1, it is not stated that an Author 'owns' an OEP or what the definition of 'ownership' is. However, in the sections under OEP Maintenance, it is stated that the Authors do own the OEP, and if the Author is not available, it seems the Arbiter does. However, given our Community is currently largely-defined by employment-based membership, it can, will, and is the case that [some OEPs](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0016-bp-adopt-bootstrap.html) have Authors and Arbiters that are no longer community members.

## proposed changes

Since 'ownership' of an OEP is not defined, I propose that we migrate to a 'stewardship' model. For now,  I propose that the Architecture Group (defined in the OEP) are stewards of the OEP; as we build an Architecture Working Group, that stewardship should transfer to them.

A definition of an OEP's transition to being stewarded by this community group will be added, and the section around transferring ownership will be removed. Sections proscribing defacto 'ownership' roles of Arbiter in the two 'Updating" sections are not addressed here, as they are addressed in https://github.com/openedx/open-edx-proposals/pull/301

Additionally, a specification of when to add a new person to the Authors or Arbiter section is defined, as it is currently unclear.

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -7,12 +7,13 @@ OEP-1: OEP Purpose and Guidelines
 +---------------+--------------------------------------------------------------+
 | Title         | OEP Purpose and Guidelines                                   |
 +---------------+--------------------------------------------------------------+
-| Last-Modified | 2022-01-12                                                   |
+| Last-Modified | 2022-02-27                                                   |
 +---------------+--------------------------------------------------------------+
 | Authors       | Calen Pennington <cale@edx.org>,                             |
 |               | Joel Barciauskas <joel@edx.org>,                             |
-|               | Nimisha Asthagiri <nimisha@edx.org>                          |
-|               | Feanil Patel <feanil@tcril.org>                              |
+|               | Nimisha Asthagiri <nimisha@edx.org>,                         |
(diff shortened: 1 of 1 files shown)


feanil (approved):

nedbat:
(BTW, it's really helpful to put the OEP number in at the start of the PR title, esp when we have a flurry of them like for OEP-1).

nedbat:
My only concern here is that we are adding another reference to "the Architecture Group," even though that is a very vaguely described group of people.  The changes themselves are fine, but it would be great to clarify who those people are at some point.

nedbat (approved):

sarina:
Agree on both counts, I was doing a bunch of prefixing but fell flat. And yes, I don't love that the Arch Group is nebulous - I'm trying to make sure we only refer to one place in the doc, and we can get that one place updated quickly. @e0d has been talking about an Arch WG with @ormsbee - I'm thinking that may be the correct info to put in this OEP.

PR #304, opened 2022-02-15
 OEP-54 - Handle commitments fullfilment 

antoviaque:
Follow-up to the discussion in https://github.com/openedx/open-edx-proposals/pull/292/files#r800459077 about handling fulfillment of commitments by core contributors.

@sarina Here is a proposed change, to tie the commitments to the responsibilities of a core contributor, and make it explicit that it is expected to be upheld, to maintain the core contributor title. It however provides plenty of room for adapting to each individual situation -- the important imho is to notice those situations and discuss them.

### DIFF
--- a/oeps/processes/oep-0054-core-contributors.rst
+++ b/oeps/processes/oep-0054-core-contributors.rst
@@ -312,7 +312,10 @@ Trigger events for removing a CC would be:
 
 * A CC informs the Program Administrators that they can no longer continue in
   the role
-* A CC has been inactive and unreachable for more than 6 months
+* A CC has not been able to fulfill the commitments made in the
+  `Revised Declaration of Commitment to the Core Contributor Program`_,
+  or has been unreachable, for more than 6 months. See
+  `fulfilling CC commitments`_
 * A CC hasn't met project quality standards consistently despite feedback
 * A CC has consistently shown poor judgement, such as merging PRs without
   due consideration or addressing outstanding concerns
@@ -321,10 +324,17 @@ Trigger events for removing a CC would be:
 For those wishing to indicate they are leaving the CC program, or to report a
 Code of Conduct violation, please `Contact the Program Administrators`_.
 
-Program administrators will be responsible for determining when a CC is
-inactive/unreachable, by doing a check at least once per quarter. Since the
-reasoning for removing inactive CCs is due to security concerns, those who've
-been removed for inactivity may get in touch with Program Administrators to
+.. _fulfilling CC commitments:
+
+Core contributors are responsible for reporting when they aren't able to
+fulfill their commitments. Program administrators
+will do regular checks, at least once per quarter, to identify CCs who are 
+unreachable or fall short of their commitments. In such cases, the program
+administrator will reach out to the CC to discuss the best corrective actions 
+to take.
+
+In case there is no reply, the inactive CC would be removed for security
+reasons. Removed CCs may get in touch with Program Administrators to 
 fast-track getting their access back.
 
 Contact the Program Administrators


edx-webhook:
Thanks for the pull request, @antoviaque! I've created [OSPR-6459](https://openedx.atlassian.net/browse/OSPR-6459) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -296,7 +296,9 @@ Trigger events for removing a CC would be:
 
 * A CC informs the Program Administrators that they can no longer continue in
   the role
-* A CC has been inactive and unreachable for more than 6 months
+* A CC has not been able to fulfill the commitments made in the
+  `Revised Declaration of Commitment to the Core Contributor Program`_,
+  or has been unreachable, for more than 6 months
```
feanil, line 301: Is it possible in the situation where a CC can't meet their commitments to update the level of commitment they agree to in their declaration?  Is that already documented somewhere and can be linked here? Or can we add something to that effect?  I see that the level of commitment varies across CCs and CC orgs so perhaps if they are not meeting their commitment, updating their level of commitment to better match the work they are actually doing is a good alternative to fully removing them from the program?

This is not a blocker in my mind but perhaps an alternative worth clarifying?  I think given that the result is a discussion with the program administrator, we already have the flexibility to enact the above option but perhaps it's worth mentioning explicitly?

feanil (approved):

oeps/processes/oep-0054-core-contributors.rst
sarina: I also think we should be a little careful with this, because hypothetically a community member (not joining on behalf of an organization) could get involved, and they wouldn't be necessarily a signatory to the Declaration.

Could we incorporate something like 

```
CCs may `Contact the Program Administrators`_ if they are struggling
with their ability to fulfill their commitment.
```
?

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -305,10 +307,15 @@ Trigger events for removing a CC would be:
...
 
-Program administrators will be responsible for determining when a CC is
-inactive/unreachable, by doing a check at least once per quarter. Since the
-reasoning for removing inactive CCs is due to security concerns, those who've
-been removed for inactivity may get in touch with Program Administrators to
+Core contributors are responsible for reporting when they aren't able to be
+sufficiently available to fulfill their commitments. Program administrators
+will do regular checks, at least once per quarter, to identify CCs who are 
+unreachable or fall short of their commitments. In such cases, the program
+administrator will reach out to the CC discuss the best corrective actions
```
sarina, line 314: nit: `reach out to the CC to discuss`

oeps/processes/oep-0054-core-contributors.rst
sarina: hm, ok I see you do this in the next section.

@feanil I think what you're asking is good, but nuanced and maybe too much/too complicated to put into the OEP. Suggesting that the CC get in touch with the Program Administrators when things are going south seems like the best first step to take.

oeps/processes/oep-0054-core-contributors.rst
```diff
@@ -305,10 +307,15 @@ Trigger events for removing a CC would be:
 For those wishing to indicate they are leaving the CC program, or to report a
 Code of Conduct violation, please `Contact the Program Administrators`_.
 
-Program administrators will be responsible for determining when a CC is
-inactive/unreachable, by doing a check at least once per quarter. Since the
-reasoning for removing inactive CCs is due to security concerns, those who've
-been removed for inactivity may get in touch with Program Administrators to
+Core contributors are responsible for reporting when they aren't able to be
+sufficiently available to fulfill their commitments. Program administrators
```
nedbat, line 311: It's enough to say, "Core contributors are responsible for reporting when they aren't able to fulfill their commitments."  One of the commitments is being available.

nedbat (approved):

oeps/processes/oep-0054-core-contributors.rst
antoviaque: @feanil @sarina 

> Is it possible in the situation where a CC can't meet their commitments to update the level of commitment they agree to in their declaration?

Yes this kind of measures would definitely be good to consider when this happens -- the intent is what you describe @sarina , to start by discussing, and letting the administrator arbitrate it individually. If some patterns emerge, we can always discuss them then, but this way we get to go through a few cases before?

antoviaque: @sarina Yup - the important bit is to establish the principle, and have it clearly acknowledged by everyone. But we keep the latitude to adapt how it is implemented to individual cases. Revisiting once we face actual problems and can have a discussion is actually the goal, so we can be practical when figuring out solutions.

sarina: Great, got it. I can see not wanting to caveat what to do if you're finding yourself falling behind in this section; you mention it later and that's a better place. However, I wonder if it's worth linking to like:

```
A CC has not been able to fulfill the commitments made in the
  `Revised Declaration of Commitment to the Core Contributor Program`_,
  or has been unreachable, for more than 6 months. See `fulfilling CC commitments`_
```

with a `.. fulfilling CC commitments` line at 309.

I'm not wedded to that wording, pick what you would like if you take this suggestion.

antoviaque: @sarina Sounds good! Done.

antoviaque:
I think I have resolved all the comments - let me know if not. If there is no new comment or objection by the end of the week, I'll merge this.

sarina (approved):

edx-webhook:
@antoviaque 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/open-edx-proposals&entry.867055334=https://github.com/openedx/open-edx-proposals/pull/304&entry.1484655318=https://github.com/antoviaque&entry.752974735=2022-02-15+15:05&entry.1917517419=2022-02-28+07:30&entry.2133058324=Yes) so we can improve your experience in the future.

PR #306, opened 2022-02-18
OEP-22 Add Clarifying Note

sarina:
Add a note to clarify that we really want you to use the templates, but what keyword to use in case you don't wanna.

### DIFF
--- a/oeps/processes/oep-0021-proc-deprecation.rst
+++ b/oeps/processes/oep-0021-proc-deprecation.rst
@@ -211,7 +211,9 @@ Do the following to document your proposal:
       While it is possible to create Issues with no template, it is strongly
       encouraged that you go to `github.com/openedx/:repo/issues`, click "New
       Issue", and choose the DEPR template so you don't miss any fields and
-      automation works properly. If you must create an Issue outside the
+      automation works properly. The template fields help us more quickly
+      address deprecation issues and reduce the amount of back and forth needed
+      to make progress on work. If you must create an Issue outside the
       template, please preface your issue title with ``[DEPR]``.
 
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0021-proc-deprecation.rst
```diff
@@ -207,6 +207,13 @@ Do the following to document your proposal:
    deprecation spans multiple repos, choose the primary/most relevant repo, or
    use the `public-engineering`_ project.
 
+   .. note::
+      While it is possible to create Issues with no template, it is strongly
+      encouraged that you go to `github.com/openedx/:repo/issues`, click "New
+      Issue", and choose the DEPR template so you don't miss any fields and
+      automation works properly. If you must create an Issue outside the
```
feanil, line 214:
```suggestion
      automation works properly. The template fields help us more quickly
      address deprecation issues and reduce the amount of back and forth needed
      to make progress on work. If you must create an Issue outside the
```

Maybe being more clear about the exact benefits the template buys will make it clear why it's valuable?

ormsbee (approved):

sarina:
I like it, thanks @feanil

sarina:
Whoops. I accidentally merged this when bulk merging others, and committed a "temp" commit. What is the best way to proceed @feanil ?

feanil:
Given that it's landed on master already, I think it's fine to just leave it.  There is no easy way to fix master without causing trouble for people who have already pulled it.  In particular, I'm not sure how the ReadTheDocs integration will handle it.

sarina:
thanks and sorry fam

PR #290, opened 2022-01-28
OEP-0055, Project Maintainers

e0d:
~This is a work-in-progress draft of a formalization of the maintainer role for the Open edX project.  While it is a draft, comments are welcome.  However, please focus on the content at this time, not syntax or document style issues.~ Update: This has entered the formal review period, which concludes on 2022-04-11.

PR #314, opened 2022-02-25
fix: Updates to OEP-55 (Maintainers)

sarina:
Hi @e0d - in the interest of moving this OEP along, I've resolved all the open comments (except one thread by @hurtstotouchfire and @nedbat surrounding adding new maintainers. Let's continue that [thread here](https://github.com/openedx/open-edx-proposals/pull/290/files#r796956092) and review this PR as responses to the remaining inline comments.

There was an error with the filename so I did rename it to match how files are conventionally named.

### DIFF
--- a/oeps/processes/oep-0055-proc-project-maintainers.rst
+++ b/oeps/processes/oep-0055-proc-project-maintainers.rst
@@ -1,6 +1,6 @@
-=============================
+===========================
 OEP-55: Project Maintainers
-=============================
+===========================
 
 .. This OEP template is based on Python's PEP standard.
 
@@ -9,7 +9,6 @@ OEP-55: Project Maintainers
 
    * - OEP
      - :doc:`OEP-0055 <oep-0055-proc-project-maintainers>`
-
    * - Title
      - Project Maintainers
    * - Last Modified
--- a/oeps/processes/oep-0055-project-maintainers.rst
+++ b/oeps/processes/oep-0055-project-maintainers.rst
@@ -8,13 +8,7 @@ OEP-0055: Project Maintainers
    :widths: 25 75
 
    * - OEP
-     - Link to the doc in the following format::
-
-        :doc:`OEP-0055 <oep-0055-proc-Project Maintainers>`
-
-        * <XXXX is the next available OEP number>
-        * <YYYY is the abbreviated Type: proc | bp | arch>
-        * <ZZZZ is a brief (< 5 words) version of the title>
+     - :doc:`OEP-0055 <oep-0055-proc-Project Maintainers>`
 
    * - Title
      - Project Maintainers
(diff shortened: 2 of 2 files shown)


sarina:
cc @feanil for review as well.

oeps/processes/oep-0055-proc-project-maintainers.rst
```diff
@@ -168,10 +166,17 @@ Technical responsibilities of maintainers include:
...
 
 It is acceptable for a component to be maintained by a team or group
-as long as that arrangement does not reduce the reachablility of the
-maintainer or reduce accountability.
+as long as that arrangement does not reduce the reachability of at least one
+maintainer or reduce accountability for the above responsibilites. The
+maintaining team must have ways of direct contact, such as an email address and
+a GitHub team, that anyone can use to get ahold of someone - it is completely
+acceptable for the team to figure out arrangements for list monitoring, such as
+an "on-call" rotation.
```
hurtstotouchfire, line 179: This looks good to me 👍

oeps/processes/oep-0055-proc-project-maintainers.rst
```diff
@@ -183,19 +188,20 @@ In order to be a repository maintainer, you must
...
-  demonstrated excellence in the role of core contributor.  From
-  time-to-time, in the case of a strong candidate and project need,
-  core contributor status and maintainer status may be conferred at
-  the same time.
+* Be a member of the `Core Contributor Program`_ with rights to the repository
+  in question who has demonstrated excellence in the role of core contributor.
+  From time-to-time, in the case of a strong candidate and project need, core
+  contributor status and maintainer status may be conferred at the same time.
+* In the case of repos being moved into the `openedx GitHub organization`_, be
+  an original maintainer of the code.
```
hurtstotouchfire, line 196: Some worries about how this sentence will age once the move is well behind us.

oeps/processes/oep-0055-proc-project-maintainers.rst
```diff
@@ -221,18 +227,23 @@ Coordination will typically be over asynchronous channels.
...
-**This section should be congruent with the Core Contributor process**
-
 A new maintainer can be added in any of the following ways:
 
+..
+  Should this follow CC procedure (nomination, public discussion?)
+  Is there reason for this to be private/within the scrum of scrums?
+  Is there any reason to call upon the `program administrators`_,
+  maybe to be a backstop or confer access priviledges if the original
+  maintainers are all gone?
```
hurtstotouchfire, line 245: It's very difficult to think this through without some concrete examples of repos which no longer have a maintainer. I even have some trouble at this stage imagining when a new maintainer would be added. I'm assuming in cases where a maintainer team is in effect that adding people to their maintainer team would be an internal process, and not something we need to consider here. Is there a case for adding new maintainers *except* when the old one is gone? Possibly a handoff type situation.

All of this would be easier to imagine with a draft list of maintainers so maybe I just need to catch up on whether that exists yet.

sarina:
@hurtstotouchfire I don't think @e0d is planning to use this PR, so I'm closing it. Will alert him to your comments.

oeps/processes/oep-0055-proc-project-maintainers.rst
sarina: @hurtstotouchfire OEPs are not immutable. I really encourage authors and reviewers to consider mostly the happy path, along with the very most likely scenarios. Trying to "future proof" the document to consider all cases makes the document extremely hard to approach. We can always add to the OEP if some corner cases start popping up repeatedly.

PR #317, opened 2022-02-28
refactor: OEP-56: use shorter-form links to keep the text more readable.

nedbat:
Tweaks to the .rst syntax to keep long URLs from interrupting the text.

### DIFF
--- a/oeps/processes/oep-0056-architectural-advisory-process.rst
+++ b/oeps/processes/oep-0056-architectural-advisory-process.rst
@@ -1,6 +1,7 @@
 ======================================
 OEP-56: Architectural Advisory Process
 ======================================
+
 .. list-table::
    :widths: 25 75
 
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0056-architectural-advisory-process.rst
```diff
@@ -142,6 +143,8 @@ We see three significant problems with the idea of having such a forum:
 
 - Our community is comprised of many independent organizations.  There are times when our component organizations may want to use the Advisory Process on "internal" decisions as well.  Do we create a separate forum meeting for those decisions?  How do we then ensure that we err on the side of having an open, public forum?  Again, logistically, this is quite difficult.
 
+.. _principles:
+
 Appendix A: Advisory Process Principles
```
nedbat, line 148: Though TBH, this seems like the core of the proposal, and so shouldn't be an appendix.

davidjoy: I don't feel strongly - decided to pull it out to shorten the core proposal itself after some early feedback.

davidjoy (approved):
Thanks for this! :D  You've taught me some more RST.

PR #324, opened 2022-03-30
OEP-52: docs: add OEP for event bus architecture (ends June 30)

robrap:
OEP that makes the case for adding an event bus, as well as a variety of other decisions around implementing an event bus.

**Notes:** 

1. This PR replaces https://github.com/openedx/open-edx-proposals/pull/233. The original OEP primarily discussed the implementation technology, which was moved to a separate ADR. See the OEP's 
"Change History" for more details of what has changed.
2. The goal is to merge this OEP with a Provisional status, and continue our implementation and learning before moving this to Accepted.

PR #338, opened 2022-06-09
squash! add arbiter and put under review

robrap:
This PR opens review on an existing OEP PR: https://github.com/openedx/open-edx-proposals/pull/324


blarghmatey:
That review period seems reasonable, so long as I don't need to be too involved at the end of it, since I'll be on vacation for the week of the 20th.

Feel free to include my email (tmacey@mit.edu)

PR #340, opened 2022-06-10
docs: Add ADR template to oep-0019

oliviaruizknott:
This is where I expected to find an ADR template, and there wasn’t one.
Now I’ve added one! This template is copied from a suggestion by Ben
Holt, with my addition of an "Options Considered" section.

PR #312, opened 2022-02-24
OEP-57: Core Repositories

kdmccormick:
This OEP is a big ol' work-in-progress. Feel free to leave early feedback, but please keep in mind that it's still a rough outline. Lots of this stuff is my stream-of-consciousness. It's not yet representative of tCRIL's or anyone else's plans/opinions. 

Fixes #295

PR #343, opened 2022-06-16
docs: Add an ADR for using Backstage to support OEP-55

feanil:
Since this is the first ADR under oep-55 I also added the relevant
directory structure and wired it up into the OEP rst.

### DIFF
--- a/oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
+++ b/oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
@@ -1,0 +1,139 @@
+0001 Use Backstage to Support Maintainers
+#########################################
+
+Status
+******
+
+**Provisional**
+
+Context
+*******
+
+* :doc:`OEP-55 <../../oep-0055-proc-project-maintainers>` laid out that we should document the maintainers of a component.
+
+* As a maintainer it is hard to know what components you are maintaining.
+
+* As a community member it is currently difficult to quickly look up who the
+  maintainers are for multiple components quickly.
+
+* Given the number of repos that make up the Open edX platform and the
+  complexity of the job of maintaining them, we need a system for cataloging,
+  and making transparent the critical details of our repositories.
+
+Decision
+********
+
+To solve the above issues and to enable the potential to ease other
+discoverability and maintainership burdens, the Open edX community shall
+maintain a `backstage`_ instance and related metadata.
+
+Consequences
+************
+
+* To support the backstage instance ownership data will have to be stored in a
+  structured way.  We will follow Backstage best practices and store this data
+  in a ``catalog-info.yaml`` file in each component repository.
+
+* Once a backstage instance exists we have the opportunity to further leverage
+  it as a centralized access point for maintainers as well as community members
+  to quickly and easily assess different components and find relevant links to
+  code, documentation, owners, etc.
+
+* Backstage will also give us the ability to expose other health metrics to
+  maintainers so they can more easily assess the components they own.
+
+* As a result of adding the ``catalog-info.yaml`` file we end with two top-level
+  files that have very similar data.  If this decision is not reversed after the
+  initial pilot, it may make sense to revisit this consequence and see if we can
+  consolidate this metadata into a single file. `tcril-engineering#331
+  <https://github.com/openedx/tcril-engineering/issues/331>`_ has been created
+  to follow up on this work.
+
+
+Rejected Alternatives
+*********************
+
+Spreadsheet
+===========
+
+Pros
+----
+
+* Easy to build and get a high level view of the system.
+
+Cons
+----
+
+* Not a beautiful UI for maintainers to work with.
+
+* Data overload as we put more and more data into it with little to no control
+  over the presentation of the data.
+
+Owners in the Readme
+====================
+
+Pros
+----
+
+* Easy to see while in the repo.
+
+Cons
+----
+
+* Harder to aggregate the data from the unstructured files in a way that would
+  allow us to easily build a `single pane of glass`_ where maintainers can see
+  all the things they own.
+
+References
+**********
+
+* `Backstage`_ project site.
+
+* `catalog-info.yaml`_ spec definition
+
+* `Backstage system model`_ has detail about the different kinds of components
+  and how they connect to each other.
+
+* Sample ``catalog-info.yaml`` file:
+
+  .. code:: yaml
+
+    apiVersion: backstage.io/v1alpha1
+    # (Required) Acceptable Values: Component, Resource, System
+    kind: Component
+    metadata:
+      name: 'docs.openedx.org'
+      description: "This repository contains the root documentation site for https://docs.openedx.org"
+      links:
+        - url: "https://docs.openedx.org"
+          title: "Deployed Site"
+          icon: "Web"
+      annotations:
+        # Annotation keys and values can be whatever you want.
+        # We use it in open edx to have a comma separated list of github user
+        # names that might be interested in changes to the architecture of this
+        # component.
+        openedx.org/arch-interest-groups: "feanil"
+    spec:
+
+      # (Required) This can be a group(`group:<group_name>` or a user(`user:<github_username>`)
+      owner: group:docs.edx.org-maintainers
+
+      # (Required) Acceptable Type Values: service, website, library
+      type: 'website'
+
+      # (Required) Acceptable Lifecycle Values: experimental, production, deprecated
+      lifecycle: 'production'
+
+      # (Optional) The value can be the name of any known component.
+      subcomponentOf: '<name_of_a_component>'
+
+      # (Optional) An array of different components or resources.
+      dependsOn:
+      - '<component_or_resource>'
+      - '<another_component_or_resource>'
+
+.. _Backstage: https://backstage.io
+.. _Backstage system model: https://backstage.io/docs/features/software-catalog/system-model
+.. _catalog-info.yaml: https://backstage.io/docs/features/software-catalog/descriptor-format
+.. _single pane of glass: https://www.webopedia.com/definitions/single-pane-of-glass/
--- a/oeps/processes/oep-0055-proc-project-maintainers.rst
+++ b/oeps/processes/oep-0055-proc-project-maintainers.rst
@@ -359,10 +359,12 @@ community to find a new maintainer.
 Documenting Maintainers
 -----------------------
 
-As it is essential that maintainers are discoverable and reachable,
-the person or persons responsible for maintaining a component should
-be listed in the repository README file along with the mechanism for
-contacting a maintainer.
+As it is essential that maintainers are discoverable and reachable, the person
+or persons responsible for maintaining a component should be listed in the
+repository.  Per :doc:`the backstage ADR
+<oep-0055/decisions/0001-use-backstage-to-support-maintainers>` we will use a
+``catalog-info.yaml`` file to capture the github handles for the maintainers of
+a repository.
 
 Jobs of the Maintainer
 ----------------------
@@ -392,7 +394,7 @@ categories
 * **Technical Vision**: Maintainers should have a documented technical
   vision for their components. This vision should minimally answer the
   following questions:
-  
+
   * What is the component for?
   * How it fits into the overall platform?
   * What are its limitations and key areas of debt?
@@ -444,7 +446,15 @@ Questions about the Program can be directed to the tCRIL administrator(s) via
 the ``#ask-tcril`` Slack room in the `Open edX Slack
 <https://openedx.slack.com/>`_. Questions or concerns of a more sensitive nature
 may be directed to ``maintainer-program-admins@tcril.org``.
-  
+
+Related Decisions
+=================
+
+.. toctree::
+   :glob:
+   :maxdepth: 1
+
+   oep-0055/decisions/*
 
 Change History
 ==============
@@ -459,7 +469,7 @@ Source Material
 * `Collective Code Construction Contract <https://rfc.zeromq.org/spec/42/>`_
 * `Best Practices for Maintainers <https://opensource.guide/best-practices/>`_
 * `Producing OSS <https://producingoss.com/en/producingoss-letter.pdf>`_
-  
+
 Notes
 =====
 


oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,111 @@
...
+		* Relevant link https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0055-proc-project-maintainers.html#documenting-maintainers
+
+Context
+*******
+
+* :doc:`OEP-55 <../../oep-0055-proc-project-maintainers>` laid out that we should document the maintainers of a component.
+
+* As a maintainer it is hard to know what components you are maintaining.
+
+* As a community member it is currently difficult to quickly look up who the
```
e0d, line 21: I would add here something like:

Given the number of repos that make up the Open edX platform and the complexity of the job of maintaining them, we need a system for cataloging, and making transparent the critical details of our repositories.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,111 @@
...
+  maintainers are for multiple components quickly.
+
+
+Decision
+********
+
+To solve the above issues and to enable the potential to ease other
+discoverability and maintainership burdens, the Open edX community shall
+maintain a backstage instance and related metadata.
```
e0d, line 30: First mention of backstage, link out using the link specified below?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,111 @@
...
+  structured way.  We will follow Backstage best practices and store this data
+  in a ``catalog-info.yaml`` file in each component repository.
+
+* Once a backstage instance exists we have the opportunity to further leverage
+  it as a centralized access point for maintainers as well as community members
+  to quickly and easily assess different components and find relevant links to
+  code, documentation, owners, etc.
+
+* Backstage will also give us the ability to expose other health metrics to
+  maintainers so they can more easily asses the components they own.
```
e0d, line 45: spelling ;)

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,111 @@
...
+      links:
+        - url: "https://github.com/openedx/docs.openedx.org"
+          title: "Source Code"
+          icon: "Code"
+      annotations:
+        openedx.org/arch-interest-groups: "feanil"
+    spec:
+      owner: group:docs.edx.org-maintainers
+      type: 'website'
+      lifecycle: 'production'
```
e0d, line 106: Do we want to list out acceptable values here for type and life-cycle?

e0d (changes requested):

e0d (approved):
Looks good.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+    apiVersion: backstage.io/v1alpha1
+    # (Required) Acceptable Values: Component, Resource, System
+    kind: Component
+    metadata:
+      name: 'docs.openedx.org'
+      description: "This repository contains the root documentation site for https://docs.openedx.org"
+      links:
+        - url: "https://github.com/openedx/docs.openedx.org"
+          title: "Source Code"
+          icon: "Code"
```
kdmccormick, line 104: The name, description, and source code URL as strike me as redundant with the repository's name, GitHub description, and GitHub URL. Is there anything we can do to de-dupe this or lint to keep it in sync?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+To solve the above issues and to enable the potential to ease other
+discoverability and maintainership burdens, the Open edX community shall
+maintain a `backstage`_ instance and related metadata.
+
+Consequences
+************
+
+* To support the backstage instance ownership data will have to be stored in a
+  structured way.  We will follow Backstage best practices and store this data
+  in a ``catalog-info.yaml`` file in each component repository.
```
kdmccormick, line 40: While I think Backstage is the right tool for our needs, I feel that we are reimplementing OEP-2, which defines `openedx.yaml` as a way to store repository metadata.

How do you feel about adding a provision this ADR along the lines of: *after X months of trying this decision out, we will either (a) roll it back, or (b) migrate `openedx.yaml` into `catalog-info.yml` and revoke/update OEP-2* ?

kdmccormick (commented):
A couple comments on the specifics, but the overall proposal sounds great to me and I'm excited to try out Backstage 🎉

robrap (commented):
Looking forward to the new tools.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
robrap: If the intention is to land this as "Provisional", it inspired me to write this update to OEP-1: https://github.com/openedx/open-edx-proposals/pull/344. Note: the same can be added for ADRs, but I want Livvy's ADR PR to land first: https://github.com/openedx/open-edx-proposals/pull/340.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+#########################################
+
+Status
+******
+
+**Draft**
+
+.. TODO Upon Acceptance
+
+	* Update OEP to mention the catalog-info.ymal file instead of the readme.
```
hurtstotouchfire, line 11: did you mean yaml?
```suggestion
	* Update OEP to mention the catalog-info.yaml file instead of the readme.
```

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+    kind: Component
+    metadata:
+      name: 'docs.openedx.org'
+      description: "This repository contains the root documentation site for https://docs.openedx.org"
+      links:
+        - url: "https://github.com/openedx/docs.openedx.org"
+          title: "Source Code"
+          icon: "Code"
+      annotations:
+        openedx.org/arch-interest-groups: "feanil"
```
nedbat, line 106: Where do we find the explanation of what "annotations" means?  Same for "spec", etc.  Is this defined by Backstage?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+    spec:
+
+      # (Required) This can be a group(`group:<group_name>` or a user(`user:<github_username>`)
+      owner: group:docs.edx.org-maintainers
+
+      # (Required) Acceptable Type Values: service, website, library
+      type: 'website'
```
nedbat, line 113: How does "type" relate to "kind"?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+
+      # (Required) Acceptable Lifecycle Values: experimental, production, deprecated
+      lifecycle: 'production'
+
+      # (Optional) The value can be the name of any known component.
+      subcomponentOf: '<name_of_a_component>'
+
+      # (Optional) The name of a different component or resource.
+      dependsOn: '<component_or_resource>'
```
nedbat, line 122: Wouldn't this usually be a list? How specific is it meant to be?  And to @kdmccormick's point above: isn't this redundant with other dependency information elsewhere (for example, requirements.txt)?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: I think right now the `openedx.yaml` file is only really used by the release process.  I think any promise here would mean slating in work to update the release scripts.  It almost feels like a new ADR rather than an addition to this one.  I'm happy to mention the consequence that we'll now have two metadata files with similar data but deciding to get rid of one, seems like another decisions with its own impact.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,126 @@
...
+  and making transparent the critical details of our repositories.
+
+Decision
+********
+
+To solve the above issues and to enable the potential to ease other
+discoverability and maintainership burdens, the Open edX community shall
+maintain a `backstage`_ instance and related metadata.
+
+Consequences
```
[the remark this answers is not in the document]
feanil: >     2U/edX were maintaining much of this information in a spreadsheet before the tCRIL-split. This ADR could acknowledge the need to update this existing tooling. Even if not mentioned in the ADR, this work needs to be communicated and planned for.

That feels like a consequence internal to 2U and not one for the open source project.  I agree the work needs to be done either way, but I'm not sure mentioning it here is valuable from the open source project perspective.

>     The existing spreadsheet also has exceptional maintenance requirements for the edx-platform monolith. Does tCRIL plan on supporting that level of detail, or is it planning on having edx-platform maintained by 2U/edX, with the split ownership being decided/documented/maintained by 2U/edX as we see fit?

I think edx-platform maintainership is being avoided for the pilot but we will eventually need to make a plan for it.  Given its complexity, I expect there will be a future ADR that addresses its ownership specifically.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: Yep, definitely did. This will resolve itself because this comment is only here as a reminder to me to update the OEP and will be removed once we've aligned on the rest of the content of this file.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: You can find it [here](https://backstage.io/docs/features/software-catalog/descriptor-format), I linked to it from just above this code-block in the references section.  Let me know if you think I need more specific stuff within the code.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: Ah, good catch, this is indeed a list, I'll correct the example.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: I added a consequence related to this, let me know what you think.

kdmccormick: > I think right now the openedx.yaml file is only really used by the release process.

In addition to the release metadata, OEP-2 also recommends ["tags" for repos](https://github.com/openedx/open-edx-proposals/blob/master/oeps/best-practices/oep-0002-bp-repo-metadata.rst#tags), which is redundant with the catalog-info spec. Like you say, though, I don't think they're used for anything.

> It almost feels like a new ADR rather than an addition to this one. I'm happy to mention the consequence that we'll now have two metadata files with similar data but deciding to get rid of one, seems like another decisions with its own impact.

I think I am seeing it differently. This feels similar when a legacy frontend is rewritten as an MFE: the decision to write the new MFE and to deprecate the old one ought to go hand-and-hand (unless there's some special reason to the legacy frontend around). Ensuring deprecation of the old thing is the responsibility of the team that adds the new thing.

>  As a result of adding the ``catalog-info.yaml`` file we end with two top-level
  files that have very similar data.  If this decision is not reversed after the
  initial pilot, it may make sense to revisit this consequence and see if we can
  consolidate this metadata into a single file.

This is a good start, but I wish it were more committal. Would you be willing to make a ticket on our backlog to investigate consolidating these two into a single file, and link it here?

feanil: Yea, I'd be happy to add the ticket, you want it linked in the text of the ADR or just onto this PR?

I see what you're saying about the perspective of it being like the MFEs. In my mind it is not like that because the only real purpose of the openedx.yaml is to indicate what repos are in the official release.  Anything else in there is not actually used in any substantive way.  This new config file has info for maintainership which while related is a different problem to solve so I was seeing them as 2 separate things.

That said, consolidating them should definitely be a long term goal, and I'm happy to put that in writing as a ticket to re-visit this and make a decision in a few months.

kdmccormick: Great, thanks Feanil.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
nedbat: Sorry, I see the backstage docs now, but I have a feeling that the pure generality of their docs could be narrowed with some specifics for our use of it.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,133 @@
...
+----
+
+* Not a beautiful UI for maintainers to work with.
+
+* Data overload as we put more and more data into it with little to no control
+  over the presentation of the data.
+
+Owners in the Readme
+====================
+
```
ghassanmas, line 76: I don't think both should be necessarily exclusive of each other, and I think we can get the best of both world. 

As I understand the `catalog-info.yml` should have the source of truth about maintners. Thus we could make an org wide action that whenever `catalog-info.yml` is changed/updated the repo's readme would be updated/overwritten accordingly. 

There are various ways to achieve this, here is for example a very simliar purpose action: https://github.com/JasonEtco/rss-to-readme. 

The only caveate that would make this thing tricky is that some repo has an rst while other has md format... But anyway this can be overcomed by defining clear steps/requirmets for both cases about how to _make a project repo compitable with the action_.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,133 @@
...
+  to quickly and easily assess different components and find relevant links to
+  code, documentation, owners, etc.
+
+* Backstage will also give us the ability to expose other health metrics to
+  maintainers so they can more easily assess the components they own.
+
+* As a result of adding the ``catalog-info.yaml`` file we end with two top-level
+  files that have very similar data.  If this decision is not reversed after the
+  initial pilot, it may make sense to revisit this consequence and see if we can
+  consolidate this metadata into a single file.
```
pomegranited, line 53: We'd also need to update [](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2810249408/Administering+a+CC+Committer+Onboarding+Runbook) process to ensure new CCs are added to the relevant `catalog-info.yaml` files. Or better yet, make that part of the [CC's onboarding](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3358458250/Core+Contributor+Committer+Onboarding) via an OSPR.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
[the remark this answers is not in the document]
hurtstotouchfire: @robrap maybe we can touch base over slack about the nature of the maintenance requirements? I'm curious how much of it would be relevant to any organization maintaining `edx-platform` vs 2U specifically.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,133 @@
...
+----
+
+* Easy to see while in the repo.
+
+Cons
+----
+
+* Harder to aggregate the data from the unstructured files in a way that would
+  allow us to easily build a `single pane of glass`_ where maintainers can see
+  all the things they own.
```
hurtstotouchfire, line 87: For some reason I have in my memory something about using github repo tags for this? I think it was @ormsbee's idea. In any case, would we want to integrate that with Backstage somehow or are they just separate concerns?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
hurtstotouchfire: maybe it could be baked into logic for the templates we start from when making new files?

See also, placeholders in yaml: https://stackoverflow.com/questions/41620674/use-placeholders-in-yaml

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: @pomegranited would this not be covered by the process to elect and assign maintainers generally?

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: I'll add another comment here, let's see if that will help.  Happy to add a lot more text here to help people understand what's up.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
ghassanmas: > We're still hashing out the layout of the README files so I think this is an open question 

Sounds reasonable, if I would like to further express my intention, where should I go? 

> A thing I've thought of before is to use something like cog to auto generate certain parts of the file from the sources of truth 

cog seems intresting!, its not exactly how I imagined it to be, but I see how it can be useful. wonder what @nedbat thinks about using it  😄

feanil: @ghassanmas we'll have another ADR for the format/layout of the README shortly so keep in this repo for that.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
feanil: More details on the system model are here: https://backstage.io/docs/features/software-catalog/system-model

I'll add it as a reference link.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -0,0 +1,133 @@
...
+  maintainers so they can more easily assess the components they own.
+
+* As a result of adding the ``catalog-info.yaml`` file we end with two top-level
+  files that have very similar data.  If this decision is not reversed after the
+  initial pilot, it may make sense to revisit this consequence and see if we can
+  consolidate this metadata into a single file.
+
+Rejected Alternatives
+*********************
+
```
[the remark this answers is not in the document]
feanil: @hurtstotouchfire do you mean were there no alternatives services or out of box tools?  If so, not that I am aware of.

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
kdmccormick: Your memory serves you right @hurtstotouchfire -- on the OEP-57 draft, [Dave mentioned using GH topics](https://github.com/openedx/open-edx-proposals/pull/312#discussion_r819157290) to group repositories into tiers like Core, Supporting, etc. Repo topics are an interesting way of storing metadata in a GH-native way that you might consider employing as you iterate on this @feanil .

Correct me if I'm wrong, but I get the sense that you are hoping to get this ADR over the line so that we can experiment with actually using backstage for the maintainer pilot. After we use backstage for a while it'll probably be easier to tell which GH features we do and don't want to tie in with, as well as what exactly the schema should look like. That is to say: I'm fine if you're not explicitly rejecting any alternatives at the moment, since we're not exactly sure what this thing will look like and how everyone will use it.

feanil:
Thanks for all the feedback everyone.  I'll be merging this in as `provisional` shortly.  If you have other concerns or feel there should be more items covered, feel free to open issues or PRs with the relevant info and I'm happy to address it.

PR #348, opened 2022-07-01
E0d/proposed footer changes

e0d:
This PR is part one of a proposal to address some of the issues with our current commit message conformance.  Conformance is measured against OEP-51.

One issue that is currently in play is that many commit messages include references to private content.  Some commit messages only contain private references and not additional narrative information.

This PR proposes that private references are allowed for convenience of contributors, but may only be provided in the footer using the `Private-ref:` token.  Commit subjects and bodies should always be necessary and sufficient to understand the purpose of the commit.

This PR is one step in addressing [an issue raised by community core contributors](https://discuss.openedx.org/t/core-contributor-sprint-retro-june-10th-june-24th/7657/3).

This PR is only the first step toward a general solution.

### DIFF
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -138,6 +138,8 @@ Footer
 ------
 
 Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.  No other footers are specified.
+Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.
+
 
 
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -140,6 +140,8 @@ Footer
 Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.  No other footers are specified.
 Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.
 
+Commits may optionally include the ``the Co-authored-by:`` footer to indicate when the commit is the result of a collaboration.
+
 
 
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -190,7 +190,9 @@ Here's an example of a conventional commit, with a one-line subject, and details
     - `make requirements` used `private.*` which included private.in, which
       pip-sync balks at.
 
-    Fixes: BOM-2345
+    Fixes: https://github.com/openedx/fake-example-repo/issues/1234
+    Co-authored-by: Alice Example <alice@example.com>
+    Private-ref: https://internal-work-tracker.example.com/BOM-2345
 
 .. _tooling section:
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -44,7 +44,7 @@ This is part of our `Change Transparency`_ initiative.
 Motivation
 ==========
 
-Developers should describe the changes they make.  This information is valuable to many audiences: deployers, integrators, end-users, operators, and so on. Developers are in the best position to accurately describe their work and its implications.
+Developers should describe the changes they make.  This information is valuable to many audiences: other developers, deployers, integrators, end-users, operators, and so on. Developers are in the best position to accurately describe their work and its implications.
 
 The nearer to the change that information is captured, the easier it will be to write, and the more accurate it will be.  Changes made by developers are consumed by many people, from the pull request reviewer to the final user of the product.  The more the developer can explain the change, the more easily the change will be understood and used, and the more efficient the overall development process will be.
 
--- a/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+++ b/oeps/best-practices/oep-0051-bp-conventional-commits.rst
@@ -128,7 +128,7 @@ The more information you can put in the body of the commit message, the better. 
 
 Breaking changes especially should have detailed information about the implications and alternatives, including a ``BREAKING CHANGE`` footer.
 
-Include references or links to supporting information, such as Jira or GitHub issues.  Some Jira issues will be private to edX.  It is better to link to a private issue than no issue at all. The issue could be made public, and at least its existence is made clear, so people can ask for the information.
+Include references or links to supporting information, such as Jira or GitHub issues.  However, the body should be necessary and sufficient for understanding the commit.  Links to private issues are support for the convenience of contributors, but should not be included in the body of the commit.  Rather, references or links to private issues must be included in the commit footer using the git trailer format.  References to private content must use the ``private-ref:`` token. It is always preferable for any linked issues to be public.
 
 Larger decisions should be recorded in Architectural Decision Records, as explained in `OEP-0019`__.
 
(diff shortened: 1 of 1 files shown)


oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -137,7 +137,11 @@ __ https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-doc
 Footer
 ------
 
-Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.  No other footers are specified.
+Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.
+
+Commits may optionally include the ``the Co-authored-by:`` footer to indicate when the commit is the result of a collaboration.
+
+Private references to content related to the commit may be included using the ``:private-ref:`` footer token.  Private references may only be included in the footer using this token.
```
nedbat, line 144: Earlier it was `private-ref:`, and here it is `:private-ref:`.  I think the first is correct.

felipemontoya (approved):
I'm fully in favor of this change

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -44,7 +44,7 @@ This is part of our `Change Transparency`_ initiative.
 Motivation
 ==========
 
-Developers should describe the changes they make.  This information is valuable to many audiences: deployers, integrators, end-users, operators, and so on. Developers are in the best position to accurately describe their work and its implications.
+Developers should describe the changes they make.  This information is valuable to many audiences: other developers, deployers, integrators, end-users, operators, and so on. Developers are in the best position to accurately describe their work and its implications.
```
kdmccormick, line 47: 👍🏻

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -128,7 +128,7 @@ The more information you can put in the body of the commit message, the better.
 
 Breaking changes especially should have detailed information about the implications and alternatives, including a ``BREAKING CHANGE`` footer.
 
-Include references or links to supporting information, such as Jira or GitHub issues.  Some Jira issues will be private to edX.  It is better to link to a private issue than no issue at all. The issue could be made public, and at least its existence is made clear, so people can ask for the information.
+Include references or links to supporting information, such as Jira or GitHub issues.  However, the body should be necessary and sufficient for understanding the commit.  Links to private issues are support for the convenience of contributors, but should not be included in the body of the commit.  Rather, references or links to private issues must be included in the commit footer using the git trailer format.  References to private content must use the ``private-ref:`` token. It is always preferable for any linked issues to be public.
```
kdmccormick, line 131:
```suggestion
Include references or links to supporting information, such as Jira or GitHub issues.  However, the body should be necessary and sufficient for understanding the commit.  Links to private issues are support for the convenience of contributors, but should not be included in the body of the commit.  Rather, references or links to private issues must be included in the commit footer using the git trailer format.  References to private content must use the ``Private-ref:`` token. It is always preferable for any linked issues to be public.
```

Non-blocking suggestion: I usually see Git trailers written with the first letter capitalized.

oeps/best-practices/oep-0051-bp-conventional-commits.rst
```diff
@@ -137,7 +137,11 @@ __ https://open-edx-proposals.readthedocs.io/en/latest/oep-0019-bp-developer-doc
 Footer
 ------
 
-Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.  No other footers are specified.
+Breaking changes must have a ``BREAKING CHANGE:`` footer in the body.
+
+Commits may optionally include the ``the Co-authored-by:`` footer to indicate when the commit is the result of a collaboration.
```
kdmccormick, line 142: Good call-out.

kdmccormick (approved):
Great change.

In the Rationale section near the bottom of the OEP, I suggest updating the example commit based on the new guidance:

```
Here's an example of a conventional commit, with a one-line subject, details in the body, and a footer::

    build: handle private.txt files properly

    The requirements/edx/private.txt file is for dev's own private package
    needs.  There are two installation mechanisms in edx-platform, and
    neither handled the file properly:

    - `paver install_prereqs` had the wrong file name.  The file was moved
      almost three years ago, and paver wasn't kept up.

    - `make requirements` used `private.*` which included private.in, which
      pip-sync balks at.

    Fixes: https://github.com/openedx/fake-example-repo/issues/1234
    Co-authored-by: Alice Example <alice@example.com>
    Private-ref: https://internal-work-tracker.example.com/BOM-2345
```

feanil (approved):
Looks good to me, I almost wonder if we should also update the `BREAKING CHANGES` footer to also follow the same format(First cap + Dashes) to make it more consistent but both are valid trailer field names so it's also fine as is.

+1 to Kyle's suggestion to capitalize the first word, it looks a little nicer but I don't consider that a thing that blocks the merge.

robrap:
[non-blocking] I admit I just did a quick search, but wondering if you've thought about how and if PR descriptions or PR templates should mimic any of these recommendations? I know it addresses a separate, but related concern.

ashultz0:
[non-blocking] I don’t have any objections to the principle but as more things are added to the ideal commit message it becomes less and less likely for humans to write one as they are trying to do other things. Can we add a template and instructions on how to use it into edx-platform and other central repos?

Also this pushes people to make fewer larger commits rather than breaking down work into individual change commits. But gihub's review process already does that so that may be moot.

kdmccormick:
@ashultz0 ,

> I don’t have any objections to the principle but as more things are added to the ideal commit message it becomes less and less likely for humans to write one as they are trying to do other things.

I hear you. Developers have a limited amount of time and energy, and we should be thoughtful whenever we begin asking everybody to do a new thing.

The reason I feel comfortable with this change, though, is because writing down what you are changing in a place others can see is a pretty base-level thing. It's something that the community has been asking for us to do better on literally for years. If the commit requirements here are too heavy when added to the other things developers need to do, then I think we should try to reduce the "other things" instead.

> Can we add a template and instructions on how to use it into edx-platform and other central repos?

[OEP-51](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html) seems like a pretty solid tutorial to me? If you we should link to it in more places (and/or make it more reader-friendly), though, I encourage you to make that change.

To help folks & hold them to these guidelines, I would love if we tightened up the commitlint workflow to check the links are accessible unless they are annotated by `Private-ref: `. @e0d , could you open a tCRIL eng issue for that?

> Also this pushes people to make fewer larger commits rather than breaking down work into individual change commits. 

@ashultz0 Does it? This change is asking that a commit that used to look like this:

```
feat: TICKET-123 add a thing

... more details ...
```

becomes this:
```
feat: add a thing

... more details ...

Private-Ref: TICKET-123
```

That's not very many keystrokes different.

(fwiw, the `Co-authored-by` footer is just a recognition of a pretty standard thing, which GitHub will automatically do for you when you work on someone else's commits)

oeps/best-practices/oep-0051-bp-conventional-commits.rst
bradenmacdonald, line 142: Nit: `include the ``the Co-authored-by:` has "the" twice; I don't think the second "the" belongs.

bradenmacdonald (approved):

e0d:
@nedbat as the original author, do you approve of this change?

nedbat (approved):

PR #351, opened 2022-07-06
fix: OEP-52 provisional notes

robrap:
1. Fixed the header to match other OEPs, using OEP-XX
in place of OEP-00XX.
2. Fixed the status to just read "Provisional".
3. Fixed the language of the note about future updates,
since this is no longer a draft under review.

Note: changes were minor enough I did not update the change history or modified date.

### DIFF
--- a/oeps/architectural-decisions/oep-0052-arch-event-bus-architecture.rst
+++ b/oeps/architectural-decisions/oep-0052-arch-event-bus-architecture.rst
@@ -1,5 +1,5 @@
 ================================
-OEP-0052: Event Bus Architecture
+OEP-52: Event Bus Architecture
 ================================
 
 .. This OEP template is based on Nygard's Architecture Decision Records.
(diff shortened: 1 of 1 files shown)


oeps/architectural-decisions/oep-0052-arch-event-bus-architecture.rst
```diff
@@ -29,7 +29,7 @@ OEP-0052: Event Bus Architecture
 Overview
 --------
 
-.. note:: Consider this OEP an early draft to provide intent and gather feedback. The language and content of the OEP will change over time as we progress on this work and hammer out certain decisions.
+.. note:: This provisional OEP is expected to be updated with references to new ADRs as work progresses.
```
rgraber, line 32: I'm hesitant to say we're only going to add ADRs. Maybe `with references to further documentation`?

robrap: I may drop the note altogether. Would you consider ADRs just adding clarity? I wasn’t sure, so I called them out. Other updates are already fine.

timmc-edx (approved):

oeps/architectural-decisions/oep-0052-arch-event-bus-architecture.rst
timmc-edx: I like the new "expected to be" language -- which I take to mean, "we might forget, but if you notice it's missing... please add!" :-P

PR #357, opened 2022-07-19
docs: Add a proposed process for transferring maintainership.

feanil:
There's lots of cleanup to do still but this outlines the process by
which we can find new maintainers for components and what to do if we
can't.


feanil:
There's a few todos left in here but I think the process content is a good starting points so please provide any feedback you might have on that.

idegtiarov (approved):
Thanks. LGTM!

 It seems that one more process gap is closing!

e0d (approved):
This looks good to me.

PR #362, opened 2022-08-01
feanil/copy button

feanil:
- feat: Add copy buttons to all code blocks.
- chore: Run `make upgrade`


jmbowman (approved):
Cool extension, looks good to me.  Maybe we should also add that to the cookiecutter for docs in new repos?

Looks like it's part of the same project that created [Markedly Structured Text](https://myst-parser.readthedocs.io/en/latest/); I wonder if that helps us at all in the reStructuredText vs. Markdown holy war or just makes things worse...

Issue #361, opened 2022-07-29
OEP-55: Change reference link to publicly accessible reference


giovannicimolin:
[Reference [5] in the OEP-55](https://github.com/openedx/open-edx-proposals/blame/master/oeps/processes/oep-0055-proc-project-maintainers.rst#L468) links to the article **What Makes a Great Maintainer of Open Source Projects?** behind a paywall (https://www.computer.org/csdl/proceedings-article/icse/2021/029600a982/1sEXoQoeO0E).

Is it possible to replace that URL with one without a paywall, hosted on the website of one of the authors? 
I've found it here: http://gustavopinto.org/lost+found/icse2021.pdf
And it's publicly discoverable from: https://gustavopinto.org/publications/

feanil:
@giovannicimolin good find and thanks for finding the alternative version.  I'm happy to make this PR but also happy to accept a PR for this change from you.  Let me know if you have time to make the PR to update the link or if you'd rather I take care of it.

PR #364, opened 2022-08-01
docs: Add an alternative link to a publication.

feanil:
The paper for `What makes a great maintainer of Open Source Projects?`
is behind a pay wall on the IEEE website but it is also available wihout
that restriction on the author's website.  So I added an alternative
link so people can get to the content.

I left the original link because the IEEE publication speaks to the fact
that it is a peer reviewed paper which lends to how much weight the
content carries.

fixes https://github.com/openedx/open-edx-proposals/issues/361


giovannicimolin (approved):
Looks good. :+1: 

That was fast! :rocket:

PR #375, opened 2022-08-31
docs: Small tweaks to ADR template

timmc-edx:
- Fix heading
- Use `1.` title instead of `0001`, since I think the former is more common (and readable)


timmc-edx:
@feanil @robrap @oliviaruizknott @hurtstotouchfire Tagging you because you were involved in the original PR adding the template. I don't think the numbering in the title was discussed, but maybe some of you have opinions on that.

robrap:
I'm fine with this. Not sure if @kdmccormick has an opinion on this?

kdmccormick (approved):

PR #377, opened 2022-09-07
docs: update archive repo OEP for 2U

robrap:
* Reversed order of "Change History".
* Remove example using older edX GitHub orgs to better clarify that this OEP only covers openedx org repos.

### DIFF
--- a/oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
+++ b/oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
@@ -74,17 +74,15 @@ First, if the repository is public, and a part of Open edX releases, follow thes
 Archive Steps
 =============
 
-These steps should be followed for all repos within the Open edX organization (forks included). After some experiments with keeping archived repos in the ``openedx`` organization, we've learned that having abandoned code show up in searches hinders work to understand the current state of the system and the risk around new work, particularly deprecations and API changes. Thus we decided to move all archived repositories to a separate org.
+These steps should be followed for all repos within the Open edX organization (forks included). After some experiments with keeping archived repos in the `openedx github organization`_, we've learned that having abandoned code show up in searches hinders work to understand the current state of the system and the risk around new work, particularly deprecations and API changes. Thus we decided to move all archived repositories to a separate org.
 
 1. Update the README.rst file in the repository to add a brief note about why the repo is being archived, and what is serving as its replacement (where applicable). This may be as simple as a linking to the appropriate DEPR ticket.
 
 2. Unless you have the relevant permissions to perform this step, create a new `GitHub Request on the tCRIL board`_  and ask them to do the following:
 
-   - Archive the repository per `GitHub's archive process`_
+  - Archive the repository per `GitHub's archive process`_
 
-   - Move the repository to the openedx-unsupported organization
-
-      - If the repo is not coming from the `openedx github org`_ then before moving it, rename it with a prefix of the source org's name. For example the ``notifier`` repo in the ``edx-solutions`` org wolud be renamed to ``edx-solutions-notifier`` before moving.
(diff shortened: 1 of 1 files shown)


oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -74,17 +74,17 @@ First, if the repository is public, and a part of Open edX releases, follow thes
...
 
 2. Unless you have the relevant permissions to perform this step, create a new `GitHub Request on the tCRIL board`_  and ask them to do the following:
 
-   - Archive the repository per `GitHub's archive process`_
+  - Archive the repository per `GitHub's archive process`_
 
-   - Move the repository to the openedx-unsupported organization
+  - Move the repository to the openedx-unsupported organization
 
-      - If the repo is not coming from the `openedx github org`_ then before moving it, rename it with a prefix of the source org's name. For example the ``notifier`` repo in the ``edx-solutions`` org wolud be renamed to ``edx-solutions-notifier`` before moving.
```
robrap, line 87: **Note to Reviewers:** I don't think tCRIL has any orgs other than `openedx`, so I removed this note and moved it to the 2U/edX private runbook. If we want something here for tCRIL or other community members, we can add a note, but it seemed to muddy the instructions if it isn't needed.

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -74,17 +74,17 @@ First, if the repository is public, and a part of Open edX releases, follow thes
...
+    - For repos in a 2U/edx owned org, first see the `private runbook with 2U adjustments to this process`_, using the edx-unsupported organization.
```
sarina, line 87: I don't really want 2U-specific things in this or any OEP, but... I think as long as it's clear that anyone else adopting this OEP can add their internal documents, that would be good. To that end, could you please make a new section that's like, "Community Adopters Additional Specifications" (or whatever) that then lists the private 2U thing?

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
```diff
@@ -172,46 +172,52 @@ Reasons rejected:
 Change History
 **************
 
-2017-01-18
+2022-09-07
```
sarina, line 175: I appreciate inverting the change history but could that be in a separate commit?

sarina (changes requested):

oeps/architectural-decisions/oep-0014-proc-archive-repos.rst
e0d: I think it would be quite strange to include this here and don't support it currently.  Is this solving for a problem that has occurred?  Is there another approach that could be employed?

robrap: My main concern is helping people easily find the documentation that they need. Allowing for private back-links is a simple solution to this problem, and could probably be done in a way that is fair and clean, even if potentially unorthodox from an open source perspective. I've got many more thoughts about this, but it should probably be in a separate Discuss topic.

> Is this solving for a problem that has occurred? Is there another approach that could be employed?
Yes. The problem is that I am trained to go to OEPs for cross-repo or cross-platform decisions. One possibility is that 2U/edX needs a repository of these types of decisions, and that our employees are trained to go there first. Then there is the possibility of people finding what they are looking for, and only having uni-directional links from private to public docs. But, I don't think this solves all the issues, and thus the Discuss topic.

For now, I can close this PR, because it sounds like this will no longer be a quick change.

robrap: I cleaned it up so it just removes the misleading text.

sarina (approved):

PR #378, opened 2022-09-09
docs: clarify some of the example catalog-info.yaml entries

nedbat:
From https://github.com/openedx/edx-rest-api-client/pull/231#issuecomment-1242311625 .

### DIFF
--- a/oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
+++ b/oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
@@ -105,10 +105,11 @@ References
 
     apiVersion: backstage.io/v1alpha1
     # (Required) Acceptable Values: Component, Resource, System
+    # A repo will almost certainly be a Component.
     kind: Component
     metadata:
       name: 'docs.openedx.org'
-      description: "This repository contains the root documentation site for https://docs.openedx.org"
+      description: "The root documentation site for https://docs.openedx.org"
       links:
         - url: "https://docs.openedx.org"
           title: "Deployed Site"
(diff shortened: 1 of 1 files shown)


feanil (approved):

oeps/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.rst
```diff
@@ -122,7 +123,9 @@ References
     spec:
 
       # (Required) This can be a group(`group:<group_name>` or a user(`user:<github_username>`)
-      owner: group:docs.edx.org-maintainers
+      # Don't forget the "user:" or "group:" prefix, and group names need their
+      # organization included: "group:openedx/devs", not "group:devs".
+      owner: group:openedx/docs.edx.org-maintainers
```
nedbat, line 128: @feanil do we agree that the organization is needed? You said the group is actually a name in Backstage, not a GitHub team name.

nedbat: The group name (and user name come to think of it) have to correlate with something. I can't write `group:nedbat-fictional-team-xyzzy` in this field.  What do the names have to correspond with?

feanil: It's just github team names or user names prefixed with either `user` or `group`.

nedbat: Must it be a team name in the `openedx` org?

feanil: Yes, it must be a team name in the `openedx` org for it to be properly linked.

PR #382, opened 2022-09-21
Put OEP templates in top level to make them easier to find

sarina:
When I was making my first OEP, I did not use a template as I did not see them buried in the processes directory. I propost to make them top-level to make them easier to find. I would also be ok with putting them into the oeps/ directory. Thoughts? @nedbat @feanil


Carlos-Muniz (changes requested):
Hey Sarina, I think this is a good idea. I wasn't aware there were templates, and these would have helped.

It does seem like there are some warnings related to the `toctrees`:

```
/home/docs/checkouts/readthedocs.org/user_builds/open-edx-proposals/checkouts/382/oeps/processes/oep-0001.rst:448: WARNING: toctree contains reference to nonexisting document 'processes/oep-templates/pep-based-template'
/home/docs/checkouts/readthedocs.org/user_builds/open-edx-proposals/checkouts/382/oeps/processes/oep-0001.rst:448: WARNING: toctree contains reference to nonexisting document 'processes/oep-templates/adr-based-template'
/home/docs/checkouts/readthedocs.org/user_builds/open-edx-proposals/checkouts/382/oeps/processes/oep-0001.rst:448: WARNING: toctree contains reference to nonexisting document 'processes/oep-templates/external-link-template'
looking for now-outdated files... none found
```

sarina:
@Carlos-Muniz thanks! I was going to hold on resolving errors until I get feedback on the substance of this change.

Carlos-Muniz (approved):

feanil:
substance makes sense to me!

sarina:
ugh, I can't do this because they're referenced in OEP-1 (and I think they should be) but that means they must be in OEP-1's directory structure - can't do relative linking (https://github.com/sphinx-doc/sphinx/issues/701)

some other solutions:
- Get them rendering somewhere on RTD and then put that rendered RTD link within the OEP-1 page
- Link to the actual templates in GitHub
- Add a top level directory with a single file that redirects to the oeps/processes/oep-templates directory (I don't like this one)

Any thoughts/other ideas?

nedbat:
I would say it doesn't make sense to render the templates anyway: do authors need to see an HTML page entitled "OEP-XXXX: OEP Template"?  They want to know how to get started making a .rst file of their own, so let's just link to the GitHub file.

sarina:
Thanks @nedbat that's what I was leaning towards. I've made the change, I hope :)

oeps/processes/oep-0001.rst
```diff
@@ -437,20 +437,15 @@ to capture the decision(s), as deemed appropriate by the Authors and Arbiter.
...
-* :ref:`PEP-based template <pep_based_template>` based on Python's PEP_ standard.
-* :ref:`ADR-based template <adr_based_template>` based on `Architecture Decision Records`_.
-* :ref:`External link template <external_link_template>` for OEPs with mostly external content.
+* `PEP-based template`_ based on Python's PEP_ standard.
+* `ADR-based template`_ based on `Architecture Decision Records`_.
+* `External link template`_ for OEPs with mostly external content.
 
-..
-  Add a hidden toctree to indicate that these files are referenced from here in the overall
-  table of contents.
```
sarina, line 446: As far as I can tell this never worked - https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0001.html - I don't see this in the table of contents at the top or the sub-section TOC here: https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0001.html#oep-structure-and-content

oeps/processes/oep-0001.rst
```diff
@@ -437,20 +437,15 @@ to capture the decision(s), as deemed appropriate by the Authors and Arbiter.
...
 
-.. toctree::
-    :hidden:
+.. _PEP-based template: https://github.com/openedx/open-edx-proposals/tree/master/oep-templates/pep-based-template.rst
+.. _ADR-based template: https://github.com/openedx/open-edx-proposals/tree/master/oep-templates/adr-based-template.rst
+.. _External link template: https://github.com/openedx/open-edx-proposals/tree/master/oep-templates/external-link-template.rst
```
sarina, line 447: these links won't work until this merges

feanil: Why make these external links?  You should refer to the new location directly so that we can detect broken links as a part of the build process in the future: https://docs.openedx.org/en/latest/documentors/references/quick_reference_rst.html#links-between-rst-documents

feanil (changes requested):

oeps/processes/oep-0001.rst
nedbat: I thought the external links were a good solution, since they are meant to be used as .rst files, not as published HTML rendered from .rst files.  The links-between-rst-documents will render the templates, which isn't helpful.

feanil: Ah, makes sense, I think I would have gone with the [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) directive for this but I don't feel strongly enough about that to hold this up.

feanil (approved):

PR #387, opened 2022-09-28
docs: Clarify next steps when OEPS merge provisionally.

hurtstotouchfire:
There are 12 OEPs in the "Provisional" status. I would like to update them all in one of the following ways.

### Option 1: Change their status

- Update statuses to "Deferred" or some other status that doesn't imply in-progress next steps

- This may also involve adding new responsible parties because in some cases the authors or arbiters are no longer working in the Open edX community.

### Option 2: Create a rollout doc

- Create a wiki page (or link to an existing one) that indicates their status, progress, and ideally what needs to happen before they go back "Under Review" to be "Accepted". This page would then be linked in the references section.

- For some OEPs, this page has been created organically, though we may not have linked the page in question on the OEP references. Examples:
  - https://github.com/openedx/platform-roadmap/issues/28 is a roadmap for [OEP 52](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0052-arch-event-bus-architecture.html)
  - [Maintainership Pilot](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3426844690/Maintainership+Pilot) is a status page for [OEP-55](https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0055-proc-project-maintainers.html)


oeps/processes/oep-0001.rst
```diff
@@ -328,7 +328,11 @@ be merged to capture a set of edits, and to make the proposal more visible to
 community comment. From that point, additional pull requests can be opened to
 edit the OEP, until it converges to being either "Accepted" or "Rejected".
 
-If an OEP has Draft or Under Review status and the PR is under review, you can either use the intended final status (e.g. Provisional, Accepted, etc.), or you can clarify both the current and intended status using something like the following: "Under Review (=> Provisional)". Either of these options is especially useful if the merged status is not intended to be Accepted.
+When an OEP merges with any non-final status, use the References section to
+link to a page which gives readers an updated status. For Provisional OEPs which
```
nedbat, line 332: Can we say the page is a wiki page, to give people some guidance? Or are there other possibilities?

oeps/processes/oep-0001.rst
```diff
@@ -328,7 +328,11 @@ be merged to capture a set of edits, and to make the proposal more visible to
...
+need to be implemented, this should be a top-level rollout plan.
```
nedbat, line 333: Do we also want an implementation plan page for Accepted OEPs? I think we say Accepted when everyone agrees it's a good idea, and then implementation follows. For that matter, Draft and Under Review could also have plan pages for finishing and implementing the OEP. For example: https://openedx.atlassian.net/wiki/spaces/COMM/pages/3511091214/OEP-57+Update+Game+Plan

oeps/processes/oep-0001.rst
hurtstotouchfire: I would rather specify the key than the value if that makes sense, e.g. what field will have the link? Can it have a consistent name such as "Rollout Plan" or "Working Plan"? I used "References" in the language here just to follow precedent.

oeps/processes/oep-0001.rst
hurtstotouchfire: Frankly yes I would love implementation plans for accepted OEPs as well. One of the use cases in my mind is when my team started getting Hooks and Signals PRs a year after it was approved but had no idea there was still implementation that might affect us. Also of course we ignored the PRs for a month 😮‍💨 before I realized they were part of an OEP, because I have never made it all the way through the OEPs... So yes, I think anything that can improve visibility around "implementation status" for OEPs is helpful!

kdmccormick: I like this proposal. Here's a suggested reword that would standardize the "key" as well as encourage an implementation plan regardless of status:

> When an OEP PR calls for significant work after it merges, add a link named
> *Follow-up Work* to the *References* section of the OEP header. Use the linked
> page to keep readers up-to-date on the plan for completing and/or implementing
> the proposal. For OEPs merging with the status of Draft or Provisional,
> a *Follow-up Work* link is required.

hurtstotouchfire: Thanks @kdmccormick I've made that update!

e0d (approved):
I clarified with @hurtstotouchfire that a link to a GitHub PR against the original OEP would quality as an acceptable link for the follow-up work documentation.  I approve.

PR #390, opened 2022-10-05
docs: amend OEP-18 to further discourage VCS URL dependencies

kdmccormick:
Part of https://github.com/overhangio/2u-tutor-adoption/issues/84
Related to https://github.com/openedx/.github/pull/26

### DIFF
--- a/oeps/best-practices/oep-0018/decisions/0001-forbid-url-dependencies.rst
+++ b/oeps/best-practices/oep-0018/decisions/0001-forbid-url-dependencies.rst
@@ -1,0 +1,287 @@
+1. Forbid VCS URL Python Dependencies
+#####################################
+
+Status
+******
+
+Accepted
+
+Follow-up work:
+
+* https://github.com/openedx/public-engineering/issues/162
+* https://github.com/openedx/public-engineering/issues/163
+
+
+Context
+*******
+
+Former Guidance
+===============
+
+``pip`` allows dependencies to be installed via VCS (version control system,
+e.g. git) URL, for example::
+
+    git+https://github.com/edx/edx-ora2.git@2.1.15#egg=ora2==2.1.15
+
+OEP-18 guidance on this was previously:
+
+  ... you should generally avoid installing requirements from a URL
+  or local directory instead of PyPI.  But there are a few circumstances where
+  it can be appropriate:
+
+  * You need to test a release candidate of the dependency to make sure it will
+    work with your code.
+  * You critically need a fix for a package which has not yet been included in
+    a release, and you cannot arrange for a release to be made in a timely
+    manner.
+
+  In most other circumstances, the package should be added to PyPI instead.  If
+  you do need to include a package at a URL, it should have both the package
+  name and version specified (end with "#egg=NAME==VERSION"). ...
+
+Issues with Former Guidance
+===========================
+
+Performance
+-----------
+
+Firstly, installing from VCS URL is significantly slower than installing from PyPI.
+We understand this in theory, based on the facts that:
+
+* ``pip`` must download and build a repository's entire contents when freshly installing
+  via VCS URL, whereas PyPI-hosted packages can be downloaded as pre-built wheels; and
+* when re-installing a dependencies, ``pip`` must fully re-install VCS-hosted
+  packages, whereas it will skip over any PyPI-hosted package whose version number has not
+  changed.
+
+We confirmed our theoretical understanding with a `cursory performance comparison`_.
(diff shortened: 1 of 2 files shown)


oeps/best-practices/oep-0018/decisions/0001-forbid-url-dependencies.rst
```diff
@@ -0,0 +1,287 @@
...
+    a release, and you cannot arrange for a release to be made in a timely
+    manner.
+
+Generally, fixes can be achieved by reverting to a previous, stable package version.
+In the rare event that reverting to a previous package version is not viable,
+Open edX repository maintainers can easily cut new releases of packages in order
+to accomodate a forward fix. If a repository maintainer is unresponsive, this
+issue can be escalated to tCRIL, which can cut an emergency release if necessary. If
+the repository is outside of the Open edX project, then the repository can be forked, and
+the package re-released to PyPI under a new name.
```
jmbowman, line 102: One of the most common historical causes of VCS dependencies is needing support for a new Python or Django version for a named Open edX release before upstream is ready to merge and/or release it (because Open edX named releases come out before Python/Django EOL dates).  The absolute number is still small, but probably happens about 5 times across repos each Django upgrade cycle.

We can technically release forks of these to PyPI each time it happens, but just want to make clear that this will cause a fair bit of overhead each time it happens; we'd want a pretty solid workflow for this, and the "temporary" forks will pretty much sit in PyPI until we're willing to break all installations of the named release that needed it.

oeps/best-practices/oep-0018/decisions/0001-forbid-url-dependencies.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Finally, even if those reasons were valid, they are *not* the reasons behind
+`the VCS URL dependencies installed in to edx-platform as of Nutmeg
+<https://github.com/openedx/edx-platform/blob/open-release/nutmeg.2/requirements/edx/github.in#L54-L77>`_
+Following ``git blame`` shows that the current VCS URL dependencies
+exist simply because it took fewer steps to set them up than it would have to
+install them from PyPI.
```
jmbowman, line 109: A few other interesting use cases that have come up in the past:

* A dependency suddenly deletes all of its old PyPI releases, as in https://github.com/openedx/edx-platform/pull/27168 .  This removes the "downgrade" option and makes it critical to fix the dependency ASAP because it's breaking all deployments and installations.
* Something about codejail is preventing it from working correctly when installed from PyPI; Arbi-BOM has a new idea regarding what causes that and is currently looking into it, but it's not fixed yet.  https://openedx.atlassian.net/browse/BOM-2481
* We had to temporarily fork celery just to get useful logging data to diagnose a recurring problem in production: https://github.com/openedx/edx-platform/pull/25649
* Sometimes upstream drops previous LTS Django support before releasing a version with the new LTS Django support, like in https://github.com/openedx/edx-platform/pull/23257 .

I don't want to argue that installing from VCS is the only way to solve problems like these.  I just want to make sure that we have a replacement process that allows us to very quickly deploy a specific commit from a dependency before we eliminate this as an option.

kdmccormick: Good to know @jmbowman . I'm glad you saw this draft PR and brought these up before I went any further with this.

I believe the codejail issue was an [instance of an incorrect setup.py](https://github.com/openedx/codejail/pull/140). I don't have good alternative workflows for the other instances you brought up, though.

To be honest, I don't think BOM teams adding temporary URL dependencies in order to speed up upgrades and handle critical issues is the problem here. The problem is when URL dependencies are added as a quick-and-dirty way of getting a feature out the door. So, maybe the solution isn't to forbid VCS URLs altogether. Perhaps it'd be better to, for example, use CODEOWNERS to require review from one of the BOM teams when editing github.in, helping ensure the file isn't misused.

jmbowman: I definitely agree that there are things in github.in that have stayed there far too long.  Like, there's never been a proper release of our drag & drop assessment XBlock.  But we may need something more like a repo health check and a followup process for remediation than an outright ban.

kdmccormick:
Based on Jeremy's feedback, I will be amending this update to be less extreme. I'll leave room in the OEP for using GitHub-based dependencies in the cases he listed.  And instead of a blocking linter, I'm going to look into using CODEOWNERS or a repo health check.

kdmccormick:
Superseded by https://github.com/openedx/open-edx-proposals/pull/450

PR #392, opened 2022-10-05
fix: Add wiki link for follow up work on OEP 55.

hurtstotouchfire:
Per #387 adding a follow up link for the maintainers OEP.

### DIFF
--- a/oeps/processes/oep-0055-proc-project-maintainers.rst
+++ b/oeps/processes/oep-0055-proc-project-maintainers.rst
@@ -28,7 +28,7 @@ OEP-55: Project Maintainers
    * - Resolution
      - <links to any discussions where the final status was decided>
    * - References
-     - `Follow Up Work <https://openedx.atlassian.net/wiki/spaces/COMM/pages/3426844690/Maintainership+Pilot>`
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0055-proc-project-maintainers.rst
```diff
@@ -28,7 +28,7 @@ OEP-55: Project Maintainers
    * - Resolution
      - <links to any discussions where the final status was decided>
    * - References
-     - <links to any other relevant discussions or relevant related materials>
+     - `Follow Up Work <https://openedx.atlassian.net/wiki/spaces/COMM/pages/3426844690/Maintainership+Pilot>`
```
kdmccormick, line 31:
```suggestion
     - `Follow-up Work <https://openedx.atlassian.net/wiki/spaces/COMM/pages/3426844690/Maintainership+Pilot>`_
```

* s/Follow Up/Follow-up
* trailing underscore is needed to turn this into an RST link

kdmccormick (approved):
Thanks for the update! Just a couple requests, but no need for re-review from me.

PR #410, opened 2022-11-09
OEP-65 Composable Micro-frontends

arbrandes:


oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,171 @@
...
+   * - OEP
+     - :doc:`OEP-XXXX <oep-XXXX-micro-frontend-domains>`
+   * - Title
+     - Micro-frontend Domains
+   * - Last Modified
+     - 2022-11-09
+   * - Authors
+     - Adolfo R. Brandes <adolfo@tcril.org>
+   * - Arbiter
+     - TBD
```
adamstankiewicz, line 17: @arbrandes Feel free to list me (astankiewicz@2u.com) as an arbiter for this OEP. I'll brush up on the responsibilities of the arbiter role [here](https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0001.html#step-3-review-with-arbiter).

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,171 @@
...
+   * - References
+     - TBD
+
+
+********
+Abstract
+********
+
+The decision to gradually implement a micro-frontend architecture for Open edX was a largely successful attempt at avoiding some limitations of the monolithic ``edx-platform`` codebase.  In particular, it allowed the project to integrate more and better features by leveraging newer frontend technology, all much faster than would have been possible otherwise.  At the same time, because features encapsulated in micro-frontends (henceforth, "MFEs") are very loosely coupled to each other, this architecture eliminates most integration bottlenecks, allowing individual teams working on any given feature to not be a lot more *productive*.
```
adamstankiewicz, line 36: nit: Is the "not" here intended? Wouldn't the loose coupling between MFEs, eliminating most integration bottlenecks, would allow teams to be productive, but "allowing individual teams working on any given feature to **not** be a lot more *productive*." makes it sound like the loose coupling is responsible for a loss of productivity.

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,171 @@
...
+
+However, roughly 3 years into this gradual re-platforming, several drawbacks to the chosen implentation of complete single-page app ("SPA") independence have been identified.  This OEP proposes a middle-ground based on Domain Driven Development ("DDD") and current industry best-practices, where a minor increase in integration costs is traded for large potential gains in user experience, customizability, ease of deployment, and in certain cases, even development efficiency.
+
+
+**********
+Motivation
+**********
+
+When moving to micro-frontends, a (now known to be) uncommon choice was made: to have *completely* independent SPAs, each with their separate deployment, endpoint, and different technology stack.  While the goal of developing platform features more efficiently was achieved, the price to pay was not insignificant.  Here are some critical issues that the new architecture introduced:
```
adamstankiewicz, line 45: nit: what is meant by "endpoint" here? Is this referring to the URL subdomain where the MFE is hosted?

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,171 @@
...
+  * The fact each MFE is fully independent means that it is harder to implement a consistent plugin interface between all of them.
+
+* Changes that affect multiple MFEs are costly
+
+  * Development effort is multiplied by the number of existing MFEs, as each change needs to be duplicated;
+  * Because no dependencies are shared, Webpack build time of any potentially shared library is multiplied by the number of existing MFEs.
+
+* Inefficient use of network bandwidth
+
+  * Dependencies have to be downloaded multiple times by the browser.
```
adamstankiewicz, line 69: In a similar vein, it may also be worth calling out that the current architecture limits our ability to take advantage of routing (e.g., with `react-router`) between MFEs. As is, to navigate from "MFE A" to "MFE B", we have to do a full page refresh, rather than more intelligent routing that avoids refreshing the whole page when navigating between MFEs.

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,171 @@
...
+#. ...
+
+TODO:
+
+* What is the relationship of this project to the Modular Learning effort?  They're both long-term projects, but is there anything else that could be coordinated between them?  Maybe this project should only target the modular codebase, and ignore ``edx-platform`` entirely?  (Are the APIs between the two even the same?)
+
+Integration strategy
+====================
+
+At time of writing there appear to be three commonly recommended strategies for integration of MFEs into container apps:
```
adamstankiewicz, line 110: There are also micro-frontend frameworks that are worth considering as well, that may or may not be compatible with these other listed strategies, e.g.: `single-spa` ([source](https://single-spa.js.org/)) and `Piral` ([source](https://github.com/smapiot/piral)).

adamstankiewicz (commented):
Just did a quick read-through and added some initial feedback. I'll come back for a deeper dive later :) I'm excited about this OEP!

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
arbrandes: It is a bad attempt at a catch-all term for individual MFEs having different apparent domains/ports/URLs even though they're supposed to be perceived as being part of the same app, and also the fact they are potentially hosted separately.  I'll think of something better to say, here.

oeps/architectural-decisions/oep-XXXX-micro-frontend-domains.rst
```diff
@@ -0,0 +1,208 @@
...
+
+To be defined.
+
+Discovery
+---------
+
+At time of writing these appear to commonly recommended strategies for integration of MFEs into modular container apps:
+
+* `Run-time integration via JavaScript <https://martinfowler.com/articles/micro-frontends.html#Run-timeIntegrationViaJavascript>`_
+* `Run-time integration via Web Components <https://martinfowler.com/articles/micro-frontends.html#Run-timeIntegrationViaWebComponents>`_
```
brian-smith-tcril, line 131: in the FWG meeting today the question of "what's going to be around in 5-10 years?" was raised in the context of which integration strategy to use. my gut feeling is web components would be a good choice based on that criteria

adamstankiewicz: FWIW, there is some desire/interest to try to have a web components implementation available for the components in the Paragon design system so that they could be potentially used in non-React environments (definitely an open question in what that would look like practically, but one of our long-term "vision" goals for Paragon is for it to become more framework agnostic in its usage).

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Shared dependencies
+===================
+
+To be defined.
+
+Discovery
+---------
+
+* What dependencies should be shared between all MFEs?  (React, Redux, ...?)
```
brian-smith-tcril, line 171: `react-router` comes to mind, this may be part of a slightly larger conversation about how to handle routing with links between MFEs (since needing a full refresh when navigating from one to another is a pain point in the current SPA architecture)

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+
+With these in mind:
+
+* What are the pros and cons of each integration option, in particular in terms of:
+
+  * Cost of conversion: how easy is it to convert existing MFEs to the new paradigm?
+  * Cost of integration: compare the level of MFE independence.
+  * Hosting of build artifacts: is it possible to host build artifacts separately on object storage or CDN?
+  * Lazy loading: is it possible to load MFEs and other components lazily?
+  * Efficiency of development: consider the level of boiler-plate code.
```
adamstankiewicz, line 145: Related, not so much in efficiency of writing code, but in being able to spin up a local dev environment: How many MFE repos do I need to have checked out and running locally in order to run an entire application? If I'm only making changes to one MFE, can I only run that one MFE locally while running the production build of all the other MFEs in the application (aka only run the local MFEs you need to for your given development task).

What tooling might exist to be able to quickly spin up/down relevant MFEs during local development?

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+
+* `Run-time integration via JavaScript <https://martinfowler.com/articles/micro-frontends.html#Run-timeIntegrationViaJavascript>`_
+* `Run-time integration via Web Components <https://martinfowler.com/articles/micro-frontends.html#Run-timeIntegrationViaWebComponents>`_
+* `Webpack Module Federation <https://webpack.js.org/concepts/module-federation/>`_
+* `single-spa <https://single-spa.js.org/>`_
+* `Piral <https://github.com/smapiot/piral>`_
+
+With these in mind:
+
+* What are the pros and cons of each integration option, in particular in terms of:
```
adamstankiewicz, line 139: A couple other considerations:
* Is it compatible with something like Next.js for server-side rendering on top of React? Should it be?
* Should the composition of MFEs happen in the browser or on a server? How is performance impacted? How is the UX impacted?

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+#. Integration strategy
+#. Domain SPAs and their interfaces
+#. Shared dependencies
+#. MFEs as components
+
+Discovery
+=========
+
+* Are there any further top-level architectural elements to consider?
+* What is the relationship of this project to the Modular Learning effort?
```
adamstankiewicz, line 116: [curious] What is the "Modular Learning effort"?

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+Domain SPAs and their interfaces
+================================
+
+To be defined.
+
+Discovery
+---------
+
+* What UX elements should go into the domain's containing app? The header? Footer?
+* Define further container SPA and interface characteristics.
```
adamstankiewicz, line 161: "Global" data like authentication state might make sense to live in the domain's containing app, which currently comes from `@edx/frontend-platform` and it's `AppProvider` component.

In the Enterprise domain, we might consider treating authentication state, the metadata about the enterprise customer (e.g., its name), the subsidies available to the enterprise/authenticated user (e.g., subscription license) as "global" data applicable to all the potential MFEs in this domain.

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Domain SPAs and their interfaces
+================================
+
+To be defined.
+
+Discovery
+---------
+
+* What UX elements should go into the domain's containing app? The header? Footer?
```
adamstankiewicz, line 160: An initial thought here... I believe the majority of the UX elements in the container would be more layout based in terms of deciding when MFEs should be dynamically loaded and where they should go in the UI (e.g., there could be more than 1 MFE to dynamically load for a given route in the container app, and the container app decides the layout for displaying each, like a main content region and a sidebar region, for example).

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+* Define further container SPA and interface characteristics.
+
+Shared dependencies
+===================
+
+To be defined.
+
+Discovery
+---------
```
adamstankiewicz, line 169: I think we'll want to deep dive a bit in terms of how version conflicts and breaking changes should be handled in this architecture (e.g., is it smart enough to fallback to older version for libraries in some MFEs, but not others?).

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
adamstankiewicz, line 171: `@edx/paragon`? I would imagine having a single Paragon version should lead to a more consistent UX in terms of design and functionality across the domain.

adamstankiewicz, line 171: `@edx/frontend-platform`

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+* `single-spa <https://single-spa.js.org/>`_
+* `Piral <https://github.com/smapiot/piral>`_
+
+With these in mind:
+
+* What are the pros and cons of each integration option, in particular in terms of:
+
+  * Cost of conversion: how easy is it to convert existing MFEs to the new paradigm?
+  * Cost of integration: compare the level of MFE independence.
+  * Hosting of build artifacts: is it possible to host build artifacts separately on object storage or CDN?
```
adamstankiewicz, line 143: Related to another comment, is (or should be) the integration strategy compatible with server-side rendering? Could an MFE be created with server-side rendering (e.g., with Next.js) but still be consumed by a container/host application?

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
adamstankiewicz, line 161: Hinted at in another comment, but is there any eventing system like PubSub for communication between MFEs and/or an MFE and its container app as part of the interface?

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+* `Webpack Module Federation <https://webpack.js.org/concepts/module-federation/>`_
+* `single-spa <https://single-spa.js.org/>`_
+* `Piral <https://github.com/smapiot/piral>`_
+
+With these in mind:
+
+* What are the pros and cons of each integration option, in particular in terms of:
+
+  * Cost of conversion: how easy is it to convert existing MFEs to the new paradigm?
+  * Cost of integration: compare the level of MFE independence.
```
adamstankiewicz, line 142: semi-related to integration costs: As a consumer of Open edX, if I want to customize/replace one MFE for another to change/swap functionality in the container app or run an experiment, would that be supported? E.g., using Webpack 5 module federation, could I change the path to the `remoteEntry.js` file to point to a different MFE instead?

adamstankiewicz, line 142: If an exposed component from an MFE changes its API but doesn't inform its consumers, how should/would errors be handled? How will this error handling impact the user experience? An example solution I've seen for this issue is to fallback to an installed version known to be compatible; that way, if the upstream components unintentionally creates a breaking change, it may be able to fallback to a known compatible experience? If that's not possible, ideally, the error would prevent the entire application from breaking and instead only break that one part of the application.

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
ghassanmas: My gut feelling is that it realtes to this https://openedx.atlassian.net/wiki/spaces/OEPM/pages/3470196737/Modular+Learning

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
adamstankiewicz: The Paragon library supports tree shaking, so it only includes the necessary code for the components an MFE is actually using, which should mitigate this concern. As far as I can tell, I don't believe that existing tree shaking behavior would be impacted even with a single copy of Paragon shared across multiple MFEs in a domain app.

adamstankiewicz: Also related to `@edx/frontend-platform`, it'll be interesting to consider how our current MFE bootstrapping / initialization strategy in `@edx/frontend-platform` will need to change, if at all, in order to support this architecture.

ghassanmas: Aha but what about the case where the MFE is expected to use a shared version of paragon, which I expect suppose to be available at runtime? _Below is a scinario of how I imagine but I might be wrong_ 

- Paragon has compoenets: (A, B, C, D, E, F, G,H) 
- MFE-1: depends on (A, B, C E, F)
- MFE-2 depends on (C, D, G). 

The de-facto situation wihhout having the host app...etc is that webapck suppose to to do tree shakeing when building MFE-1 and MFE-2, such that they will denepd on a version of paragon that has a subset of paragon compoenents. 

However in the host/container app siutation, then we are thinking we have a copy of paragon for all other MFEs, _so we don't load paragon everytime a user change MFE_? If yes then which componenet do we initially include in the host app, my top of head suggestion can be: 

1. Just include all of them, _if the added complexity doesn't worth the saved bandwidth..._ 
1. Include an _Union of all MFEs paragon compoenets_  i.e. $MFE1 \cup MFE2... \cup MFEn = PU$
  - The more MFEs we have the closet $PU$ gets to $P$ where $P$ is a set of all the compoenet 
2. Incldue an _Intersection of all MFEs paragon compoenets_ i.e $MFE1 \cap MFE2 ...\cap MFEn = PI$
  - The more MFEs we have then probably the less compoenents $PI$ would contain. 
  - This path of course would add a bit complexity beacus  loading a particualr  MFE, then we need to load the difference in compoenets, i.e $MFE1 - PI$  _The component that are in $MFE1$ but not in $PI$
    - One other example is load each compoenet just it loads, but that of course cause that for each componenet in the MFE that doesn't exists in the third set, we to have a roundtrip to the server

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
ghassanmas: Also good to add to the context, that webpack module federation isn't mature yet in development mode... The important missing featuer is **[Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/)**.

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
ghassanmas: Related to SSR
- Can we utilize the lms/cms-worker to do SSR rendering. i.e. what is the trade of utilizing/extending current servive vs. intergarting new one like Next.js

arbrandes:
Status update: I didn't have time to reply to and incorporate comments (thanks for them, btw!) last sprint because of the Olive upgrade work, but tCRIL is moving forward with selecting a contractor to start answering the questions posed.

More on this as work starts on discovery.

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+Abstract
+********
+
+The decision to gradually implement a micro-frontend architecture for Open edX was a largely successful attempt at avoiding some limitations of the monolithic ``edx-platform`` codebase.  In particular, it allowed the project to integrate more and better features by leveraging newer frontend technology, all much faster than would've been possible otherwise.  At the same time, because features encapsulated in MFEs are very loosely coupled to each other, this architecture eliminates most integration bottlenecks, allowing individual teams working on any given feature to be a lot more *productive*.
+
+However, roughly 3 years into this gradual re-platforming, several drawbacks to the choice of complete MFE independence have been identified.  This OEP proposes a middle-ground based on DDD and other industry best-practices, where a minor increase in integration costs is traded for large gains in user experience, customizability, ease of deployment, and development efficiency.
+
+
+**********
+Motivation
```
davidjoy, line 51: I think we need to be careful to zero in on what problems we hope MFE modularity (is composability the right word?) will solve.  There are a lot of problems listed here as motivation that are legitimate critiques of the architecture, but which I don't believe will be appreciably improved by this effort.

My take is that the main things we hope to get out of this are:

1. Reduced overall bundle size via shared files at runtime
2. Improved page load performance and cross-page UX via loading a 'shell' application once that contains both common UI elements, like headers and footers, and shared files.
3. Prioritizing "consistency" over "flexibility" - this is a choice we made early on in favor of flexibility.  We allowed all MFEs to have shared libraries, yes, but to use whatever version of those they wanted.  I think we probably went too far, and are learning we care more about consistency.

These correspond roughly to "Changes that affect multiple MFEs are costly" and "Inefficient use of network bandwidth" below.

Things this does not significantly impact:

1. Code reusability - Having a place to put shared code so we can reuse it - we have shared libraries that all MFEs use - Paragon, frontend-build, frontend-platform, the header and footer libraries - if you want to add a plugin interface, or consistent UI elements, we already have places to put those.
2. Documentation - loading our code differently doesn't impact the documentation written for our independent repos.
4. UX design - our design teams coordinate through Paragon to make consistent experiences.  A technical solution for code loading won't affect that much.  Prioritizing consistency of version numbers _will_ help, yes, but it won't change how designers work on its own.
5. Deprecation of old UIs - I'm not sure how this changes as a result of this.
6. Changes across MFEs that aren't covered by the shared libraries - I'm not convinced this is common... if something is used by many MFEs, it should be in a shared library already (i.e, paragon, frontend-platform, etc.)

Long story short, I'm wondering if we can tighten up this list of motivations so we can prove we're addressing them with the proposed new technical capabilities.

oeps/architectural-decisions/oep-XXXX-modular-micro-frontend-domains.rst
```diff
@@ -0,0 +1,209 @@
...
+====================
+
+To be defined.
+
+Discovery
+---------
+
+At time of writing these appear to commonly recommended strategies for integration of MFEs into modular container apps:
+
+* `Run-time integration via JavaScript <https://martinfowler.com/articles/micro-frontends.html#Run-timeIntegrationViaJavascript>`_
```
davidjoy, line 131: I believe @adamstankiewicz mentions this elsewhere, but I think we should seriously consider Next.js or Remix to solve these problems, not just purely client-side solutions.  An important critique of our MFE strategy is that we threw out the server completely.

arbrandes:
Thanks everybody for the input over the lifetime of this draft, but I'm now closing this in favor of https://github.com/openedx/open-edx-proposals/pull/575.  It takes into account all the requirements and work done here, but goes in a different technical direction.  I suggest folks take a peek. ;)

PR #418, opened 2022-12-01
docs: Update the security announcement process

feanil:
Make it so we announce on discourse to everyone when there will be a
security update coming instead of just to the security-notifications
mailing list.  Given that it's pretty easy to get on the
security-notifications mailing list, this doesn't make things less
secure (django and other major projects use the approach we're moving
to).


alangsto (approved):
👍 thank you!

PR #452, opened 2023-02-24
docs: Remove old information about installation methods.

feanil:
We used to support different installation methods when the release
process was first defined, remove mentions of the old methods and link
to the current methods.

### DIFF
--- a/oeps/processes/oep-0010-proc-openedx-releases.rst
+++ b/oeps/processes/oep-0010-proc-openedx-releases.rst
@@ -37,7 +37,7 @@ process.
 Motivation
 **********
 
-Open edX releases happen roughly every six months. EdX engineers are typically
+Open edX releases happen roughly every six months. Open edX engineers are typically
 focused on the much more frequent edx.org releases.  This document standardizes
 aspects of the release process to ensure that all involved understand and
 participate appropriately.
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0010-proc-openedx-releases.rst
```diff
@@ -187,58 +187,25 @@ Currently, none of the supported installation methods are intended for
...
-    We haven't determined how best to allow developers to configure which
-    services to run and which should be editable.
-
-The native installation will use an Ansible playbook to install Open edX
-components onto the machine.
-
-.. note::
-
-    We will update this OEP later with specifics of the playbook used.
+- `Tutor <https://overhang.io/tutor/>`_
```
sarina, line 197: Should we be opinionated here ("Tutor is the community supported installation...") or just leave as-is? I'm not sure I feel super strongly as the sites themselves probably convey that.

sarina (approved):

oeps/processes/oep-0010-proc-openedx-releases.rst
feanil: I think this is a good improvement and because of the complex state if usability of tutor for dev and it's intent to be production capable, I decided to not say a lot more and let the links speak for themselves.

oeps/processes/oep-0010-proc-openedx-releases.rst
```diff
@@ -37,7 +37,7 @@ process.
 Motivation
 **********
 
-Open edX releases happen roughly every six months. EdX engineers are typically
+Open edX releases happen roughly every six months. Open edX engineers are typically
```
kdmccormick, line 40: @feanil This change doesn't jive with the following sentence.

feanil: Good spotting thanks.  I think I'll drop the mention of edx.org entirely.

PR #466, opened 2023-04-18
fix: updating a broken link

johnnagro:
Updating the link for the wiki area for the Rest Best Practices.


rgraber (approved):
LGTM, good catch

PR #471, opened 2023-04-26
Fix examples of enumerations and attrs in OEP-49 App Patterns: data.py

bradenmacdonald:
In [OEP-49](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0049-django-app-patterns.html#data-py), it gives this example of how to use [`attrs`](https://www.attrs.org/en/stable/) in data.py:

```python
from enum import Enum

import attr

def ProgramStatus(Enum):
    ACTIVE = "active"
    RETIRED = "retired"

@attr.attrs(frozen=True)
class ProgramData:
    uuid: str = attr.attrib(
        validator=[attr.validators.instance_of(str)]
    )
    title: str = attr.attrib(
        validator=[attr.validators.instance_of(str)]
    )
    status: str = attr.attrib(
        validator=[
            attr.validators.instance_of(str),
            attr.validators.in_(ProgramStatus)
        ]
    )
```

There are two problems with this:
1. It should not be using `def` but rather `class` to define the enum. As shown, the declaration is accepted but the enum won't work.
2. The `status` field of `ProgramData` will not actually accept any value, because enum members are not strings and strings are not enum numbers, so the validators conflict:

```python
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE)
...
TypeError: ("'status' must be <class 'str'> (got <ProgramStatus.ACTIVE: 'active'> that is a <enum 'ProgramStatus'>).", ...)
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE.value)
...
ValueError: ("'status' must be in <enum 'ProgramStatus'> (got 'active')", ...)
```

I believe that a better, working, and modern example would be this, which is what I've put in the PR:

```python
from enum import Enum

from attrs import field, frozen, validators

class ProgramStatus(Enum):
    ACTIVE = "active"
    RETIRED = "retired"

@frozen
class ProgramData:
    uuid: str = field(validator=validators.instance_of(str))
    title: str = field(validator=validators.instance_of(str))
    status: ProgramStatus = field(validator=validators.in_(ProgramStatus), converter=ProgramStatus)
```

This works, is more concise, uses standard python enums, and also uses [the more modern `attrs` syntax](https://www.attrs.org/en/stable/names.html) which they recommend for all new code.

Putting `converter=ProgramStatus` allows this new type to accept three different forms of the enum value, but coerces them all to a standard form for consistency:

```python
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE)
ProgramData(uuid='', title='', status=<ProgramStatus.ACTIVE: 'active'>)
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE.value)
ProgramData(uuid='', title='', status=<ProgramStatus.ACTIVE: 'active'>)
>>> ProgramData(uuid="", title="", status="active")
ProgramData(uuid='', title='', status=<ProgramStatus.ACTIVE: 'active'>)

>>> ProgramData(uuid="", title="", status="invalid")
ValueError: 'invalid' is not a valid ProgramStatus
```

## Alternative 1

Alternately, it could be suggested to not use `converter=ProgramStatus`, though that can be potentially confusing as it's very strict about what it accepts:

```python
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE)
ProgramData(uuid='', title='', status=<ProgramStatus.ACTIVE: 'active'>)
>>> ProgramData(uuid="", title="", status=ProgramStatus.ACTIVE.value)
...
ValueError: ("'status' must be in <enum 'ProgramStatus'> (got 'active')", ...)
>>> ProgramData(uuid="", title="", status="active")
...
ValueError: ("'status' must be in <enum 'ProgramStatus'> (got 'active')", ...)
```

## Alternative 2

It could be nice to use Django's [`models.TextChoices`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#field-choices-enum-types) etc. instead of a stdlib python enum, because they support internationalized names and can be used directly with Django model `choices`. However, this goes against the recommendation in OEP-49 which says "This file should not import anything other than stdlib modules". (Though it clearly recommends `attrs` for the same file which is not a stdlib module.)

```python
from attrs import field, frozen, validators
from django.db import models
from django.utils.translation import gettext_lazy as _


class ProgramStatus(models.TextChoices):
    ACTIVE = "active", _("Active")
    RETIRED = "retired", _("Retired")

@frozen
class ProgramData:
    uuid: str = field(validator=validators.instance_of(str))
    title: str = field(validator=validators.instance_of(str))
    status: ProgramStatus = field(validator=validators.in_(ProgramStatus), converter=ProgramStatus)
```

When using the Django enums, `converter=ProgramStatus` is necessary as there is otherwise some auto-conversion going on which makes it accept an enum member value or plain string value, resulting in equivalent but inconsistently typed values being stored in the data structure. So that potential footgun could be a reason to prefer regular enums in situations where localization is not useful.

## Alternative 3

A nice option could be python stdlib's [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) but it's only in Python 3.11, so we can't use it yet. It avoids most of the data type complexities that I've found (see below) though it doesn't support internationalizing the labels like Django enums do.

### DIFF
--- a/oeps/best-practices/oep-0049-django-app-patterns.rst
+++ b/oeps/best-practices/oep-0049-django-app-patterns.rst
@@ -203,26 +203,17 @@ Example
 
   from enum import Enum
 
-  import attr
-
-  def ProgramStatus(Enum):
+  from attrs import field, frozen, validators
+  
+  class ProgramStatus(Enum):
       ACTIVE = "active"
       RETIRED = "retired"
 
-  @attr.attrs(frozen=True)
+  @frozen
   class ProgramData:
-      uuid: str = attr.attrib(
-          validator=[attr.validators.instance_of(str)]
-      )
-      title: str = attr.attrib(
-          validator=[attr.validators.instance_of(str)]
-      )
-      status: str = attr.attrib(
-          validator=[
-              attr.validators.instance_of(str),
-              attr.validators.in_(ProgramStatus)
-          ]
-      )
+      uuid: str = field(validator=validators.instance_of(str))
+      title: str = field(validator=validators.instance_of(str))
+      status: ProgramStatus = field(validator=validators.in_(ProgramStatus), converter=ProgramStatus)
 
 .. _rest_api:
 
@@ -289,6 +280,11 @@ At this time, there is no plan to enforce any of these guidelines. The vast majo
 
 Change History
 **************
+
+2023-04-26
+==========
+
+* In the section about data.py, fixed some broken examples of enum handling and converted to modern ``attrs`` syntax
 
 2023-04-20
 ==========


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

bradenmacdonald:
Actually I have found some problems using the standard python enums.

If you define a Django field like this:

```
class Program(models.Model):
    status = models.CharField(max_length=20, choices=[(s.value, s.value) for s in ProgramStatus])
```

and then use DRF to serialize it with:

```
class ProgramSerializer(serializers.ModelSerializer):
    class Meta:
        model = StagedContent
        fields = [..., 'status', ...]
```

you get a ` TypeError: Object of type ProgramStatus is not JSON serializable`

In addition, it becomes very tricky to write your ORM queries because it's easy to make them subtly wrong and you won't see any warning:

```
# Using standard python enum values in the model choices
In [4]: models.StagedContent.objects.all()[0].status
Out[4]: 'expired'
In [5]: models.StagedContent.objects.all()[0].status == data.StagedContentStatus.EXPIRED
Out[5]: False
In [6]: models.StagedContent.objects.filter(status=data.StagedContentStatus.EXPIRED).count()
Out[6]: 0
In [7]: models.StagedContent.objects.filter(status=data.StagedContentStatus.EXPIRED.value).count()
Out[7]: 7
```

Using `from django.db.models import TextChoices` avoids these problems: it works with DRF `ModelSerializer` automatically, and it has no footguns for the ORM:

```
# Using TextChoices
In [2]: models.StagedContent.objects.all()[0].status
Out[2]: 'expired'
In [3]: models.StagedContent.objects.all()[0].status == data.StagedContentStatus.EXPIRED
Out[3]: True
In [4]: models.StagedContent.objects.filter(status=data.StagedContentStatus.EXPIRED).count()
Out[4]: 7
In [5]: models.StagedContent.objects.filter(status=data.StagedContentStatus.EXPIRED.value).count()
Out[5]: 7
```

ormsbee (approved):

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/open-edx-proposals&entry.867055334=https://github.com/openedx/open-edx-proposals/pull/471&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2023-04-26+21:01&entry.1917517419=2023-04-27+14:16&entry.2133058324=Yes) so we can improve your experience in the future.

itsjeyd:
label: core contributor

PR #475, opened 2023-05-11
docs: [OEP-19] update status example in ADR template

robrap:
Minor update to the ADR template in OEP-19, providing a specific example for status. The new example provides a date format, and also removes the bullet that may be used if copying from the list of examples of status.

### DIFF
--- a/oeps/best-practices/oep-0019-bp-developer-documentation.rst
+++ b/oeps/best-practices/oep-0019-bp-developer-documentation.rst
@@ -9,7 +9,7 @@ OEP-19: Developer Documentation
    * - Title
      - Developer Documentation
    * - Last Modified
-     - 2022-06-22
+     - 2023-05-11
    * - Authors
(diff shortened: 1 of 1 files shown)


oeps/best-practices/oep-0019-bp-developer-documentation.rst
```diff
@@ -468,6 +470,11 @@ Next Steps
 Change History
 ==============
 
+2023-05-11
+----------
+
+* Minor update to the ADR template in OEP-19, providing a specific example for status. The new example provides a date format, and also removes the bullet that may be used if copying from the list of examples of status.
```
feanil, line 476: I see the addition but not the removal.  Did something not get checked in?

robrap: Some were setting status to `- **Accepted** *2023-05-11*`, rather than `**Accepted** *2023-05-11*`, because the hyphen was being copied from the list. This hyphen was not included in the new specific example.

feanil (approved):

PR #499, opened 2023-06-21
docs: OEP-14 update for DEPR transfer when archiving

robrap:
After archiving a repo, all issues become read-only. This is not ideal for DEPR tickets, so the DEPR issue should be transferred out before archiving. This step is now detailed in OEP-14.

Also fixes the directory that OEP-14 is in.


kdmccormick (approved):
👍🏻

robrap:
Oops. Forgot I had 3 instead of 2 commits. Oh well. :)

PR #501, opened 2023-06-22
docs: adr that outlines tools for enzyme migration

jesperhodge:
TLDR;

We propose that existing enzyme tests in openedx migrate over to https://github.com/openedx/react-unit-test-utils.
For new tests, we still recommend react-testing-library.

The reason is that it takes much too long and is too risky to migrate directly to react-testing-library, because we would have to rewrite every test file.


arbrandes:
Assigning myself as arbiter.

PR #508, opened 2023-08-11
OEP-11: move rejected alternatives into ADR pages

brian-smith-tcril:
This PR does not change OEP-11 or the decisions made within, it simply reorganizes them into ADRs. This will allow us to better reference individual decisions as we discuss future changes and write new ADRs.

For example: we will be able to reference `OEP-11/ADR-4` when discussing the use of TypeScript, or `OEP-11/ADR-2` when discussing Providence ([1](https://opencraft.com/better-state-management-for-react-using-providence/), [2](https://eye-of-providence.readthedocs.io/en/latest/index.html)) or [React Context](https://react.dev/learn/passing-data-deeply-with-context
).

### DIFF
--- a/oeps/best-practices/oep-0011/decisions/0002-redux.rst
+++ b/oeps/best-practices/oep-0011/decisions/0002-redux.rst
@@ -1,0 +1,31 @@
+Use Redux
+#########
+
+Status
+******
+
+Accepted
+
+Context
+*******
+
+The Open edX community needs a library to handle complex state management in React applications.
+
+Decision
+********
+
+For state management of complex
+client-side interactions, Redux must be used. This library was chosen
+because it sees strong use in the React community, but is also flexible
+enough to be used in situations where a hybrid React/Backbone architecture
+exists.
+
+Consequence
+***********
+
+Redux has been adopted by the Open edX community as the library to use to handle complex state management in React applications.
+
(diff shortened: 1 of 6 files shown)


oeps/best-practices/oep-0011/decisions/0002-redux.rst
```diff
@@ -0,0 +1,37 @@
+Use Redux
+#########
+
+Status
+******
+
+??? people are debating this one
```
feanil, line 7: I think marking this old decision as accepted is fine.  This also sets up the stage for an official discussion about dropping redux via a future ADR that supersedes this one.

oeps/best-practices/oep-0011/decisions/0004-js-ecma.rst
```diff
@@ -0,0 +1,50 @@
+Target the latest standardised JavaScript version
+#################################################
+
+Status
+******
+
+????? it seems like typescript is going to happen
```
feanil, line 7: Again we accept this decision and allow for a future ADR for typescript that can supersede it potentially.

brian-smith-tcril:
* replace some "should" language with "must"
* document decisions already made in correct tense

feanil (approved):

PR #510, opened 2023-08-17
OEP-66 RBAC (Role Based Access Control) - DRAFT

hsinkoff:


oeps/architectural-decisions/oep-0066-arch-role-based-access-control.rst
```diff
@@ -0,0 +1,88 @@
...
+The new system will initially be rolled out for 1 - 3 new roles, but once it is proven 
+effective, it will be rolled out for the roles that exist in course access roles and discussion roles. 
+At that time we will sunset the usage of course access roles and discussion roles.
+
+Consequences
+************
+
+Once completed this will remove one level of complexity from the overall authorization picture 
+by decreasing the systems involved by one, however during the interim it will add complexity by increasing 
+the number of involved systems and requiring all access checks to check for a role and a permission.
```
robrap, line 65: The term "system" is possibly confusing me. Will this check be within a service, or cross-service? Depending on the answer, we can discuss potential language. "System" made me think cross-service, but I don't know that that is the case.

oeps/architectural-decisions/oep-0066-arch-role-based-access-control.rst
```diff
@@ -0,0 +1,88 @@
...
+that grant access as they will continue to check for the same permissions no matter how many roles grant those permissions.
+
+Adding this new system will allow for eventual customized roles on an organization basis. 
+This system will continue to allow for assigning a course level role to a user at 
+an organization or instance level. 
+
+References
+**********
+
+.. image:: oep-0066/rbac_overview.png
```
robrap, line 79: This image contains much more detail that is not in this document. Is there another doc that contains this information? Do you want a sub-document or other sections? For accessibility purposes (which would probably help most readers), having this information in doc form would be useful. We could discuss options.

Also, this image contains so much. Even just breaking it into 2 images, rather than one that is so wide, would help give each more room.

oeps/architectural-decisions/oep-0066-arch-role-based-access-control.rst
```diff
@@ -0,0 +1,88 @@
...
+     - Draft
+   * - Type
+     - Architecture
+   * - Created
+     - 2023-08-16
+   * - Review Period
+     - TBD
+
+*******
+Context
```
robrap, line 30: I was surprised that there was no mention of the edx-rbac library. Is that being described below, but in other language? Is that not being discussed at all? It feels like an RBAC OEP should at least acknowledge this library and explain why it is or is not part of this OEP.

robrap (commented):
Some initial thoughts. Thank you @hsinkoff.

hsinkoff:
After speaking with @robrap and @feanil it has been decided that a best practice OEP for Authorization is needed and will include information on course level roles, but that an architecture OEP for this change is not needed. This information will be included in an ADR in the edx-platform repo and then documented as an authorization option in the best practice OEP. 

OEP-66 will be transformed into the Authorization best practice OEP. A new PR will be opened once a first draft of the authorization best practice OEP is ready.

PR #536, opened 2023-11-01
docs: Suggest linking to the developer how-to docs rather than inlining

timmc-edx:
See https://github.com/openedx/edx-cookiecutters/issues/396 -- we'd like to centralize this information rather than templating it to every repo's README.


sarina (approved):
Love it

PR #547, opened 2023-12-12
docs: Fix an error that slipped through CI.

feanil:
You can't have different number of columns in different rows of a
list-table.


feanil:
This got through CI because it turns out that the docs build was not a required check before.  I've updated the branch protection rules to require a passing docs build before merging PRs so it doesn't happen again and this should fix the error that I introduced when I merged https://github.com/openedx/open-edx-proposals/pull/544

bmtcril (approved):

PR #557, opened 2024-01-25
docs: add translation removal instructions to repo archive process

brian-smith-tcril:


oeps/processes/oep-0014-proc-archive-repos.rst
```diff
@@ -88,6 +88,14 @@ These steps should be followed for all repos within the Open edX organization (f
 
   - Move the repository to the openedx-unsupported organization
 
+  - Remove the repo from ``openedx-translations`` and Transifex
+    
+    - Remove the repo entry from the ``extract-translation-source-files`` workflow
```
sarina, line 93: To be clear, this workflow is in the ``openedx-translations`` repo? (and the following step same question)

oeps/processes/oep-0014-proc-archive-repos.rst
```diff
@@ -88,6 +88,14 @@ These steps should be followed for all repos within the Open edX organization (f
...
+    
+    - Remove the repo directory from the ``translations`` directory
+
+    - Remove the repo resource from the ``openedx-translations`` project on Transifex
```
sarina, line 97: Would it be better to say "ask a Transifex admin to..."? I don't think most people will be able to do this

sarina (commented):
Can you add a changelog entry too?

sarina:
@brian-smith-tcril - can we wrap this one up?

oeps/processes/oep-0014-proc-archive-repos.rst
sarina: I think an axim-engineering request. And maybe as part of this PR, we do an audit of admins on Transifex and remove people who aren't around or who don't need access and add a couple more Axim folx.

brian-smith-tcril:
@sarina I think my latest changes should address all the previous comments.

image

sarina (approved):

PR #564, opened 2024-02-22
docs: Add a detail to the CC nomination process.

feanil:
We want to make sure that maintainers see CC nominations for repos that
they maintain.  To that end, we should tag maintainers in the CC
nomination post.


sarina (approved):
This makes sense to me. @hurtstotouchfire is this sufficient for your concerns?

PR #580, opened 2024-04-24
docs: CSS best practices

jesperhodge:
This is a very early-stage rough draft for an OEP that intends for us to avoid future CSS conflicts.

The motivation comes from bugs I have encountered in the past, for example a CSS file in the header library breaking style in the Authoring MFE, which really shouldn't happen.

There are two solutions that make sense to me: Either a very rudimentary approach of just adopting very minimal rules for how we should scope and implement CSS - by just wrapping it in a unique selector for your component or repo;
or adopting CSS Modules.

I lean towards CSS Modules but this is a good start for a discussion.

We can also discuss rejected alternatives; the OEP is just a draft at this point so we can change it in any way the community wishes.

### DIFF
--- a/oeps/best-practices/oep-0068-frontend-best-practices.rst
+++ b/oeps/best-practices/oep-0068-frontend-best-practices.rst
@@ -1,0 +1,155 @@
+OEP-68: Frontend Best Practices
+###############################
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - Link to the doc in the following format::
+
+        :doc:`OEP-XXXX <oep-XXXX-YYYY-ZZZZ>`
+
+        * <XXXX is the next available OEP number>
+        * <YYYY is the abbreviated Type: proc | bp | arch>
+        * <ZZZZ is a brief (< 5 words) version of the title>
+
+   * - Title
+     - <OEP title>
+   * - Last Modified
+     - <date string, in YYYY-MM-DD format>
+   * - Authors
+     - <list of authors' real names and optionally, email addresses>
+   * - Arbiter
+     - <Arbiter's real name and email address>
+   * - Status
+     - <Draft | Under Review | Deferred | Provisional | Accepted | Rejected | Withdrawn | Final | Replaced>
+   * - Type
+     - <Architecture | Best Practice | Process>
+   * - Created
+     - <date created on, in YYYY-MM-DD format>
+   * - Review Period
+     - <start - target end dates for review>
+   * - Resolution
+     - <links to any discussions where the final status was decided>
+   * - References
+     - <links to any other relevant discussions or relevant related materials>
+
+Abstract
+********
+
+
+
+Motivation
+**********
+
+In the Openedx ecosystem, we have a complex Microfrontend architecture that follows
+architecture principles like modular boundaries. However, a well-working architecture
+like this also needs to rely on code not breaking these principles. Modular boundaries
+for example have little value if the code does not respect them; however,
+some of our code routinely does, which makes this OEP important.
+This OEP aims to establish some basic code quality standards that we can agree upon
+to guide us. The goal of course is for our applications and ecosystem to be functional,
+reliable, performant, easy to use, maintainable, secure, portable and scalable.
+
+Specification
+*************
+
+Ultimately, the goal is for our applications and ecosystem to be functional,
+reliable, performant, easy to use, maintainable, secure, portable and scalable.
+This is a good benchmark for quality standards proposed: if the proposed standard
+does not help us achieve these goals, it is likely not a good standard.
+
+CSS Quality Standards
+=====================
+
+CSS is famous for being laden with conflicts between different files and statements.
+All the most important modern UI and CSS frameworks attempt to deliver some solution
+to this problem, and standards like BEM that only exist for this reason
+have been around for a long time. Our own Paragon component library uses Bootstrap,
+which attempts to keep modular. However, in the past we didn't have standards
+for CSS code that was living outside of paragon. This has caused bugs and conflicts.
+
+An important cause of problems has been CSS that is not just in conflict with other CSS
+in the same Microfrontend but even with CSS in other Microfrontends, which produces
+extremely unpredictable bugs. For example, if you write some CSS in the Header or Footer
+components, that will override classes with the same names in Course-Authoring and won't
+show up in tests.
+
+This ADR will prescribe modular CSS but will not choose a certain technology or
+pattern to implement this. There are many solutions that are good.
+
+The following standards are proposed:
+- CSS in one Microfrontend must not be able to affect HTML / components in any other
+  Microfrontend.
+- CSS belonging to one React component or HTML file must not be able to affect any other
+  HTML or components except for children.
+  This means that CSS must be scoped to the component.
+- When you export a component for example from a component library, the CSS attached
+  must be scoped to that component and not affect anything outside of it.
+- When you import a component into your code, you should rely on the CSS it comes with,
+  and then you could override that CSS with CSS in your own code that's more specific, for
+  example. However, This code overriding the important component can only override it
+  in the place you are currently importing it. It must not override the imported
+  component's CSS in any other place that it's used.
+
+Rationale
+*********
+
+The rationale adds to the specification by describing the events or
+requirements that led to the proposal, what influenced the design, and why
+particular design decisions were made. The rationale could provide evidence
+of consensus within the community and discuss important objections or
+concerns raised during discussion. It could identify any related work,
+for example, how the feature is supported in other systems.
+
+Backward Compatibility
+**********************
+
+This statement identifies whether the proposed change is backward compatible.
+An OEP that introduces backward incompatibilities must describe the
+incompatibilities, with their severity and an explanation of how you propose to
+address these incompatibilities.
+
+Reference Implementation
+************************
+
+There are many ways to implement this, and we need to make sure we don't create
+conflicts between different ways to do it. That can mean that we make an openedx-wide
+decision on the technology used or, if there are diverse needs,
+that we use different technologies but ensure they are compatible. As stated,
+this OEP is agnostic to this choice.
+
+However, if no other solution is chosen, the following is the minimal requirement
+for any new CSS in openedx:
+
+- A CSS file cannot just define CSS for a selector like `.button`, because
+  this class name may be used in other places and you may overwrite other CSS in other
+  places by accident.
+- CSS must always be scoped to the MFE or repository it lives in, whether that's
+  an MFE, a component library, an HTML file, or anything else. This means that the
+  top-level HTML element of this repository needs to get a unique class or id;
+  for example the top-level element of the `frontend-component-header` repository
+  may simply have a class `.frontend-component-header` since repository names are unique,
+  and no other HTML element may have a class that's an openedx repository name.
+  Every CSS statement in this repo needs to include this class in its selector;
+  with SASS it's easy to just wrap everything in `.frontend-component-header {`.
+- Every component or HTML file that has CSS needs to have a top-level html element
+  with a classname that is unique within the repository (because of scoping rules above
+  it should be fine if the name re-occurs in a different repository). Then all
+  CSS that belongs to this component or HTML file needs to be wrapped in this top-level html
+  selector.
+
+Rejected Alternatives
+*********************
+
+This statement describes any alternative designs or implementations that were
+considered and rejected, and why they were not chosen.
+
+Change History
+**************
+
+YYYY-MM-DD
+==========
+
+* Document created
+* `Pull request #XXX <https://github.com/openedx/open-edx-proposals/pull/XXX>`_


bradenmacdonald:
# Utility-First CSS

I'm a huge fan of [Tailwind](https://tailwindcss.com/), and when you build an app with that approach, this problem never happens. We can't adopt Tailwind (and some people might not want to anyways), but we can focus on its style of **utility-first CSS** as a first step to avoiding this problem. We are already using this pattern a bit in the MFEs and in Paragon, but I personally think we should make more of an effort to use it: [Paragon CSS Utility Classes](https://paragon-openedx.netlify.app/foundations/css-utilities/)

In other words:

**Don't do**:

`Header.tsx`:
```tsx
import styles from './Header.scss';
const Header = () => {
    return <div className={styles.header}>...</div>
};
```

`Header.scss`:
```tsx
.header {
    margin-bottom: 1em;
    padding-top: 0.25rem;
    padding-bottom: 0.25rem;
}
```

**But instead**:
```tsx
const Header = () => {
    return <div className="mb-3 py-1">...</div>
};
```

Not only is the second approach much more concise and productive for a developer (no switching between source files), it completely avoids conflicts.

----------------

As much as you can, I would prefer to use utility classes as a first step to avoiding the issue, and then secondly use [CSS modules](https://github.com/css-modules/css-modules) for styles that cannot be expressed using the utility classes.

Unfortunately, the Paragon utility classes are currently quite limited. I'm used to Tailwind where you can express literally any styling using utility classes (and you get auto-completion of the utility class names in VS Code). The Paragon utility classes on the other hand, are still missing a lot of basic functionality (e.g. try to set a min-width of 50%, or give an element a minimum height on mobile only). So if others like the utility-first approach, I'd love to see an effort to expand the available utility classes in Paragon.

bradenmacdonald:
# CSS Modules 

[CSS modules](https://github.com/css-modules/css-modules) is a nice approach, but there's a challenge we face with implementing it: our .scss files usually need to import the paragon SCSS variables/mixins/functions, and doing that in a naive way results in a lot of duplicate SCSS.

For example, `learner-dashboard` uses some per-page CSS files as one would do when implementing CSS modules, and because each .scss file starts with `@import "@openedx/paragon/scss/core/core";`, you'll find that [there are 10+ separate copies of the full Paragon SCSS rules included in the production build](https://github.com/openedx/frontend-app-learner-dashboard/issues/325).

In other words:

**Don't do**:

`Header.tsx`:
```tsx
import styles from './Header.scss';
const Header = () => {
    return <>...<div className={styles.something}>...</div></>
};
```

`Header.scss`:
```scss
@import "@openedx/paragon/scss/core/core";
.something {
    box-shadow: $input-box-shadow;
    max-height: $card-image-vertical-max-height;
    border-radius: $alert-border-radius;
}
```

**But instead**:

`Header.tsx`: same as above.

`Header.scss`:
```scss
@import "@edx/brand/paragon/variables";
@import "@openedx/paragon/scss/core/_functions";
@import "@openedx/paragon/scss/core/_variables";
@import "~bootstrap/scss/mixins";
@import "@openedx/paragon/src/Form/_variables.scss"; // for $input-box-shadow.
@import "@openedx/paragon/src/Card/_variables.scss"; // for $card-image-vertical-max-height
@import "@openedx/paragon/src/Alert/_variables.scss"; // for $alert-border-radius
.something {
    box-shadow: $input-box-shadow;
    max-height: $card-image-vertical-max-height;
    border-radius: $alert-border-radius;
}
```

--------------------

If we want to support CSS modules (which I think is a nice approach), we need to either use the "one module per app" approach where each MFE has one giant .SCSS entry point for the whole app ([example from `course-authoring`](https://github.com/openedx/frontend-app-course-authoring/blob/a88066a2c5424601a76f5aacada0a9b17c568729/src/index.scss)) (ugh), or we need to implement support in Paragon so that you can have many separate .SCSS entry points and they each start with something simple like:

```scss
@use "@openedx/paragon/scss/paragon";  // Guaranteed not to emit any rules; this export contains only variables/functions/mixins
.something {
    box-shadow: paragon.$input-box-shadow;
    max-height: paragon.$card-image-vertical-max-height;
    border-radius: paragon.$alert-border-radius;
}
```
(here I'm also switching to the new `@use` syntax.)

jesperhodge:
These are great points. I love the idea of extending Paragon's utility classes. We typically try to use Paragon utility classes when possible, but there are many cases when we do introduce CSS files, and sometimes - if you already have a CSS file - it's just less complicated to do something in there. But that all comes from the utility classes lacking functionality, as you said.

It would be great to compose a list of missing features in Paragon's utility classes and then add classes that are as close to Tailwind as possible. The reason is that Tailwind has proven its ability to do pretty much everything that you can do with CSS in a simple way, so if we can add missing utility's that are closely modeled after tailwind we should have good confidence that we'll have all important features.

Partially because of the need to overwrite child component CSS for paragon components and such from the place you're using it in, I think we'll never reach quite 100% reliance on utility classes but we can get pretty close.

If we go that route, I'm not sure I would additionally recommend CSS Modules. It entails quite a bit of overhead and has the import problems you mentioned.

I think it might be sufficient for the transition period and for CSS that for some reason can't be replaced with utility classes to just put some pretty strict rules in place how your CSS must be scoped. That should avoid most bugs, and if we extend the utility classes and use them everywhere I think the amount of SCSS files we have will be tiny enough that it's unlikely to accidentally break the rules and create a bug.

What do you think @bradenmacdonald ?

bradenmacdonald:
> and then add classes that are as close to Tailwind as possible.

I like that in theory, but it may be a lot of work in practice.

Tailwind also has a lot of combinatorial classes like `dark:md:max-xl:hover:bg-yellow-100` (in dark mode, on devices from `md` to `xl` size but not smaller or larger, when hovered, use a light yellow background). I think you can use Sass `@for` loops to generate equivalents, but it's going to produce a ton of classes as the number of combinations get huge. Thus, it might not be practical to replicate entirely. But a simpler subset is probably doable. Either way, it would also require integrating something like PurgeCSS or PurifyCSS to remove unused utility classes from the final build, because there would be a lot. That wouldn't be difficult though.

One way to make this easier could be to find a way to do it incrementally - for example, when someone making an MFE encounters a "missing" utility class that exists in Tailwind but not Paragon, then they add it to `mfe/src/scss/new-utilities.scss`, and every few months, everything from that file is upstreamed into paragon proper.

> I think it might be sufficient for the transition period and for CSS that for some reason can't be replaced with utility classes to just put some pretty strict rules in place how your CSS must be scoped. 

That sounds fine to me. Maybe we can even find some existing linter that can help to enforce that to some extent.

adamstankiewicz:
> As much as you can, I would prefer to use utility classes as a first step to avoiding the issue, and then secondly use [CSS modules](https://github.com/css-modules/css-modules) for styles that cannot be expressed using the utility classes.

+1. I agree there's likely opportunities to extend the Paragon utility classes for common use cases beyond what currently exists (most of which are provided by Bootstrap 4), but share @bradenmacdonald's sentiment around being non-trivial / not practical. Also, generally, the CSS utility classes should be used over consuming the SCSS variables as well.

Worth noting, I don't believe the current, default Webpack configuration provided by `@openedx/frontend-build` supports CSS modules out-of-the-box due to the the `css-loader` options (i.e., `options.modules.compileType: 'icss'`) [[source](https://github.com/openedx/frontend-build/blob/250797d8db83880dd707fc214ea7ea3e32c34557/config/webpack.prod.config.js#L105-L107)]). This can be seen when trying to implement CSS modules with frontend-template-application. IIRC, the `compileType` was set related to [Paragon's `:export` of breakpoints widths](https://github.com/openedx/paragon/blob/9993fb3b938e50012b5aa220c1e75f4b8ff684c6/scss/core/_exports.module.scss#L15-L22) from SCSS (to be [consumed](https://github.com/openedx/paragon/blob/9993fb3b938e50012b5aa220c1e75f4b8ff684c6/src/utils/breakpoints.js#L1) by `breakpoints.js`).

I also agree with @jesperhodge's comment below around the overheard of migrating to CSS modules from a code implementation perspective (i.e., introducing `className={styles.something}` throughout the app):

> It entails quite a bit of overhead and has the import problems you mentioned.

> [CSS modules](https://github.com/css-modules/css-modules) is a nice approach, but there's a challenge we face with implementing it: our .scss files usually need to import the paragon SCSS variables/mixins/functions, and doing that in a naive way results in a lot of duplicate SCSS.

Regarding the import problems, agreed neither of the workarounds are ideal. I've typically relied on the single SCSS entry point aggregating component-specific SCSS files as that historically has personally felt like a better tradeoff than having multiple copies of Paragon.

That said, it is worth noting that with the (forthcoming) Paragon design tokens work (see [ADR](https://github.com/openedx/paragon/blob/master/docs/decisions/0019-scaling-styles-with-design-tokens.rst), Axim has a funded contribution to get it over the line), Paragon will be less reliant on SCSS variables/mixins in favor of relying on CSS variables instead. This change is to support runtime theming improves support for multi-tenancy and helps set up the (future) capability to have an official dark mode.

As a result, the current import problem with SCSS should be a non-issue in a design tokens / CSS variables world as module-specific CSS/SCSS files in MFEs could reference CSS variables without needing to import anything from Paragon first.

Also worth noting as part of the design tokens project is support for loading the Paragon CSS hosted by a CDN which would ensure a single (cached) copy of Paragon CSS is downloaded by users while navigating across multiple MFEs (kind of like the shared dependencies proposed by [OEP-65](https://docs.openedx.org/projects/openedx-proposals/en/latest/architectural-decisions/oep-0065-arch-frontend-composability.html)). Using the CDN approach of consuming Paragon's compiled CSS would be opt-in (i.e., MFEs could still choose to import the Paragon CSS from the NPM package). With OEP-65, the "host" app would likely provide the Paragon CSS variables, but "guest" apps could use CSS variables directly.

> Either way, it would also require integrating something like PurgeCSS or PurifyCSS to remove unused utility classes from the final build, because there would be a lot. That wouldn't be difficult though.

[inform] `@openedx/frontend-build` integrates with PurgeCSS today via the default Webpack configuration ([source](https://github.com/openedx/frontend-build/blob/250797d8db83880dd707fc214ea7ea3e32c34557/config/webpack.prod.config.js#L32-L36)), enabled via a `USE_PURGECSS` environment variable when running the build.

IIRC, a handful of MFEs tried to use PurgeCSS to improve performance for similar reasons in the past but ended up disabling it because it ended up being too brittle and often removed important CSS that shouldn't have been removed. Anecdotally, when my team recently focused on frontend performance in `frontend-app-learner-portal-enterprise` recently, we attempted to enable PurgeCSS and also found it to be too brittle.

Also, related to the aforementioned design tokens project, if MFEs do migrate to consuming Paragon from a CDN instead of importing CSS/SCSS from the `@openedx/paragon` NPM package like they do today, the Paragon CSS may not go through the Webpack build process where something like PurgeCSS wouldn't access its CSS to do its thing.

> > I think it might be sufficient for the transition period and for CSS that for some reason can't be replaced with utility classes to just put some pretty strict rules in place how your CSS must be scoped.
>
> That sounds fine to me. Maybe we can even find some existing linter that can help to enforce that to some extent.

Yeah, if there's any ways to help enforce this to some extent, that would be awesome. Unfortunately, I'm not sure the risk of (unintended) conflicts when introducing custom styles is evident to all engineers so making this more obvious at time of implementation would be great; we've had several bugs introduced by multiple engineers who weren't aware of the risk of introducing erroneous CSS conflicts. If nothing else, calling it out explicitly in documentation would be a good start.

That said, this issue of local vs. global scope is the primary advantage/reason for adopting CSS modules so engineers don't need to worry about potential conflicts (i.e., might warrant at least some investigation into how far away we would be from supporting CSS modules, what scope would there be in practice to migrate an MFE to use CSS modules for any of its custom SCSS, etc.).

jesperhodge:
We continued discussing this in our fedx meeting.

How about something like this?

### CSS and SASS Scoping Guidelines

**Utility Class Expansion:**
- Extend the Paragon utility classes by identifying missing functionalities compared to Tailwind. This includes creating a differential list of features supported by Tailwind versus what's currently supported by Bootstrap 4/Paragon, and incorporating the needed ones.

**SASS Declaration and Nesting:**
- Adhere to a structured way of declaring and nesting SASS to minimize scope conflicts, as detailed below:

1. **Scoped SASS Declarations:**
   - Always nest SASS statements within a unique class that specifies the component or microfrontend scope. 
   - **Example:**
     ```scss
     .frontend-app-course-authoring--course-outline { 
       .fancy-button { /* styles */ }
     }
     ```

2. **Unique Class Names:**
   - Ensure the class specifying the scope is always unique, especially when similar components exist within different parts of the application.
   - **Example:**
     ```scss
     .frontend-app-course-authoring--alert-component-unique { .fancy-button { /* styles */ } }
     .frontend-app-course-authoring--status-bar-alert-component { .fancy-button { /* styles */ } }
     ```

3. **File Organization:**
   - Place each SASS file next to its corresponding component to ensure locality and relevance of styles.

4. **Global and Local Styles Distinction:**
   - Avoid global style declarations. Use variables, design tokens, and utility classes instead.
   - Global styles must be placed in a specific 'global' file and should clearly state their scope.
   - **Example:**
     ```scss
     .frontend-app-course-authoring--global { 
       body { /* styles that affect the entire HTML document */ } 
     }
     ```

5. **Component-Specific Overrides:**
   - When necessary to override styles in a child component or one imported from a library, ensure these styles are scoped to the correct component file.
   - For instance, to change the `AlertComponent` background color on just the `CourseOutline` page:
     ```scss
     .frontend-app-course-authoring--course-outline {
       .frontend-app-course-authoring--alert-component { 
         .fancy-button { background-color: yellow; } 
       }
     }
     ```

6. **Alternative Nesting Pattern:**
   - Consider using a project-wide class that nests component-specific classes for broader scope definition.
   - **Example:**
     ```scss
     .frontend-app-course-authoring {
       .course-outline {
         .alert-component { .fancy-button { /* styles */ } }
       }
     }
     ```

**Enforcement:**
- These rules will be enforced through linting tools where feasible, incorporated into OEPs, and included in PR templates for routine checks during code reviews.

These guidelines are designed to maintain a clear, consistent approach to styling across all our microfrontends, ensuring that styles do not inadvertently affect unrelated parts of the application.

bradenmacdonald:
@jesperhodge Sounds good 👍🏻

sarina:
Hi @jesperhodge - any movement on this?

sarina:
Hi @jesperhodge - pinging again on this. Do you (or anyone else on this thread) intend to pick this up again? If I don't receive a response in the next two weeks, I'm going to close this PR.

PR #582, opened 2024-05-02
Add README for OEP-57 diagram sources

sarina:
We authored these in LucidChart and while that text is contained within the OEP, it feels even better to put that info directly where someone looking at the diagram source would want to know about it.

### DIFF
--- a/oeps/conf.py
+++ b/oeps/conf.py
@@ -96,7 +96,7 @@ language = 'en'
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', 'README.rst']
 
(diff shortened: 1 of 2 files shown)


sarina:
@feanil I want to add this as a README for this folder: https://github.com/openedx/open-edx-proposals/tree/master/oeps/processes/oep-0057 but I'm getting an error that the README isn't included in any toctree. Any advice?

feanil:
If you don't want it visible or served with the rest of the docs, you can add it to the [exclude_pattern](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-exclude_patterns) here: https://github.com/openedx/open-edx-proposals/blob/master/oeps/conf.py#L99

sarina:
@feanil thanks - I added `'README.rst'` to that file but I'm still getting the error

checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/open-edx-proposals/checkouts/582/oeps/processes/oep-0057/README.rst: WARNING: document isn't included in any toctree

sarina:
@feanil bump on this, so I can be done with it? 🙃

feanil (approved):

PR #586, opened 2024-05-17
Add a Needs Revision status for OEPs

sarina:
Over time, some OEPs may stay relevant - for example, they may have many
sections or core ideas that are still relevant to the project - while containing
many details that have become stale over time. When we are in agreement that the
OEP needs updating, we use this status to indicate to those browsing the OEPs
that this particular one requires some renewed attention.

When changing status to "Needs Revision", a row titled "Revision Ticket" should
be added to the preamble (directly under the status field) that directs to the
GitHub issue or draft pull request in the ``open-edx-proposals`` repository that
describes what about the OEP that needs revisioning.

An example is added with OEP-45. OEP-45 is directionally correct and we shouldn't mark it as _obsolete_ - it contains many great ideas. However it very much needs revision. We've captured the revision needs in https://github.com/openedx/open-edx-proposals/issues/587 and linked that issue to the OEP.

It is not always feasible that we find something wrong in an OEP and have the time right then and there to update the OEP. This status is a nod to that reality and a signpole to community members that they should tread carefully.

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -8,11 +8,11 @@ OEP-1: OEP Purpose and Guidelines
 +---------------+--------------------------------------------------------------+
 | Last-Modified | 2022-06-22                                                   |
 +---------------+--------------------------------------------------------------+
-| Authors       | Calen Pennington <cale@edx.org>,                             |
-|               | Joel Barciauskas <joel@edx.org>,                             |
-|               | Nimisha Asthagiri <nimisha@edx.org>,                         |
-|               | Feanil Patel <feanil@axim.org>,                              |
-|               | Sarina Canelake <sarina@axim.org>                            |
+| Authors       | - Calen Pennington <cale@edx.org>                            |
+|               | - Joel Barciauskas <joel@edx.org>                            |
+|               | - Nimisha Asthagiri <nimisha@edx.org>                        |
+|               | - Feanil Patel <feanil@axim.org>                             |
+|               | - Sarina Canelake <sarina@axim.org>                          |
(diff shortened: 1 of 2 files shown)


feanil (approved):

felipemontoya (approved):
Adding this new status sounds very reasonable

oeps/processes/oep-0001.rst
```diff
@@ -245,7 +245,7 @@ OEP Status
         "Under Review" -> { "Deferred" "Provisional" } [dir=both]
         "Under Review" ->  { "Accepted" "Rejected" "Withdrawn" }
         "Accepted" -> "Final"
```
pomegranited, line 247: Once an OEP is in "Needs Revision", where can it go from there? I haven't tried rendering it, but maybe this is enough? 

```suggestion
        "Accepted" -> "Final"
        "Needs Review" ->  "Final"
```

pomegranited (approved):
👍 with one comment -- thank you for this change @sarina !

oeps/processes/oep-0001.rst
sarina: Good point, and, it's tough because (a) Needs Review should just wrap around to Under Review (and all its possible states), and (b) GitHub isn't rendering the graph in preview and I'm having some issue with my local env at the moment.

sarina: Ha! I forgot we have a build :) I added like this:

image

How does that look? I don't _love_ it but otoh I love how it's autogenerated!

pomegranited: Oh yes, autogenerated FTW!

That looks good enough.

Issue #428, opened 2022-02-11
Enforce OEP-1 clause that all OEPs should be merged even if rejected/withdrawn


kdmccormick:
* OEP-1 says that you shouldn't close an OEP PR - you should instead merge it, utilizing the "withdrawn" or "rejected" statuses.
* We should clarify this with folks, and demonstrate it by merging in a few rejected/abandoned OEP PRs that were initially closed instead of merged.
* Should do openedx/open-edx-proposals#427  first though.

kdmccormick:
Unassigning myself as I have no current plans to do this.

sarina:
@kdmccormick  for this one, would you suggest I go through all unmerged PRs and reopen and merge them with an appropriate status?

kdmccormick:
That's what OEP-1 says to do. Personally, I think it's a little silly. If I wanted to know the rejected OEPs, I could just look at closed PRs.

sarina:
Ultimately I agree with you, @kdmccormick . I don't think it's worth the time to merge PRs that never materialized. I updated OEP-1 accordingly and discussion can continue on #601

sarina:
Completed by #601

PR #601, opened 2024-07-04
Remove "Rejected" OEP status, and indicate they should be left unmerged

sarina:
Reflect reality that rejected proposals are frequently left unmerged even though OEP-1 had previously called for it.

Remove the unused "Rejected" state and codify that you should just close the PR, and leave a note as to why.

Addresses #428

### DIFF
--- a/oep-templates/adr-based-template.rst
+++ b/oep-templates/adr-based-template.rst
@@ -28,7 +28,7 @@ OEP-XXXX: OEP Template
    * - Arbiter
      - <Arbiter's real name and email address>
    * - Status
-     - <Draft | Under Review | Deferred | Provisional | Accepted | Rejected | Final | Replaced>
+     - <Draft | Under Review | Deferred | Provisional | Accepted | Final | Replaced>
    * - Type
      - <Architecture | Best Practice | Process>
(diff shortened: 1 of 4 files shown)


kdmccormick (commented):
Thanks Sarina. I think this is an improvement over the current docs and I would not oppose merging it as-is.

That said, I feel that every instance in our official processes where we say "you may do this, or you can do that" without clear reason for doing one or the other is introducing a point of unnecessary ambiguity for our contributors. Writing an OEP is already challenging since it requires so much nuance and revision and convincing and difficult decision-making to reach a solid final document.... I think we can help OEP authors by making the mechanics of the process to be as clear-cut as possible. So, I would suggest to remove the "Rejected" state entirely, and instruct that rejected OEPs be closed instead of merged.

sarina:
@kdmccormick thanks for the feedback. I agree and added that change and reworked the PR title & description. Removing "Rejected" was a bit of a sweeping change and I'm not certain I did it all properly so I'd love your eyes for a second round.

kdmccormick (approved):
💯

pdpinch (approved):
👍

Issue #493, opened 2023-06-02
OEP-1 could be more clear about announcing


robrap:
The main section that notes how to announce an OEP is here: https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0001.html#step-3-review-with-arbiter. It doesn't have a clear process for announcing merges, which is especially important for OEP updates that didn't necessarily go through a wider review.

sarina:
@robrap what is the problem you're experiencing? Do you have a suggestion for how we could resolve the problem?

If you're looking for a push notification that something has merged, would a follow up post on the OEP thread in Discourse or another post in Slack that announces the merge suffice? If the latter, we could just set up an auto post connection between merges in this repo with a slack room.

robrap:
I personally started watching the repo, so I personally don't have an issue. That said, if decisions are being made/updated, I think it would be good for the community to see that in anyway that works for people. A slack room might work well.

All this being said, feel free to move forward however you think best. Thank you.

sarina:
I don't think it's unreasonable to add a section like what we have in DEPR about announcing - https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0021-proc-deprecation.html#removed - I'll assign this to myself to address in the coming weeks.

PR #602, opened 2024-07-04
docs: OEP-1: Add a section on announcing changes

sarina:
Previously, OEP-1 didn't have a clear process for announcing merges,
which is especially important for OEP updates that didn't necessarily
go through a wider review.

Addresses issue #493

### DIFF
--- a/oeps/processes/oep-0001.rst
+++ b/oeps/processes/oep-0001.rst
@@ -6,7 +6,7 @@ OEP-1: OEP Purpose and Guidelines
 +---------------+--------------------------------------------------------------+
 | Title         | OEP Purpose and Guidelines                                   |
 +---------------+--------------------------------------------------------------+
-| Last-Modified | 2024-07-08                                                   |
+| Last-Modified | 2024-07-25                                                   |
 +---------------+--------------------------------------------------------------+
 | Authors       | - Calen Pennington <cale@edx.org>                            |
 |               | - Joel Barciauskas <joel@edx.org>                            |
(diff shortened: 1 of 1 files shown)


oeps/processes/oep-0001.rst
```diff
@@ -223,6 +223,9 @@ The Open edX community is given the opportunity to comment on the OEP.
 The Arbiter serves to keep the discussion on track and to bring the review
 process to a final resolution.
 
+Finally, please follow the `After Merging`_ guidelines once your change has been
```
robrap, line 226: Why not just make a Step 4. Announcing Changes section as part of the workflow section?

robrap (commented):
Nice. Minor question about where the new section belongs.

Also, if you search for “Updating…”, there are a couple of other sections that could get a new step/sentence with a link.

oeps/processes/oep-0001.rst
sarina: I guess my initial thought was to follow the example of the [DEPR OEP](https://docs.openedx.org/projects/openedx-proposals/en/latest/processes/oep-0021-proc-deprecation.html#removed) and put it at the bottom, as a "final" step, but I'm not wedded to that idea. I think I'll move it.

oeps/processes/oep-0001.rst
```diff
@@ -415,6 +432,9 @@ OEP.
 * Additional diagrams or clarifying material (as long as the `Architecture
   Group`_ agrees that the substance of the OEP isn't changed).
 
+For the latter two points, please follow `Step 4. Announcing Changes`_ to inform
```
robrap, line 435: I’m not clear on exactly what “the latter two points” refers to. Instead, could we say “If any changes are made…”, or something like that?

robrap (commented):
Looks good. Thanks, one minor suggestion.

oeps/processes/oep-0001.rst
sarina, line 435: Yeah I meant, of the above four bullets, the latter two (so, points 3 & 4). So not if _any_ changes are made, but like, any substantive changes. How about,

```suggestion
When making changes larger than simple formatting or spelling corrections, please follow `Step 4. Announcing Changes`_ to inform
```

sarina:
@robrap - anything else you'd like to see here?

oeps/processes/oep-0001.rst
```diff
@@ -565,6 +585,11 @@ at the top of the list.
 Change History
 **************
 
+2024-07-25
+==========
+* Add a guidance on announcing changes (new as well as changed OEPs and ADRs)
```
robrap, line 590: Nit:
```suggestion
* Add guidance on announcing changes (new as well as changed OEPs and ADRs)
```

robrap (approved):
Looks great. Sorry I missed approval. Thanks.

Issue #573, opened 2024-03-19
Amend or revoke OEPs that were not implemented


kdmccormick:
Several OEPs were authored aspirationally, and then never implemented. Other were partially implemented, but have stalled or diverged from reality.

In order to establish OEPs as a trustworthy source of documentation, we should mark OEPs that never came to be as "Replaced" or "Obsolete" (whichever or is more appropriate). We should amend OEPs that are still relevant but mismatched with reality.

Browsing https://docs.openedx.org/projects/openedx-proposals/en/latest/, ones that jump out at me are:

- [x] OEP-56 Arch Advisory Process: Never implemented.
- [ ] [OEP-37 Dev Data: Partially implemented. Should evaluate whether to carry this forward or revoke it.](https://github.com/openedx/open-edx-proposals/issues/607)
- [x] OEP-06 Context-Scoped XBlock Fields: Never implemented, no plans to implement.
- [ ] https://github.com/openedx/open-edx-proposals/issues/587

Finally, various OEPs are in the state "Provisional". At this point we should transition them to "Accepted", if appropriate.

* See https://github.com/openedx/open-edx-proposals/issues/608

sarina:
Agreed on this. I'd be happy to help develop a plan with you - maybe we could do a working hour 1-1 in person and discuss an approach, maybe categorize a couple as well. I think that would kickstart it enough for me to run with finishing it out. If that works for you, let's aim for after we hire a data engineer.

sarina:
Closing this issue, as finer-grained new issues have been made for the remaining work.

PR #606, opened 2024-07-05
Mark OEP-56 and OEP-6 as Obsolete, as they were never implemented

sarina:
Partially addresses #573


davidjoy (approved):
💔 but ✅

bradenmacdonald (approved):
Makes sense. Thanks for cleaning this up.

PR #597, opened 2024-06-25
docs: remove withdrawn status and update OEP 1

saraburns1:
Closes #591

PR #609, opened 2024-07-05
docs: Remove remaining references to 'Withdrawn' state in OEP-1

sarina:
Follows up on #597 - finishes removing remaining references in OEP-1 and in templates


saraburns1 (approved):
sorry I missed those!!

kdmccormick (approved):

PR #618, opened 2024-07-30
build: tag correct owner for review on uprgade PRs

iamsobanjaved:
Currently, Arbi-BOM is being tagged on upgrade PRs but this repo is owned by Sarina as per the catalog-info.yml file, so updating github actions to tag Sarina instead.

### DIFF
--- a/.github/workflows/upgrade-python-requirements.yml
+++ b/.github/workflows/upgrade-python-requirements.yml
@@ -13,8 +13,8 @@ jobs:
    call-upgrade-python-requirements-workflow:
     with:
        branch: ${{ github.event.inputs.branch }}
-       team_reviewers: "arbi-bom"
-       email_address: arbi-bom@edx.org  
+       user_reviewers: "sarina"
+       email_address: sarina@axim.org  
        send_success_notification: false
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-python-requirements.yml
```diff
@@ -13,8 +13,8 @@ jobs:
    call-upgrade-python-requirements-workflow:
     with:
        branch: ${{ github.event.inputs.branch }}
-       team_reviewers: "arbi-bom"
-       email_address: arbi-bom@edx.org  
+       user_reviewers: "sarina"
+       email_address: sarina@axim.org
```
sarina, line 17: I don't think the email address is needed, it's not used here: https://github.com/openedx/docs.openedx.org/blob/main/.github/workflows/upgrade-python-requirements.yml

sarina (approved):

sarina:
Also is there a note in the documentation about how to become a maintainer that you need to update catalog-info.yaml AND this file? Is there possibly a way to leverage catalog-info.yaml so it's less manual/error-prone?

Issue #480, opened 2023-05-15
OEP-11 Frontend Technology Updates (Epic)


robrap:
Problem: [OEP-11: Front End Technology Standards](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0011-bp-FED-technology.html) is outdated and doesn't reflect current technology usages as well as directional changes we're making or need to make moving forward (for example, moving to Typescript as a dependency of upgrading to React 19). Thus, the OEP needs to be updated to reflect changes. Please add comments in this ticket if you will make a PR to update one of the following technology types, or if you have found additional sections that need updating.

## Changes Requested

- [x] Enzyme deprecated #518
- [ ] Typescript update
- [ ] Redux update
- [ ] Legacy server-side template updates

## Original comment

There is a long-standing, but undocumented stance that we don’t want new code in legacy server-side templates, and that all new code should be in an MFE. In theory, it would be great to require an ADR or comments explaining any exception to this rule.

It was acknowledged that we don't have this documented, and that [OEP-11: Front End Technology Standards](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0011-bp-FED-technology.html) would be a good place for it. The sections “1. Use React and Redux” and the intro to “12. Server-side content should be rendered with Django Templates” may be affected. However, ideally, this decision would be made much more explicit.

FYI: @sarina

sarina:
@arbrandes or @brian-smith-tcril  have you seen this proposed change? Do you have any thoughts?

arbrandes:
Hadn't seen it yet.  And now that I have, I have no objections.  On the contrary: this stance should definitely be documented!

robrap:
@arbrandes: FYI: Another OEP-11 issue may be coming regarding TypeScript, because it sounds like sentiment has been changing on that? I'm not sure if someone will create it, but this at least notes it for you.

arbrandes:
@robrap, thanks!  And for reference, Paragon [has already started](https://github.com/openedx/paragon/blob/master/docs/decisions/0018-typescript-support.md) to go down the typescript road.

robrap:
@arbrandes: Not sure if you want another issue to collect other OEP-11 updates, or if you want to use this issue? Others are providing thoughts in a private Slack, and I tried to get that moved to a ticket, but I don't see that happening. Here is an example:

> I definitely think we should revisit that OEP. There are a couple of points here I think would be interesting.
[...]

PR #675, opened 2025-02-03
docs: Update OEP-67 - use CSS variables, DEPR tech radar

sarina:
Addresses #480

Build link: https://docsopenedxorg--675.org.readthedocs.build/projects/openedx-proposals/en/675/best-practices/oep-0067-bp-tools-and-technology.html

### DIFF
--- a/oeps/best-practices/oep-0067-bp-tools-and-technology.rst
+++ b/oeps/best-practices/oep-0067-bp-tools-and-technology.rst
@@ -9,7 +9,7 @@ OEP-67: Standard Tools and Technologies
    * - Title
      - Tools and Technology Standards
    * - Last Modified
-     - 2025-01-30
+     - 2025-02-10
    * - Authors
      - Feanil Patel <feanil@axim.org>
    * - Arbiter
@@ -205,6 +205,14 @@ Frontend Technology Selection
 
 #. **CSS should be generated using SCSS**
 
+   .. note::
+
+      While there are no current plans to deprecate SCSS, the recommendation
+      as of February 2025 is to use CSS variables instead of SCSS variables
+      wherever possible now that browsers support them directly.
+
+      See :ref:`CSS variables should be used to support runtime theming <use-css-variables>`
+
    **Rationale**: Sass's SCSS syntax is an extension of CSS that adds power
    and elegance to the basic language. It makes the maintenance of large
    CSS files more manageable though the ability to use variables, mixins,
@@ -212,6 +220,14 @@ Frontend Technology Selection
    ability to override variables that define colors, fonts etc.
 
    You can find out more about Sass in the official `Sass documentation`_.
+
+#. **CSS variables should be used to support runtime theming**
+
+   .. _use-css-variables:
+
+   **Rationale**: SCSS variables are evaluated by the SASS preprocessor, and cannot be
+   changed afterwards. CSS variables are evaluated by the browser and
+   therefore can be changed at runtime, enabling dynamic theming support.
 
 #. **API calls should be made with the edX Frontend Auth Client or Axios**
 
@@ -330,11 +346,15 @@ Consequences
 
 * Future decisions for technology changes will require an ADR and an update to this OEP
 
-* The tech radar will serve as a snapshot of the current state of those decisions and to simplify
-  making changes, we will co-locate it with the supporting documents for this OEP.
-
 Change History
 **************
+
+2025-02-10
+==========
+
+* Add note encouraging CSS Variables instead of SCSS
+* Remove note about the Tech Radar as it is deprecated (`DEPR ticket <https://github.com/openedx/public-engineering/issues/264>`_)
+* `Pull request #675 <https://github.com/openedx/open-edx-proposals/pull/675>`_
 
 2025-01-30
 ==========


oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -207,6 +207,16 @@ Frontend Technology Selection
...
 
+   .. note::
+
+      While there are no current plans to deprecated SCSS, the recommendation
+      as of February 2025 is to use CSS variables instead of SCSS variables
+      wherever possible now that browsers support them directly.
+
+      Given that browsers support CSS nesting natively too, there is not so much
+      need for SCSS in general either, but we probably still need it since bootstrap
+      and thus Paragon are built using it.
```
brian-smith-tcril, line 218: I'm not sure we need this part. It feels to me like it takes focus away from the variables.

brian-smith-tcril (commented):
Couple little notes. One question: do we want to add anything about *why* using CSS variables is preferred over SCSS variables? I know this already mentions browsers supporting them directly, but the main benefit is that browsers evaluate them at runtime, whereas SCSS variables are compiled down into CSS that no longer knows about the variable.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
adamstankiewicz: I agree. It is also slightly implying that Paragon doesn't use CSS variables, whereas `v23` is largely the primary driver for migrating to CSS variables across the MFEs (despite still using SCSS for things other than variables).

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -207,6 +207,16 @@ Frontend Technology Selection
 
 #. **CSS should be generated using SCSS**
```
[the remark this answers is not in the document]
brian-smith-tcril: > @brian-smith-tcril Do you know if we've decided whether the local import of Paragon's styles directly in MFEs will remain the default option, with the CDN URLs, opt-in via configuration? Or is the intent to cutover to utilizing the CDN URLs by default within MFEs, where defining `PARAGON_THEME_URLS` becomes a requirement?

That depends on how Tutor wants to handle it. I'd be in favor of using URLs being the default (with the Tutor default being the Paragon styles coming from a Tutor hosted URL). That would make it so switching between local imports and URLs would be a rare occurrence.

bradenmacdonald:
> do we want to add anything about _why_ using CSS variables is preferred over SCSS variables? I know this already mentions browsers supporting them directly, but the main benefit is that browsers evaluate them at runtime, whereas SCSS variables are compiled down into CSS that no longer knows about the variable.

Just to continue this chain of thought: a big benefit is that it allows rapidly changing variables (especially for theming purposes), without having to re-build and re-deploy the compiled CSS. Much faster for testing out new things, and even allows "live" theming changes.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
PKulkoRaccoonGang: I would also probably drop this paragraph, since support for CSS nesting is not yet at a high enough level. According to [Can I use](https://caniuse.com/css-nesting), CSS nesting support is 91%, meaning 9% of users will not be able to fully use web applications.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -207,6 +207,16 @@ Frontend Technology Selection
...
+
+      While there are no current plans to deprecated SCSS, the recommendation
+      as of February 2025 is to use CSS variables instead of SCSS variables
+      wherever possible now that browsers support them directly.
+
+      Given that browsers support CSS nesting natively too, there is not so much
+      need for SCSS in general either, but we probably still need it since bootstrap
+      and thus Paragon are built using it.
+
    **Rationale**: Sass's SCSS syntax is an extension of CSS that adds power
```
PKulkoRaccoonGang, line 220: [proposal]: Given that it is preferable to use CSS variables where possible from February 2025, I think we should update this paragraph and focus on CSS variables instead of SCSS. [Mozilla MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) has pretty good documentation on using CSS custom properties.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -215,6 +223,14 @@ Frontend Technology Selection
 
    You can find out more about Sass in the official `Sass documentation`_.
 
+.. _use-css-variables:
```
sarina, line 226: Using this cross-reference causes the numbers to start over. I propose we don't use this cross-reference and just say in the above note to look at the next item.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -215,6 +223,14 @@ Frontend Technology Selection
...
+
+#. **CSS variables should be used to support runtime theming**
+
+   **Rationale**: SCSS variables are evaluated by the SASS preprocessor, and cannot be
+      changed afterwards. CSS variables are evaluated by the browser and
+      therefore can be changed at runtime, enabling dynamic theming support.
```
brian-smith-tcril, line 232: It looks like moving the ref prevents the number reset issue

```suggestion
#. **CSS variables should be used to support runtime theming**

   .. _use-css-variables:

   **Rationale**: SCSS variables are evaluated by the SASS preprocessor, and cannot be
      changed afterwards. CSS variables are evaluated by the browser and
      therefore can be changed at runtime, enabling dynamic theming support.
```

it's not an *ideal* solution as it causes the page to scroll just a little bit too far

image

but it feels "good enough" imo.

oeps/best-practices/oep-0067-bp-tools-and-technology.rst
```diff
@@ -207,6 +207,14 @@ Frontend Technology Selection
 
 #. **CSS should be generated using SCSS**
 
+   .. note::
+
+      While there are no current plans to deprecate SCSS, the recommendation
+      as of February 2025 is to use CSS variables instead of SCSS variables
+      wherever possible now that browsers support them directly.
+
+      See :ref:`CSS variablos should be used to support runtime theming <use-css-variables>`
```
brian-smith-tcril, line 216:
```suggestion
      See :ref:`CSS variables should be used to support runtime theming <use-css-variables>`
```

brian-smith-tcril (commented):
I'm definitely a fan of having CSS vars as a separate section!

Left a suggestion for a typo fix and a possible solution to the ref numbering issue

brian-smith-tcril (approved):

sarina:
@PKulkoRaccoonGang @adamstankiewicz - would you like to do another review?

PKulkoRaccoonGang (approved):
LGTM! Thanks

PR #690, opened 2025-03-10
docs: Relax CC requirement for maintaining.

feanil:
This can happpen when there is a team that maintains a repo and the team
is adding a new member.  That new member will not have gone through the
CC process yet but will be listed as a maintainer by virtue of being on
the team in the `catalog-info.yaml` file.  Allow for this to be
possible.  In practice we won't run into a lot of issues with this as
long as at least one person on the maintaining team can merge
maintenance work.


kdmccormick (changes requested):
I was thinking something more like this: https://github.com/openedx/open-edx-proposals/pull/691

If we approve and merge that ^, do you think this change ("on track to become or already be") would be necessary too?

feanil:
Yea, I like that better, let's go with that, I don't think we need both.

PR #691, opened 2025-03-10
docs(oep-55): Only one group member needs to meet maintainer prereqs

kdmccormick:


feanil (approved):

feanil:
With this change I don't think https://github.com/openedx/open-edx-proposals/pull/690 is necessary.

antoviaque:
@kdmccormick @feanil Does that change mean that we will have maintainers, with the full merge privileges associated with it, which don't match the requirements from core contributors? Or would the "individual within the group" who is a proper core contributor be the one who has the merge rights, and is the actual "official" maintainer?

kdmccormick:
>  Does that change mean that we will have maintainers, with the full merge privileges associated with it, which don't match the requirements from core contributors?

No, this edit does not make any changes to how write access is granted to any repo. That is (and will remain to be) granted via the CC program with nomination and voting. I imagine any change to that would require a proper OEP review process.

>  Or would the "individual within the group" who is a proper core contributor be the one who has the merge rights

Yes, this.

>  and is the actual "official" maintainer

No--the _group is the official maintainer_. In order to be named a maintainer, the group must contain at least one member with write access (via the CC program).

The OEP, unmodified, already contained the idea that a "maintainer" may be a group rather than a single person. This edit is based on that idea.

antoviaque:
@kdmccormick Ah, I see, that makes sense - thanks for the details!

robrap:
@kdmccormick: It is much simpler for a team to remain unblocked with 2 people, because both the reviewer and merger often requires write access. I don't think 2 needs to be a requirement, but you could add a recommendation around this.

(Apologies, but I originally added this comment in a wiki page and not on this PR.)

PR #734, opened 2025-09-07
docs: add OEP - Multi tenant mobile support

RawanMatar89:
This PR proposes OEP-68: Multi-Tenant Support in Open edX Mobile Apps.

The OEP introduces a mechanism for users to select their institution (tenant) at runtime, with each tenant providing its own branding, configuration, and API endpoints. This reduces duplication, enables a white-labeling model, and lowers long-term maintenance costs compared to maintaining separate app builds.

Key highlights:
	•	Tenant selector at login if multiple tenants exist.
	•	Tenant metadata stored in a JSON config (branding, URLs, auth).
	•	Separate database per tenant for isolation.
	•	Push notification manager updated to handle tenant context.
	•	Backwards compatible: single-tenant configs behave like current app.

Future enhancements may include fetching tenants dynamically from an API and supporting runtime tenant switching.

### DIFF
--- a/oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
+++ b/oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
@@ -1,9 +1,7 @@
-.. _OEP-68 Multi-Tenant Mobile Support:
-
 OEP-68: Multi-Tenant Support in Open edX Mobile Apps
 ####################################################
 
-.. list-table::
+ .. list-table::
    :widths: 25 75
 
    * - OEP
@@ -120,6 +118,59 @@ Push Notifications
 - Payloads include ``tenant_id``.  
 - Notifications routed to the correct tenant listener.  
 
+
+Tradeoffs
+=========
+We considered three approaches:
+
+**1. Single App + Bundled Config (proposed)**
+
+- **Pros**:
+  - Single codebase and release process, reducing duplication.  
+  - Works offline immediately since tenant definitions are bundled.  
+  - Reliable theming and branding without network dependency.  
+- **Cons**:
+  - Updating tenant metadata requires an app update (until remote fetch is added).  
+  - Larger app size if many assets are included.  
+- **Implications**:
+  - **Push notifications**: One device token registered with all tenants, routed by ``tenant_id``.  
+  - **Theming**: Strong, since assets/colors are local.  
+  - **Offline**: Strong support — tenants always available.  
+
+**2. Remote Configuration**  
+
+- **Pros**:
+  - Tenant metadata can change without an app update.  
+  - Centralized management of tenant definitions.  
+- **Cons**:
+  - Requires robust caching and fallbacks to work offline.  
+  - Higher runtime complexity and security considerations.  
+- **Implications**:
+  - **Push notifications**: Requires re-registration if tenant config changes.  
+  - **Theming**: More flexible, but must be cached to avoid broken UI offline.  
+  - **Offline**: Dependent on last successful fetch.  
+
+**3. Separate Builds (current state)**  
+
+- **Pros**:
+  - Full isolation per tenant (bundle IDs, push notifications, store presence).  
+  - Simple runtime logic.  
+- **Cons**:
+  - High maintenance and CI/CD overhead.  
+  - Risk of divergence across builds.  
+- **Implications**:
+  - **Push notifications**: Simplest — separate certs per build.  
+  - **Theming**: Static and guaranteed.  
+  - **Offline**: Guaranteed, but at the cost of scalability.  
+
+Impact on Single-Tenant Deployments
+===================================
+For single-tenant deployments, this proposal introduces minimal change:
+
+- If only one tenant is defined in the configuration, the tenant selector is skipped and the app behaves exactly as today.  
+- Branding, theming, and login flows remain static to that single tenant.  
+- No additional runtime overhead is introduced for single-tenant cases.  
+
 Alternatives
 ************
 - Separate app per tenant (high maintenance).  
@@ -144,3 +195,7 @@ Change History
 2025-03-18
 ==========
 * Initial draft created based on ADR.  
+
+2025-09-23
+==========
+* Added Tradeoffs and Single-Tenant Impact sections.  
--- a/oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
+++ b/oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
@@ -1,0 +1,146 @@
+.. _OEP-68 Multi-Tenant Mobile Support:
+
+OEP-68: Multi-Tenant Support in Open edX Mobile Apps
+####################################################
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :ref:`OEP-68 Multi-Tenant Mobile Support`
+   * - Title
+     - Multi-Tenant Support in Open edX Mobile Apps
+   * - Last Modified
+     - 2025-03-18
+   * - Authors
+     - 
+       * Rawan Matar <rawan@zeitlabs.com>
+       * Ivan Stepanok <ivan.stepanok@raccoongang.com>
+
+   * - Arbiter
+     - TBD
+   * - Status
+     - Draft
+   * - Type
+     - Architecture Decision
+   * - Created
+     - 2025-03-18
+   * - Review Period
+     - TBD
+   * - References
+     -
+       * `Multi-tenant PR <https://github.com/zeit-labs/openedx-app-ios-contrib/pull/10>`_
+
+.. contents::
+  :local:
+  :depth: 1
+
+Abstract
+********
+This OEP proposes adding **multi-tenant support** to the Open edX mobile applications. 
+Currently, the apps support only a single institution per build. 
+This proposal introduces a mechanism for users to select their institution (tenant) at runtime, 
+with each tenant providing its own branding, configuration, and API endpoints. 
+This reduces duplication and enables a white-labeling approach 
+within a single mobile app module.
+
+Motivation
+**********
+The Open edX mobile app is increasingly used by multiple institutions. 
+Releasing separate app builds for each tenant creates duplication, higher maintenance burden, 
+and inconsistent user experience. Multi-tenancy enables:
+
+* A single app build for multiple institutions.
+* Dynamic branding (logo, color scheme, names).
+* Institution-specific API and authentication endpoints.
+* Future extensibility to fetch tenant definitions dynamically.
+
+Current State
+=============
+- Mobile apps support only one tenant per build (YAML-based config).
+- Persistence layer uses a single Core Data/SQLite DB, risking data leakage across tenants.
+- Push notification manager doesn’t account for tenant context.
+- Branding and login flows are static per build.
+
+Decision
+********
+We propose supporting **multi-tenancy** in one app build by:
+
+* Providing a tenant selector screen at login.
+* Defining tenant metadata (branding, API URLs, auth URLs, etc.) in a JSON configuration file.
+* Allowing runtime tenant switching without reinstallation.
+* Maintaining isolated databases per tenant.
+* Enhancing push notifications with tenant context.
+
+Specification
+*************
+
+Tenant Selection
+================
+- On launch, show a tenant selector if multiple tenants are configured.  
+- If only one tenant exists, preserve the current flow.  
+
+Tenant Configuration
+====================
+Tenant metadata will be stored in a JSON config file, including:
+
+* ``name`` (tenant_id)
+* ``display_name``
+* ``logo_url``
+* ``primary_color``
+* ``api_base_url``
+* ``auth_url``
+
+Example::
+
+  {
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @RawanMatar89!


This repository is currently maintained by `@sarina`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

IvanStepanok:
Two concerns that may block Multi-Tenant adoption
1. Third-party authentication (Google/Microsoft/Facebook, SAML/OIDC).
Today the mobile apps ship with a single OAuth client per LMS installation, which makes third-party sign-in effectively app-scoped at build time. In a multi-tenant runtime, this constrains us to the basic username/password flow unless we add a per-tenant web-based login. Pragmatically: drive login via a WebView/deep link to each tenant’s LMS auth (OIDC/SAML) and return to the app via the standard OAuth redirect, so every tenant controls its own IdPs without baking them into the app.  ￼
2. Theming/branding.
If we intend to adopt Paragon’s design-token theming, we should land that first and then enable multi-tenant branding. Otherwise we risk building a bespoke theme manager now and re-implementing it to align with Paragon tokens later. I have an MVP that switches themes at runtime by loading CSS from the LMS for web content. Native surfaces can read the same tokens to keep parity.  ￼

https://stepanok.com/test-themes/theme1_light.css

https://github.com/user-attachments/assets/b18dc70e-a531-40bf-8f1a-1c8dc62f87dc

edschema:
Thanks @IvanStepanok and @RawanMatar89 
To me a WebAuth is totally acceptable; it's common with many apps at the moment. 

@IvanStepanok Design tokens seem very logical. Is there an interim solution to allow us to get this off the ground without design token theming that also allows us to avoid too much re-work? Would creating the config files/ adding multi-tenancy without design tokens add significantly more work than current builds necessitate?

IvanStepanok:
> @IvanStepanok Design tokens seem very logical. Is there an interim solution to allow us to get this off the ground without design token theming that also allows us to avoid too much re-work? Would creating the config files/ adding multi-tenancy without design tokens add significantly more work than current builds necessitate?

Sure, we can change only accent color by using our current Theme implementation. We already have an update function to change any colors. That’s really enough for the first implementation. Also, we can get the tenant logo from the web and save the URL for an PNG image in the config file.

oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
```diff
@@ -0,0 +1,146 @@
...
+Impact
+******
+- **Users**: Access multiple institutions in one app.  
+- **Institutions**: Branded experiences without custom builds.  
+- **Developers**: Reduced duplication and simpler long-term maintenance.  
+- **Backend**: No changes required, except for including ``tenant_id`` in notification payloads.  
+
+Future Work
+***********
+- Fetch tenant metadata dynamically from an API.
```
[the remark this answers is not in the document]
Kelketek: I also like the idea of using a QR code for setup. QR codes can include a protocol specifier for mobile apps that can allow the appropriate app to launch, fetch the configuration from the remote system, and add it to the phone. This one-time setup can be cached and I don't think it's asking too much for initial setup to have an internet connection.

The proposal as given allows a single operator to allow multiple tenants that they manage to go together in one app. However, I think having a single app that any instance with a relatively recent version can use is more useful. The one difficulty here is possible issues with API versioning that we'd need to be more careful about (some instances may update at different rates-- we'd need to decide how big of a window the app would support in terms of instance versions), but as long as we're deliberate about that, it should be doable.

If we require each provider to make their own copy of the app it's redundant work. It might increase adoption somewhat because apps would be per provider setting up their apps rather than doing it per client. But then you'd have to name them-- the OpenCraft Open edX Mobile app, the Edly Open edX Mobile App-- really that would be the only thread linking the various institutions that use a particular provider. There would also be multitenant instances, but from that perspective, it's still maintaining one app per organization that runs one of these multitenant instances.

No, the real value comes from having a central app any instance can use. Requiring pre-baked configurations means maintaining a central registry, which is going to mean setting one up and administrating it. Requiring a remote tenant listing ALSO requires making this registry, but just changes how we push it through.

It's better if there's no registry at all-- perhaps the app opens and lets you either manually input an LMS domain name, or alternatively scan a QR code you can get from your account page.

The data which has to be accessed and cached for tenant setup on the phone is minimal and is not expected to change much. How often do universities change their logos and colors? Perhaps occasionally, but even if your phone's colors and icon are out of date, it's not a usability issue, and it can be corrected whenever you get a better connection. Not a big deal.

Single-tenant builds of the app can just skip half of the work by asking the user to log in immediately, since they already know what endpoint is needed. But if we want to increase adoption, the instance operators shouldn't have to do anything special to take advantage of the mobile app. No custom builds, no submitting to the app store. Just using the prebuilt thing which already exists.

OmarIthawi:
> > @IvanStepanok Design tokens seem very logical. Is there an interim solution to allow us to get this off the ground without design token theming that also allows us to avoid too much re-work? Would creating the config files/ adding multi-tenancy without design tokens add significantly more work than current builds necessitate?
> 
> @edschema Sure, we can change only accent color by using our current Theme implementation. We already have an update function to change any colors. That’s really enough for the first implementation. Also, we can get the tenant logo from the web and save the URL for an PNG image in the config file.

Even if we go with design tokens, if we limit the options down to low number e.g. 7 customized colors or even lower it would be a really good start. 

I recommend getting the smallest amount of work as a corner stone to kickstart this effort. We need to set major goals and a roadmap with minor steps that _eventually_ reach those goals.

itsjeyd:
Hey @RawanMatar89, just checking in to see if you're still planning to continue working on this PR?

oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
```diff
@@ -0,0 +1,286 @@
...
+     - Multi-Tenant Support in Open edX Mobile Apps
+   * - Last Modified
+     - 2026-08-28
+   * - Authors
+     -
+       * Rawan Matar <rawan@zeitlabs.com>
+       * Ivan Stepanok <ivan.stepanok@raccoongang.com>
+
+   * - Arbiter
+     - TBD
```
OmarIthawi, line 19:
```suggestion
     - Marco Morales
```

oeps/architectural-decisions/oep-0068-multi-tenant-mobile-support.rst
```diff
@@ -0,0 +1,286 @@
...
+   * - Status
+     - Draft
+   * - Type
+     - Architecture Decision
+   * - Created
+     - 2025-03-18
+   * - Review Period
+     - TBD
```
OmarIthawi, line 27:
```suggestion
     - 2026-08-28 - 2026-09-22
```

@marcotuts let me know if that makes sense so we commit to it.

juancamilom:
Hi all — picking this back up now that the momentum around this is high. 

I want to make the connection explicit for anyone following both: the direction OEP-68 laid out — runtime tenant/instance selection, JSON-based per-instance config, backward-compatible single-instance behavior — is the same direction we've continued developing through the Mobile Product Working Group, under [[Multi-instance configuration for the Mobile Apps](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/6532923408/Proposal+Multi-instance+configuration+for+the+Mobile+Apps)].

That conversation moved into a new thread rather than continuing here, but it's a continuation of this work, not a separate one — @RawanMatar89 , @IvanStepanok , @edschema  and others have been part of it alongside @kdmccormick .

Where things stand now: we've landed on a curated-mode-first v1, deferring the fully open/no-registry model discussed here. Schema is close to what's sketched in this OEP. 
We are aligning on the additional enablers for things like Concurrent multi-tenant sessions, alternative login workflows (login before site selections and / or site selection before login).
As per @OmarIthawi 's suggestion, we are likely going to split  the product/UI/infrastructure layers into separate, more reviewable PRs

For process: we're aligning on this through the product proposal first — once that reaches consensus, we'll move to formalize it back into this OEP, rather than running both in parallel indefinitely.

OmarIthawi (commented):
@marcotuts

RawanMatar89:
Thanks for tying this together — appreciated. @juancamilom.

Good news on convergence: now that v1 is scoped to **curated mode only** (a static, publisher-curated JSON list, no live registry/routing backend — open mode explicitly descoped), that's materially the same shape already documented and shipped in OEP-68's Tenant Configuration — a JSON site list fetched/bundled, no backend router. 

To keep this moving without the two threads drifting further apart: since @OmarIthawi already proposed OEP → Infra → UI as the sequence, and a suggested review period now open (2026-08-28 – 2026-09-22, @marcotuts  as arbiter), our preference is to keep this PR the place where the architecture gets approved — rather than treating OEP formalization as something that waits behind the product proposal reaching its own consensus first.

**Separately:** 
1. the proposal uses "instance"/"site list", OEP-68 uses "tenant"/tenant config — since the schemas are converging, worth reconciling the vocabulary in OEP-68 directly so there's one term of record once this formalizes back here. Happy to fold that in on our side.

2. Now that OEP-68 documents concurrent sessions, there's a set of "what gets torn down on logout" behaviors that are currently inconsistent or unspecified. Rather than ask for one global policy discussion, proposing a concrete solution: a single app-level boolean config flag — `CLEAR_DATA_ON_VOLUNTARY_LOGOUT` (default `false`, matching today's behavior) — applied uniformly across all tenants, not per-tenant. When `true`, a voluntary (Settings) logout tears down everything a forced logout already does today.

This one flag would cover, consistently, across the app:
- **Downloaded course content** — currently cleared on forced logout only.
- **Push notification token** — currently unregistered on forced logout only; a tenant a user voluntarily logged out of can still be pushed to.
- **Calendar sync** (course due dates synced to the device calendar) — not addressed anywhere in the OEP yet.
- **Others of the same shape**: cached video/offline exam or quiz progress, queued analytics events — any other tenant-scoped data falls under the same flag rather than getting its own ad hoc rule.

This needs no per-tenant decision process — it's one config value, not a policy debate. If this shape works, I'll add it to OEP-68's top-level config and spell out the tie-in in Concurrent Sessions and Push Notifications.

RawanMatar89:
One thing I want to flag directly rather than let slide: the proposal lists by @IvanStepanok openedx-app-ios#670(https://github.com/openedx/openedx-app-ios/pull/670) and openedx-app-android#496(https://github.com/openedx/openedx-app-android/pull/496) as **already-open PRs**. Per @OmarIthawi's ask to follow OEP → Infra → UI, and since OEP-68 is the OEP this work is meant to formalize into, could we confirm those PRs are gated on OEP-68 reaching consensus rather than landing in parallel with it? I'd rather avoid infra merging ahead of the architecture decision it's supposed to implement.

PR #738, opened 2025-09-17
docs: Clarify that a CC sponsor is a "yes" vote

kdmccormick:
This was not explicitly outlined, but it would be very silly if that Sponsor did not count as a yes vote, because then organic candidates would have an illogical advantage over sponsored candidates.


sarina (approved):
This makes sense to me (and how I actually thought it worked...) but I'm going to post this in #open-edx-proposals in Slack and see if anyone has any dissent. I'll give a time window until next Monday and if there's no dissents I'll merge.

PR #773, opened 2026-03-18
docs: OEP-68 Learning Content Identifiers

kdmccormick:
## Abstract

Open edX code has four kinds of identifiers for learning content: **Integer Primary Keys**,
**Codes**, **OpaqueKeys**, and **UUIDs**. Choosing the wrong kind—or naming it ambiguously—can
cause subtle bugs, broken exports, and painful code reviews. This OEP explains what each kind
is, when to reach for it, and how to name it consistently.

The primary audience is **developers working on openedx-core, openedx-platform, and other
repositories that manage learning content**. Plugin developers, site administrators, and
nontechnical Open edX Core Contributors who work with learning content will also find these
concepts useful.

🌐 This is part of a broader "[Open edX Core v1.0](https://github.com/openedx/openedx-core/issues/353)" effort.

## Review 👀 

**Open**: Monday, March 23rd, 2026
**Closed**: ~Monday, April 13th, 2026~ Wednesday, April 15th, 2026 (beginning of day)

You can preview the rendered OEP here: https://open-edx-proposals--773.org.readthedocs.build/en/773/best-practices/oep-0068-bp-content-identifiers.html

## AI Note 🤖 

I taught Claude about the concepts in this OEP and then had it write a first draft. I then rewrote most of the content myself and corrected Claude's misunderstandings. I then told Claude "make it easier to read" and then reviewed and tweaked that.


oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,361 @@
...
+   * - Title
+     - Learning Content Identifiers
+   * - Last Modified
+     - 2026-03-28
+   * - Authors
+     - Kyle McCormick <kyle@axim.org>
+   * - Arbiter
+     - Sarina Canelake <sarina@axim.org>
+   * - Status
+     - Draft
```
kdmccormick:
```suggestion
     - Accepted
```

oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,361 @@
...
+come first.
+
+These conventions apply everywhere learning content is referenced: Python variables, JS
+variables, Django model field names, REST API arguments, event schema fields, admin
+interfaces, application logs, and so on.
+
+Summary
+=======
+
+.. list-table::
```
bmtcril: It would be nice if this table had guidance on database / model types for each identifier type as well. Some places where this comes up:

- `OpequeKeyField` vs `CharField`
- `UUIDField` vs `CharField` vs `BinaryField`
- How long to make the `CharField` for a code? Not that there's a single answer, but guidance might be useful

bmtcril (commented):
Looks good, just one comment

bradenmacdonald (approved):

Agrendalath (approved):

oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,361 @@
...
+
+Integer Primary Keys
+====================
+
+Every Open edX Django model should use ``django.db.models.BigAutoField`` as its primary key.
+This is an auto-incremented integer assigned by the database, meaningful only within that
+database.
+
+**When to use:** Primary keys are the default way to reference a database row within a single
+IDA on a single instance. Always use them for Django model foreign key relationships—as
```
ormsbee:
```suggestion
service on a single instance. Always use them for Django model foreign key relationships—as
```

Since IDA isn't standard industry terminology, I think we either need to explain it or just use "service".

ormsbee (approved):
This is great work—thank you so much for doing it. Minor questions and suggestions, but nothing that I would consider blocking.

bmtcril (approved):

rodmgwgu (approved):
Thanks for this proposal, this is great, I've been confused in the past with course_id's that are actually course keys, so this is really needed.

oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,397 @@
...
+integers, they can be indexed with almost no overhead, making lookups, joins, and constraint
+enforcement as fast as possible.
+The trade-off is that primary keys are meaningless outside the database that assigned them.
+
+**How to name:** On Django models, the primary key is ``id`` and foreign keys automatically
+get the ``_id`` suffix (e.g. ``collection_id``). Outside of model definitions—in variable
+names, REST APIs, event schemas, and so on—use the suffix ``_pk`` instead (e.g.
+``collection_pk``). When accessing the primary key on a django model, prefer ``.pk``, e.g.
+``collection.pk``.  "id" is an overloaded term that means many things; ``_pk`` leaves no
+doubt that the value is a Django model integer primary key.
```
kdmccormick: Unfortunately Braden's recent research shows that the name `pk` is in conflict with our desire to use typed model keys: https://github.com/openedx/openedx-core/pull/534#discussion_r3047918770 . He suggested changing this guidance to recommend `.id`.

I'm happy switching back to `.id` for accessing the PK on the model object and for queries. Anyway, that's what 95% of our code does, so we'd have way less code to update.

What I'll miss about `_pk` is that it disambiguates *non*-model fields very well:

```python
some_data = {
    "course_id": ...,  # ambiguous :(
    "course_key": ..., # clearly an opaque key
    "course_pk": ...,  # clearly a PK
}
```

Maybe we could offer a new convention for ambiguous situations, while still respecting the `_id` suffix?  Like, `_db_id`:
```python
some_data = {
    "course_id": ...,  # ambiguous :(
    "course_key": ..., # clearly an opaque key
    "course_db_id": ...,  # clearly a PK
}
```

Lmk your thoughts @ormsbee , @bradenmacdonald , et al

ormsbee: I'm okay with having vars named `course_id` an dealing with a bit of ambiguity, but I do strongly wish that we could keep `model_instance.pk` working somehow. It's a standard Django practice, and it's useful for 1:1 models like `Component`. I suppose we could modify `PublishableEntityMixin` so that the OneToOneField is named `id`, which would probably help. Right now, `my_component.id` would just error because the actual pk is at `my_component.publishable_entity_id`.

bradenmacdonald: In my PR, I added a typed accessor so that you can use `.id` and it returns `self.publishable_entity_id` with the correctly-typed `PK` type.

> I do strongly wish that we could keep model_instance.pk working somehow

Yeah. It works, but it has `Any` type, so it kind of bypasses all the advantages of typed primary keys if you use it.

------------

If we want the best of both worlds (fully typed + using `pk` terminology), I think the only feasible option (working around django-stubs limitations) is to use `modelname_pk` on each model, e.g. `container.container_pk`, `entity.entity_pk`, `other_entity.entity_pk`. This lets us be fairly consistent across the board (except if we need to access the ID field in django ORM queries/joins, e.g. `obj__entity_id` derived from `obj.entity.id`). But it's a lot more verbose.

Or we go with `id` which is simple, concise, can be made fully-typed, but is ambiguous in old code.

Or something novel like `db_id`, but I'd rather avoid that.

I think I'm personally leaning toward just `id`/`..._id` everywhere as it's by far the most consistent in new code, and we can try to work on bringing the old code in line over time.

kdmccormick: > I think I'm personally leaning toward just id/..._id everywhere as it's by far the most consistent in new code, and we can try to work on bringing the old code in line over time.

I'm supportive of this as well. At the very least, we'll be able to make openedx-core fully consistent, and most platform code will naturally be consistent. `course_id` (a CourseKey) will be the major offender, but we can work that out over time.

@ormsbee @bradenmacdonald , I can make this OEP update, any objections?

Btw @bradenmacdonald , I'd love if we could add guidance to the OEP about typed primary keys, but I'm fine merging this OEP first and then updating it after your openedx-core PR merges.

bradenmacdonald: Sounds good to me.

> Btw @bradenmacdonald , I'd love if we could add guidance to the OEP about typed primary keys, but I'm fine merging this OEP first and then updating it after your openedx-core PR merges.

I'd like to see if we can find any ways to improve this, as adding two `id_fields.py[i]` files to each app is slightly more overhead than I'd like. But if we only do it for "major" models, I think it's still a good approach.

bradenmacdonald: > Btw @bradenmacdonald , I'd love if we could add guidance to the OEP about typed primary keys, but I'm fine merging this OEP first and then updating it after your openedx-core PR merges.

I do have a recommendation on this now. Let me know what you think: https://github.com/openedx/openedx-core/pull/534#issuecomment-4210318770

kdmccormick:
@bmtcril @Agrendalath @rodmgwgu @ormsbee @bradenmacdonald , based on [the latest discussion](https://github.com/openedx/openedx-proposals/pull/773#discussion_r3052394999), I've pushed a commit to change it so that we'll use `_id` instead of `_pk` for database primary keys.

Please let me know before Wednesday morning (15 April) if you have concerns or if you'd like more time to re-review! cc @sarina .

oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,417 @@
...
+   * - Arbiter
+     - Sarina Canelake <sarina@axim.org>
+   * - Status
+     - Draft
+   * - Type
+     - Best Practice
+   * - Created
+     - 2026-03-28
+   * - Review Period
+     - 2026-03-28 - 2026-04-13
```
kdmccormick:
```suggestion
     - 2026-03-28 - 2026-04-15
```

oeps/best-practices/oep-0068-bp-content-identifiers.rst
```diff
@@ -0,0 +1,417 @@
...
+   * - Resolution
+     - TBD
```
kdmccormick:
```suggestion
   * - Resolution
     - `Original pull request`_

.. _Original pull request: https://github.com/openedx/openedx-proposals/pull/773
```

PR #786, opened 2026-04-30
docs: Add step to close open PRs & Issues when archiving a repo.

sarina:
This causes some data issues in the Contributions board, where `openedx-unsupported` PRs still show up in the Contributions board but we have no way to edit or close them anymore.

This is already a step in the Axim on-call playbook but is getting missed, so adding it to OEP-14 as well.

image


oeps/processes/oep-0014-proc-archive-repos.rst
```diff
@@ -183,6 +185,12 @@ Reasons rejected:
 Change History
 **************
 
+2026-04-30
+===========
+
+* Add step to close open PRs & Issues when archiving a repo.
+* `Pull request #785 <https://github.com/openedx/openedx-proposals/pull/785>`_
```
sarina:
```suggestion
* `Pull request #786 <https://github.com/openedx/openedx-proposals/pull/786>`_
```

feanil (approved):

PR #802, opened 2026-06-23
[FC-0118] docs: extend OEP-66 with a queryset-scoping pattern for list endpoints

Abdul-Muqadim-Arbisoft:
Adds a Django REST Framework subsection to OEP-66 covering the separation of three authorization concerns in list endpoints: endpoint access(permission_classes), record visibility (a scoping policy applied in get_queryset), and user-driven filtering (django-filter).
Record visibility delegates to openedx-authz via get_scopes_for_subject_and_permission — a single bulk scope lookup translated into a queryset filter, rather than a per-row enforce — and the section documents the list vs. single-object access paths and the modulestore fallback.

Supersedes the standalone ADR proposed in openedx/edx-platform#38354, moving the architectural pattern into OEP-66 per review feedback.

### DIFF
--- a/oeps/best-practices/oep-0066-bp-authorization.rst
+++ b/oeps/best-practices/oep-0066-bp-authorization.rst
@@ -11,7 +11,7 @@ OEP-66: User Authorization
    * - Title
      - User Authorization
    * - Last Modified
-     - 2025-12-18
+     - 2026-06-17
    * - Authors
      - Hilary Sinkoff (hsinkoff@2u.com), Jeremy Bowman (jbowman@edx.org), Maria F Magallanes (maria.magallanes@edunext.co)
    * - Arbiter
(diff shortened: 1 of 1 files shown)


oeps/best-practices/oep-0066-bp-authorization.rst
```diff
@@ -305,6 +305,192 @@ could be used by default for all view classes which don't override it.
...
+  ``get_queryset()``. Answers: "Which rows is this user allowed to see?" Rows the
+  user may not see are simply absent from the response; their absence is never a
+  ``403``.
+* **User-driven filtering**: a ``django-filter`` ``FilterSet``. Answers: "Of the
+  visible rows, which did the user ask for?" (for example, ``?org=edX``). This
+  narrows an already-authorized queryset and must never widen it.
+
+Keeping these separate makes access behavior consistent across endpoints,
+auditable in one place per model, and testable in isolation. The record
+visibility layer is the row-level security (RLS) concern; the rest of this
```
bradenmacdonald: This is a bit of a nit, but I don't think this is really "row-level security (RLS)"? 

As I understand, RLS is when you configure policies within the database system like MySQL or PostgreSQL, so that it enforces detailed row-level permissions automatically.

> tables can have row security policies that restrict, on a per-user basis, which rows can be returned by normal queries or inserted, updated, or deleted by data modification commands. This feature is also known as Row-Level Security [- Postgres docs](https://www.postgresql.org/docs/current/ddl-rowsecurity.html#:~:text=tables%20can%20have%20row%20security%20policies%20that%20restrict%2C%20on%20a%20per%2Duser%20basis%2C%20which%20rows%20can%20be%20returned%20by%20normal%20queries%20or%20inserted%2C%20updated%2C%20or%20deleted%20by%20data%20modification%20commands.%20This%20feature%20is%20also%20known%20as%20Row%2DLevel%20Security)

([Supabase also has great docs on RLS](https://supabase.com/docs/guides/database/postgres/row-level-security) since they encourage use of it in Supabase projects.)

Applying a "Record visibility" filter to `get_queryset()` is affecting rows in the query, but it's not what people normally think of as RLS, because it's only at the application layer and even there it can be easily bypassed (e.g., by querying `CourseOverview.objects.all()` directly instead of using the `get_queryset` queryset).

I would call this an "Authorization-filtered queryset" or "queryset scoping" but not RLS. RLS is a much higher level of security, but not something we are able to implement at the moment given that we don't configure such policies nor use different user accounts at the database level.

oeps/best-practices/oep-0066-bp-authorization.rst
```diff
@@ -305,6 +305,192 @@ could be used by default for all view classes which don't override it.
...
+                   course_ids.append(scope.course_id)
+
+           return queryset.filter(Q(org__in=org_keys) | Q(id__in=course_ids))
+
+
+   class CourseListView(RLSQuerysetMixin, ListAPIView):
+       permission_classes = [IsAuthenticated]   # endpoint access
+       rls_policy = CourseRLSPolicy()            # record visibility
+       filterset_class = CourseFilterSet         # user-driven filtering
+       queryset = CourseOverview.objects.all()
```
bradenmacdonald: This is very optional, but:

We're trying to encourage use of the new [`CourseRun` table](https://github.com/openedx/openedx-core/blob/33f133519e34846c519bdd4f5ddcba93cd99cc11/src/openedx_catalog/models/course_run.py#L30-L38) instead of `CourseOverview`. `CourseRun` uses integer primary keys and loads much less data on each query (it has way fewer columns), so it should be way more efficient for most use cases. It would be awesome if you could update the examples here to use `CourseRun` instead of `CourseOverview` to help raise awareness of `CourseRun` being the better option moving forward.

bradenmacdonald (commented):
I think this way of structuring things makes a lot of sense, and the explanation here is clear.

I just have some concerns about the term "RLS" and am wondering if we can update the example to use the newer DB table.

oeps/best-practices/oep-0066-bp-authorization.rst
Abdul-Muqadim-Arbisoft: Completely Agree, thank you for bringing this up ,i just switched the examples to CourseRun.

oeps/best-practices/oep-0066-bp-authorization.rst
Abdul-Muqadim-Arbisoft: You're right, thanks for the precise definition and references. This is application-level queryset scoping, not database-enforced RLS and as the Postgres/Supabase docs make clear.by the way I've renamed it throughout to "queryset scoping" / "authorization-filtered queryset" (ScopingPolicy, ScopedQuerysetMixin, scoping_policy) and added a note explicitly drawing that distinction, and ofcourse with a link to the PostgreSQL RLS docs so the difference is unambiguous.

Abdul-Muqadim-Arbisoft: Updated pr title and description as well

bradenmacdonald (approved):
Thanks for those tweaks. LGTM.

Faraz32123 (approved):

PR #805, opened 2026-07-05
[FC-0118] docs: add OEP-69 REST API Conventions

Abdul-Muqadim-Arbisoft:
Consolidate the FC-0118 REST API standardization ADRs (openedx-platform docs/decisions/0025-0037, plus the 0017 authentication pointer ADR) into a single Best Practice OEP, as requested during community review of the ADRs.

The OEP captures 13 conventions covering DRF serializers, ViewSets, endpoint consolidation, JWT authentication, permission classes, idempotent GETs, pagination, filtering/sorting, response shaping, error responses, API documentation, versioning, and the canonical MFE configuration endpoint.

Refs: https://github.com/openedx/openedx-platform/issues/38137
Discussion: https://discuss.openedx.org/t/request-for-community-review-on-adrs-for-standardizing-the-open-edx-platform-api-endpoints/18717

### DIFF
--- a/oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
+++ b/oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
@@ -1,0 +1,612 @@
+.. _OEP-69 Open edX REST API Conventions:
+
+OEP-69: Open edX REST API Conventions
+#####################################
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :ref:`OEP-69 <OEP-69 Open edX REST API Conventions>`
+   * - Title
+     - Open edX REST API Conventions
+   * - Last Modified
+     - 2026-07-06
+   * - Authors
+     - Abdul-Muqadim (abdul.muqadim@arbisoft.com), Taimoor Ahmed (taimoor.ahmed@arbisoft.com), Muhammad Faraz Maqsood (faraz.maqsood@arbisoft.com)
+   * - Arbiter
+     - Feanil Patel (feanil@axim.org)
+   * - Status
+     - Draft
+   * - Type
+     - Best Practice
+   * - Created
+     - 2026-07-05
+   * - Review Period
+     - <start - target end dates for review>
+   * - Resolution
+     - <link to the discussion where the final status is decided>
+   * - References
+     -
+       * `Request for community review on ADRs for standardizing the Open edX platform API endpoints <https://discuss.openedx.org/t/request-for-community-review-on-adrs-for-standardizing-the-open-edx-platform-api-endpoints/18717>`_
+       * `FC-0118 epic issue #38137: "Add ADRs for technical Recommendations on Open edX REST API Standards" <https://github.com/openedx/openedx-platform/issues/38137>`_
+       * `openedx-platform ADRs 0025–0037 (docs/decisions) <https://github.com/openedx/openedx-platform/commit/36a192d980faada26a423905f2a64c471928761b>`_
+       * :ref:`OEP-4 Application Authorization (Scopes)`
+       * :ref:`OEP-42 Authentication`
+       * :ref:`OEP-66 User Authorization`
+       * :ref:`OEP-65 Frontend Composability <OEP-65 Frontend Composibility>`
+       * :ref:`OEP-68 Content Identifiers <OEP-68 Learning Content Identifiers>`
+       * :ref:`OEP-21 Deprecation and Removal <OEP-21 DEPR>`
+
+Abstract
+********
+
+This OEP consolidates a set of Architecture Decision Records (ADRs) developed
+under Funded Contribution **FC-0118** into a single, authoritative set of REST
+API conventions for the Open edX platform. It establishes clear, shared patterns
+for how HTTP APIs should be built going forward, and describes how existing
+endpoints should be brought in line incrementally.
+
+The conventions align the platform more closely with the Django REST Framework
+(DRF) and with widely accepted REST and OpenAPI best practices, covering
+serializers, view structure, authentication, authorization, pagination,
+filtering and sorting, response shaping, error responses, documentation,
+versioning, and endpoint topology.
+
+The individual ADRs remain in ``openedx-platform`` (``docs/decisions/0025`` through
+``0037``) as the detailed, code-adjacent record of each decision. This OEP is the
+consolidated, cross-cutting reference (the "one big, clear REST API Conventions"
+document requested during community review) and is the canonical statement of the
+conventions for the Open edX community.
+
+Motivation
+**********
+
+This effort proposes to standardize Open edX APIs in a practical and incremental
+way. The intent is **not** to redesign the entire platform or introduce breaking
+changes unnecessarily, but to establish clear, shared patterns for how APIs should
+be built going forward, while gradually improving existing endpoints. By aligning
+more closely with DRF and with widely accepted REST and OpenAPI best practices, the
+platform can offer a more consistent and reliable API experience. In addition, it
+will lower integration costs for external systems, make APIs reusable, and provide
+a stronger foundation for future growth.
+
+Beyond the general improvement of the codebase and the developer experience (DX),
+one of the goals of this project is to facilitate the integration of Open edX with
+agentic generative AI: by standardizing API endpoints, enforcing strict schema
+consistency, and providing thorough API documentation, we greatly ease the creation
+of remote tools that can then be used by AI agents. In particular, it becomes much
+easier to expose the platform through machine-discoverable tools (for example, MCP
+servers and tools) that operate reliably against predictable, well-documented
+endpoints.
+
+The current state of the platform's REST surface works against these goals:
+
+* **Inconsistent construction.** Many endpoints manually build JSON responses
+  instead of using serializers, fragment a single resource across several views
+  instead of using ViewSets, and re-implement validation, permission, and logging
+  boilerplate per view.
+* **Inconsistent contracts.** Pagination, filtering, sorting, error shapes, and
+  authentication vary from app to app, forcing every consumer to write
+  endpoint-specific logic.
+* **Poor discoverability.** Documentation is incomplete and, where it exists, is
+  produced through an OpenAPI 2.0 toolchain that is no longer a viable long-term
+  home. This blocks SDK generation, contract testing, and reliable machine
+  consumption.
+* **Unclear evolution.** Versioned and unversioned paths coexist with no
+  consistent deprecation contract, so clients cannot tell which endpoint is
+  authoritative or how long it will last.
+
+These problems are individually addressable, but they were decided separately as a
+series of ADRs. Reviewers noted that it had become "a bit confusing as to what info
+is found in which ADR" and asked for the decisions to be consolidated into one clear
+"REST API Conventions" document. This OEP is that document.
+
+Specification
+*************
+
+The conventions below are the normative output of this OEP. Each convention names
+its source ADR in ``openedx-platform`` for the fully detailed rationale, code
+references, and rollout notes. The key words **MUST**, **MUST NOT**, **SHOULD**,
+and **MAY** are used in their conventional (RFC 2119) sense.
+
+All conventions share three cross-cutting principles:
+
+* **Incremental, backward-compatible migration.** New code follows these
+  conventions immediately. Existing endpoints are migrated over time. Where a
+  change would be backward-incompatible, it **MUST** be delivered as a new API version
+  and the old version deprecated per :ref:`OEP-21 Deprecation and Removal <OEP-21 DEPR>`.
+* **DRF as the integration surface.** These conventions standardize how endpoints
+  are expressed in DRF; they do not mandate a specific business-logic or policy
+  engine behind that surface.
+* **Machine-first contracts.** Predictable, documented, schema-covered contracts
+  are treated as a first-class requirement so that external systems and AI agents
+  can consume the platform reliably.
+
+Structure and framework
+=======================
+
+Convention 1: Use DRF serializers for all request and response handling
+-----------------------------------------------------------------------
+
+*Source: ADR 0025, Standardize Serializer Usage Across APIs.*
+
+All Open edX REST APIs **MUST** use DRF serializers for request and response handling
+rather than hand-constructing JSON.
+
+* API views **MUST** define explicit serializers for request and response
+  handling, and use them for both input validation and output formatting.
+* Manual JSON construction (``dict`` building, ``HttpResponse(json.dumps(...))``)
+  **MUST** be replaced with serializer-based responses.
+* Serializers **MUST** be documented with field ``help_text`` and validation
+  rules.
+* Input and output serializers are often different: the request body may accept
+  only a subset of fields, while the response includes computed or read-only
+  fields the caller cannot set. When they differ, define them separately and point
+  ``serializer_class`` at the **output** serializer (used by ``drf-spectacular``
+  for schema generation).
+
+.. code-block:: python
+
+   class CourseEnrollmentInputSerializer(serializers.Serializer):
+       """Validates the request body for enrollment creation."""
+       course_id = serializers.CharField(help_text="The course to enroll in.")
+       mode = serializers.CharField(default="audit", help_text="Enrollment mode.")
+
+   class CourseEnrollmentOutputSerializer(serializers.Serializer):
+       """Shapes the response: includes read-only fields not accepted on input."""
+       course_id = serializers.CharField(help_text="The enrolled course.")
+       mode = serializers.CharField(help_text="Active enrollment mode.")
+       is_active = serializers.BooleanField(help_text="Whether the enrollment is active.")
+       created = serializers.DateTimeField(help_text="Enrollment creation timestamp.")
+
+   class CourseEnrollmentViewSet(viewsets.ViewSet):
+       serializer_class = CourseEnrollmentOutputSerializer  # used by drf-spectacular
+
+       def create(self, request):
+           input_serializer = CourseEnrollmentInputSerializer(data=request.data)
+           input_serializer.is_valid(raise_exception=True)
+           enrollment = _enroll_user(request.user, **input_serializer.validated_data)
+           return Response(self.serializer_class(enrollment).data, status=201)
+
+Convention 2: Consolidate related endpoints into DRF ViewSets
+-------------------------------------------------------------
+
+*Source: ADR 0028, Migrating RESTful & Legacy Django API Endpoints to Standard
+DRF ViewSets.*
+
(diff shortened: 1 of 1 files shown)


oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,612 @@
...
+  handling, and use them for both input validation and output formatting.
+* Manual JSON construction (``dict`` building, ``HttpResponse(json.dumps(...))``)
+  **MUST** be replaced with serializer-based responses.
+* Serializers **MUST** be documented with field ``help_text`` and validation
+  rules.
+* Input and output serializers are often different: the request body may accept
+  only a subset of fields, while the response includes computed or read-only
+  fields the caller cannot set. When they differ, define them separately and point
+  ``serializer_class`` at the **output** serializer (used by ``drf-spectacular``
+  for schema generation).
```
bradenmacdonald, line 147: But don't we also need the input serializer to be included in the schema too? From the example below, it looks like the API tooling would only be aware of the output schema, so half of the endpoint's spec is missing.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,612 @@
...
+  ``action``/``mode`` field (or HTTP verbs where REST conventions apply cleanly).
+* Shared validation, audit logging, response shaping, and permission enforcement
+  **MUST** move into a common service layer or mixin. Consolidation removes
+  duplicated boilerplate; it **MUST NOT** flatten the authorization model: the view
+  performs a coarse access check and each operation handler enforces its own
+  specific permission.
+* The unified endpoint **MUST** document its enumerated ``action``/``mode`` values
+  and their request/response shapes in OpenAPI.
+* Legacy URLs **MUST** be preserved as deprecated aliases (emitting a
+  ``Deprecation`` header) for a defined transition window.
```
bradenmacdonald, line 213: This whole section is a bit vague and confusing to me. At first I thought it was saying that if you have two different HTTP verbs for a resource, like `POST /users/5/profile/` and `GET /users/5/profile/`, that you have to implement them with a single function. ("endpoints that operate on the same resource and differ only by the operation applied... **MUST** be consolidated into a single parameterised view"). That would be messy!

At the very least, this needs some examples. But from looking at [the ADR](https://github.com/openedx/openedx-platform/pull/38262), I see it only lists one known example: "Relevance in edx-platform: Certificate views".

If this is just about cleaning up that one poorly-organized endpoint, I don't think it rises to the level of needing to be included as a section of this OEP. If you do think it's important enough to leave in, it needs to be clarified and given examples.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,612 @@
...
+The reference implementation is the ongoing FC-0118 work in ``openedx-platform``,
+tracked under the umbrella issue `#38137
+<https://github.com/openedx/openedx-platform/issues/38137>`_ and recorded as ADRs
+``docs/decisions/0025`` through ``0037`` (plus the pointer ADR
+``openedx/core/djangoapps/oauth_dispatch/docs/decisions/0017``). Priority
+migration targets identified by the ADRs include the Enrollment API ViewSet
+consolidation, the ``drf-spectacular`` rollout, the ``BearerAuthentication``
+deprecation via the ``view_auth_classes`` decorator, and the canonical
+front-end configuration endpoint. This OEP may move to "Final" once the conventions
+have representative, merged implementations across the highest-impact endpoints.
```
bradenmacdonald, line 574: Shouldn't the reference implementation be in a single library, and outside of `openedx-platform`?

e.g. `openedx_rest_lib` could provide `DefaultPagination`, `permission_classes`, `django-filter`, `drf-spectacular`, Error-response standardization, CI tooling, and any of the other requirements specified in this OEP. So then following this OEP would be mostly about strictly using `openedx_rest_lib`.

I would personally make `openedx_rest_lib` re-export all of the DRF API as well, so that consumers can enjoy a much cleaner import story:

Today 🤢

```python
from drf_spectacular.utils import (
    OpenApiParameter,
    OpenApiRequest,
    OpenApiResponse,
    extend_schema,
)
from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
from edx_rest_framework_extensions.paginators import DefaultPagination
from rest_framework import permissions, status, viewsets
from rest_framework.decorators import action
from rest_framework.exceptions import NotFound, ValidationError
from rest_framework.generics import ListAPIView
from rest_framework.response import Response
from rest_framework.views import APIView
from openedx.core.lib.api.mixins import StandardizedErrorMixin
from openedx.core.lib.api.permissions import ApiKeyHeaderPermissionIsAuthenticated
```

Tomorrow? 🌈

```python
from openedx_rest_lib import (
    OpenApiParameter,
    OpenApiRequest,
    OpenApiResponse,
    extend_schema,
    StandardViewSet,  # Has DefaultPagination, JwtAuthentication, StandardizedErrorMixin, etc.
    ListAPIView,  # Has DefaultPagination, JwtAuthentication, StandardizedErrorMixin, etc.
    APIView,  # Has DefaultPagination, JwtAuthentication, StandardizedErrorMixin, etc.
    rest_exceptions,  # DRF NotFound, ValidationError, etc. (standardized)
    permissions,
    Response,
)
```

bradenmacdonald (commented):
Thanks, this is great! I left some comments inline.

One major omission (?) though seems to be URLs.

Our REST API URLs today follow no consistent patterns whatsoever:

```
/transcripts/upload
/xblock/:id/handler/:handlerName/
/xblock/container/:id/
/organizations/
/api/toggles/
/event/
/api/user/v1/accounts
/api/v1/course_runs/
/course_team/:course_id/
/course_rerun/:course_id/
/videos/:course_id/:video_id/
/api/learning_sequences/...
```

I think this OEP should clearly define URL expectations like:

* Whether we should use singular nouns (`/user/:x`) or plural nouns (`/users/:x`)
* That REST API URLs should always start with `/api/`
* That new CMS and LMS URLs should not conflict with each other so they can [potentially be merged](https://discuss.openedx.org/t/proposal-combined-headless-lms-cms/18880) without needing major changes to API clients other than domain name.
* Whether APIs are organized hierarchically using logical components `/api/course/:course_key/teams/:team_id/` or the URLs should strictly map to the underlying organization of the codebase `/api/teams/v1/course/:course_id/:team_id` or some other organizational scheme ?
* etc.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
Abdul-Muqadim-Arbisoft: Agree, serializer_class only drives the response schema, so as written the request body is missing/misinferred. This is actually already handled correctly in the merged pilots, e.g. the Course Detail v3 view uses @extend_schema(request=OpenApiRequest(request=CourseDetailsSerializer), responses={200: OpenApiResponse(...)}). The OEP example just under-specified it. I'll fix the example to declare both explicitly and add a sentence: when request and response differ, both MUST be registered via @extend_schema(request=..., responses=...); serializer_class is only the fallback.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
Abdul-Muqadim-Arbisoft: Agree this is bit unclear, and you've identified the real issue: it's the thinnest section. It's specifically about RPC-style action-named endpoints (the certificate enable_/start_/regenerate cluster → one certificate_tasks resource with a mode), not about how HTTP verbs on a resource are implemented (that's the ViewSet convention). Also worth noting: unlike the other conventions, this one has no merged implementation yet, certificates remain the only example. Given that, I'll (a) reframe it narrowly as "avoid proliferating action-scoped URLs," (b) add the concrete before/after, (c) explicitly clarify it does not mean collapsing verbs into one handler, and (d) soften MUST→SHOULD.  your call.

Abdul-Muqadim-Arbisoft:
> Thanks, this is great! I left some comments inline.
> 
> One major omission (?) though seems to be URLs.
> 
> Our REST API URLs today follow no consistent patterns whatsoever:
> 
> ```
> /transcripts/upload
> /xblock/:id/handler/:handlerName/
> /xblock/container/:id/
> /organizations/
> /api/toggles/
> /event/
> /api/user/v1/accounts
> /api/v1/course_runs/
> /course_team/:course_id/
> /course_rerun/:course_id/
> /videos/:course_id/:video_id/
> /api/learning_sequences/...
> ```
> 
> I think this OEP should clearly define URL expectations like:
> 
> * Whether we should use singular nouns (`/user/:x`) or plural nouns (`/users/:x`)
> * That REST API URLs should always start with `/api/`
> * That new CMS and LMS URLs should not conflict with each other so they can [potentially be merged](https://discuss.openedx.org/t/proposal-combined-headless-lms-cms/18880) without needing major changes to API clients other than domain name.
> * Whether APIs are organized hierarchically using logical components `/api/course/:course_key/teams/:team_id/` or the URLs should strictly map to the underlying organization of the codebase `/api/teams/v1/course/:course_id/:team_id` or some other organizational scheme ?
> * etc.

This is a great point. The OEP already fixes URL versioning (Convention 12 mandates /api/<area>/vN/), but you're right that it says nothing about URL naming and organization and none of the source ADRs (0025–0037) did either. So this is worth pinning down. It would cover the items you listed:

- REST URLs always under /api/
- singular vs plural nouns (/user/:x vs /users/:x)
- CMS and LMS not defining conflicting paths for the same logical resource, so they can be merged behind one domain without breaking clients
- hierarchical/logical organization (/api/courses/:course_key/teams/:team_id/) vs. URLs mapping to the codebase layout (/api/teams/v1/course/:course_id/:team_id), or another scheme
Given how the rest of this work was done (ADR first, then consolidated here), I'd suggest capturing this as a new ADR and validating it against the FC-0118 endpoints as we standardize them, then summarizing the outcome back into OEP-69 so it's still defined here. The one catch is scope: FC-0118 is almost at close.

@feanil  could we get your consensus on whether it makes sense to add one more ADR (URL conventions) under FC-0118 and implement it against the selected endpoints now, given we're near close, or capture it as a documented follow-up to pick up separately?

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
Abdul-Muqadim-Arbisoft: Great idea. The pieces already exist but are split: DefaultPagination/JwtAuthentication are in edx-drf-extensions (a library, reusable), but StandardizedErrorMixin + the ADR-0029 handler live in openedx.core.lib.api inside the platform, and each pilot app defines its own permission classes (e.g. HasCourseAuthorAccess in the xblock PR). So plugins and other IDAs can't reuse the standardized bits today, exactly the problem a shared library solves. edx-drf-extensions looks like the natural home since it already ships pagination + JWT.

One caveat from the current implementation: the ADR-0029 handler currently imports ignored_error_exception_handler from openedx.core.lib.request_utils, so it's coupled to platform internals, extracting it means breaking that coupling first, not a straight move. On the re-export idea: my only caution is keep-in-sync cost, so I'd lean toward a curated set + a StandardViewSet/StandardAPIView base (pre-wiring pagination, JWT, and the error mixin) rather than a 1:1 re-export of all of DRF.

Since this is the same shape as the URL point, a design decision best captured as its own ADR + implementation rather than folded into this OEP, and FC-0118 is almost at close: @feanil , could we get your consensus on whether to add another ADR (extract a reusable REST library into edx-drf-extensions) under FC-0118 and start migrating the pilot endpoints onto it now, or track it as a documented follow-up? We'll then update OEP-69's Reference Implementation section to point at the library as the primary deliverable.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,612 @@
...
+endpoints should be brought in line incrementally.
+
+The conventions align the platform more closely with the Django REST Framework
+(DRF) and with widely accepted REST and OpenAPI best practices, covering
+serializers, view structure, authentication, authorization, pagination,
+filtering and sorting, response shaping, error responses, documentation,
+versioning, and endpoint topology.
+
+The individual ADRs remain in ``openedx-platform`` (``docs/decisions/0025`` through
+``0037``) as the detailed, code-adjacent record of each decision. This OEP is the
```
feanil, line 57: Make these links to the folder of ADRs in `openedx-platform`

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
feanil: Just making a note that the text and code here has not been updated yet.  It sounds like we would use something like `get_serializer_class` to differentiate the serializer class but then we would need to annotate the api schema to make sure it can see the right info.  This is definitely worth noting and providing a thorough example.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
feanil: I think this section definitely merits some example code.  As well as more guidance on when we should do this. I can't think of times when we would want to have the same resource have separate views such that actions would be split across views. Perhaps in the case where we separated the detail view from the list view because the logic is significantly different for those two.  But I think even then it could all be enclosed.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
feanil: I agree with Braden here, that a lot of the re-usable bits should live in edx-drf-extensions if possible.  I'm not as sure about re-exporting the DRF Base classes, I think that can lead to more complexity when reading if you are new to the codebase.

feanil (changes requested):
All of the sections should link to their respective ADRs directly.

I also think it might be worth it to have code examples for these here, but could be convinced that having them in the linked ADRs is sufficient.  Though in that case, we should explicitly link to the code examples sections.

Abdul-Muqadim-Arbisoft:
> All of the sections should link to their respective ADRs directly.
> 
> I also think it might be worth it to have code examples for these here, but could be convinced that having them in the linked ADRs is sufficient. Though in that case, we should explicitly link to the code examples sections.

We already have the respective code examples in the ADRs so i think linking them over here would suffice.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
Abdul-Muqadim-Arbisoft: As discussed, since I couldn't find more examples in the code, I think I'll add a few illustrative ones instead to make this section stronger.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
Abdul-Muqadim-Arbisoft: Hi @feanil @bradenmacdonald 
The artifacts we're planning to move to edx-drf-extensions are the following:

1.  StandardizedErrorMixin | mixins.py
2.  Envelope formatters + Conflict | errors.py
3.  project() + MinimalViewMixin | shaping.py
4.  IterablePaginationMixin + paginate_manually() | paginators.py (existing)
5.  Opaque-key lookup regex constants | routers.py
6.  assert_error_envelope() | testing.py
7.  Error-type URI catalog + extension helper | errors.py
8.  ErrorResponseSerializer (plain DRF) | errors.py
9.  Exception-handler dependency inversion | errors.py + settings

These were chosen based on the feasibility of moving them without introducing much coupling into already-implemented pieces. Let mw know if you think we can move any other then these.

Abdul-Muqadim-Arbisoft:
> Thanks, this is great! I left some comments inline.
> 
> One major omission (?) though seems to be URLs.
> 
> Our REST API URLs today follow no consistent patterns whatsoever:
> 
> ```
> /transcripts/upload
> /xblock/:id/handler/:handlerName/
> /xblock/container/:id/
> /organizations/
> /api/toggles/
> /event/
> /api/user/v1/accounts
> /api/v1/course_runs/
> /course_team/:course_id/
> /course_rerun/:course_id/
> /videos/:course_id/:video_id/
> /api/learning_sequences/...
> ```
> 
> I think this OEP should clearly define URL expectations like:
> 
> * Whether we should use singular nouns (`/user/:x`) or plural nouns (`/users/:x`)
> * That REST API URLs should always start with `/api/`
> * That new CMS and LMS URLs should not conflict with each other so they can [potentially be merged](https://discuss.openedx.org/t/proposal-combined-headless-lms-cms/18880) without needing major changes to API clients other than domain name.
> * Whether APIs are organized hierarchically using logical components `/api/course/:course_key/teams/:team_id/` or the URLs should strictly map to the underlying organization of the codebase `/api/teams/v1/course/:course_id/:team_id` or some other organizational scheme ?
> * etc.

Hi @bradenmacdonald @feanil ,so the URL conventions sahred by Branden now have a dedicated ADR, docs/decisions/0038-standardize-rest-api-url-structure.rst ([PR #39003](https://github.com/openedx/openedx-platform/pull/39003)), and I've consolidated it into OEP-69 as well named as Convention 14: Standardize REST API URL structure, Moreover Convention 14's Source link will give 404 until PR #39003 merges.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,691 @@
+.. _OEP-69 Open edX REST API Conventions:
+
+OEP-69: Open edX REST API Conventions
```
feanil:
```suggestion
OEP-69: REST API Conventions
```
It's all Open edX in this repo, I don't think we need that in the title.

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,691 @@
...
+#####################################
+
+.. list-table::
+   :widths: 25 75
+
+   * - OEP
+     - :ref:`OEP-69 <OEP-69 Open edX REST API Conventions>`
```
feanil:
```suggestion
     - :ref:`OEP-69 <OEP-69: REST API Conventions>`
```

oeps/best-practices/oep-0069-bp-rest-api-conventions.rst
```diff
@@ -0,0 +1,691 @@
...
+   * - Title
+     - Open edX REST API Conventions
```
feanil:
```suggestion
     - REST API Conventions
```

feanil (commented):
A couple of nits but otherwise this looks good to me.  I'd like ADR 39 to land first but then I think this is good to land as well.

Faraz32123 (approved):

feanil (approved):

PR #815, opened 2026-08-04
docs: frontend branching and publishing ADRs

### LINKED ISSUE #564: Frontend branching and publishing ADRs under OEP-10 and OEP-65
### Description

OEP-10 pins a repository into a release by cutting a `release/RELEASENAME` branch and tagging it, which works for components that are deployed from, or built out of, a checkout of themselves. Under OEP-65 a frontend app is neither, so a release branch in an app repository would be inert: nothing deploys from it and nothing consumes it. Nothing in `openedx-proposals` currently says how such a repository participates in a release instead.

OEP-10 itself changes to distinguish components that a release pins by git reference from those it pins by published version, saying where the record of a pinned version lives and that being pinned by version does not lower a component's obligations under levels of support. None of that is frontend-specific; Python libraries pinned in `edx-platform`'s requirements files work the same way.

Three ADRs follow. OEP-10 ADR 0002, moved here from `frontend-base` ADR 0012, settles what branches a semantically-released frontend repository has, what each of them publishes, how changes travel between them, and how `stable` moves onto a new major. OEP-10 ADR 0003 settles how such a repository takes part in an Open edX release, by published version rather than by a branch cut at release time. OEP-65 ADR 0004 settles how an app repository's `frontend-base` conversion branch takes over as its main branch, that the OEP-10 ADRs govern its branches from that point on, and where the micro-frontend it replaces continues to live.

Part of #560.

Implemented by openedx/openedx-proposals#815

arbrandes:
### Description

OEP-10 pins a repository into a release by cutting a `release/RELEASENAME` branch and tagging it, which works for components that are deployed from, or built out of, a checkout of themselves. Under OEP-65 a frontend app is neither: it is published as an NPM package and composed into a Site at build time. A release branch in an app repository would be inert, since nothing deploys from it and nothing consumes it. Nothing in this repo currently says how such a repository participates in a release instead.

The first change is to OEP-10 itself, which now distinguishes components that a release pins by git reference from those it pins by published version. It says where the record of a pinned version lives, and that being pinned by version does not lower a component's obligations under levels of support. None of this is frontend-specific: Python libraries pinned in `edx-platform`'s requirements files work the same way.

Three ADRs follow from that. Under OEP-10, ADR 0002 (moved here from [frontend-base's ADR 0012](https://github.com/openedx/frontend-base/blob/main/docs/decisions/0012-frontend-branching-strategy.rst)) settles what branches a semantically-released frontend repository has, what each of them publishes, how changes travel between them, and how the `stable` branch moves onto a new major once the main branch is ready for one. ADR 0003 settles how such a repository takes part in an Open edX release, which is by published version rather than by a branch cut at release time.

Under OEP-65, ADR 0004 settles how an app repository's `frontend-base` conversion branch takes over as its main branch, that the OEP-10 ADRs govern its branches from that point on, and where the micro-frontend it replaces continues to live.

### LLM usage notice

Built with assistance from Claude.

Closes openedx/public-engineering#564

### DIFF
--- a/oeps/processes/oep-0010/decisions/0002-frontend-stable-branches.rst
+++ b/oeps/processes/oep-0010/decisions/0002-frontend-stable-branches.rst
@@ -1,0 +1,233 @@
+.. _Frontend Stable Branches:
+
+0002. Frontend Stable Branches
+##############################
+
+Status
+******
+
+**Accepted**
+
+Context
+*******
+
+Semantically-released frontend repositories in the Open edX org need a rallying
+point where new features integrate: a place where developers have enough leeway
+to push the envelope and fix bugs without having the code immediately shipped
+to thousands of users.  That branch is commonly called "master" or "main" in
+projects that follow the "main is unstable" strategy; others call it "next",
+"develop", or "alpha".  What is almost universal is that at least one branch
+serves this purpose.
+
+In contrast, main branches here have historically been deemed stable, with every
+commit that lands expected to be production ready.  Most repositories have no
+"next" branch either.
+
+At the other end of the spectrum, consumers who are not ready for a new major
+version of a package still need fixes for the one they are on.  A single stable
+branch cannot serve them: once it has moved on, the versions they are running
+have nowhere to be patched from.  The structure therefore has to include lines
+that outlive the current stable major, each publishing on its own.
+
+What follows settles which branches such a repository has, what each of them
+publishes, and how changes travel between them.  Which of those lines a given
+Open edX release depends on, and for how long, is settled by
+:ref:`ADR 0003 <Frontend Release Strategy>`.
+
+Decision
+********
+
+The following decisions apply to any frontend repository in the Open edX org
+once it starts publishing to NPM with ``semantic-release``.
+
+1. Main is unstable
+===================
+
+A repository's main branch is deemed unstable. This has two major consequences:
+
+1. Using the main branch in production is not supported;
+2. The DEPR process does not apply to it; breaking changes can land with no
+   warning.
+
+Notably, though, the main branch retains the following properties:
+
+1. New features should be developed for and merged to it before any other
+   branch;
+2. To facilitate collaboration, it is never rebased.
+
+2. Publication of the main branch
+=================================
+
+A repository's main branch is published semantically to NPM on an "alpha"
+prerelease tag, with a monotonically incremented number suffix.  For instance::
+
+  frontend-base@2.0.0-alpha.4
+  frontend-app-instructor-dashboard@2.0.0-alpha.2
+
+3. Stable and maintenance branches
+==================================
+
+Stable code is published from a long-lived ``stable`` branch, which always
+carries the newest stable major.  It owns the NPM "latest" dist-tag and is
+published semantically, with no breaking changes allowed after publication::
+
+  frontend-base@1.0.5
+  frontend-app-instructor-dashboard@1.4.3
+
+When the main branch contains breaking changes over what ``stable`` carries
+and is deemed ready for widespread use, its work graduates onto ``stable`` as
+the next major (see `4. Graduating main onto stable`_).  To continue maintaining
+the previous major, an "n.x" branch is cut from
+its last tag, where "n" is that major's number and the ".x" is literal.  For
+instance, once ``stable`` has moved from the 1.x line to the 2.x line, a ``1.x``
+branch is cut at the last 1.x tag::
+
+  stable    2.0.0, 2.1.0, ...    (dist-tag: latest)
+  1.x       1.4.3, 1.5.0, ...    (dist-tag: 1.x)
+
+These maintenance branches are also published semantically, with no breaking
+changes allowed.  Each one owns the NPM dist-tag matching its own name, which
+is what keeps "latest" pointing at ``stable``. Consumers, for their part,
+select a maintained major by semver range::
+
+  "@openedx/frontend-base": "1.x"
+
+An "n.x" branch accepts both new minors and patches within its major.  To patch a
+minor that the line carrying it has moved past, whether that line is ``stable``
+or an "n.x" branch, an "n.m.x" branch is cut from that minor's last tag, where "m"
+is the minor's number.  These are patch-only: a
+feature landing on one still results in a patch release.  Continuing the example
+above, with 1.5.0 already shipped from ``1.x``, a ``1.4.x`` branch cut at the last
+1.4 tag carries on the 1.4 line::
+
+  stable    2.0.0, 2.1.0, ...    (dist-tag: latest)
+  1.x       1.5.0, 1.6.0, ...    (dist-tag: 1.x)
+  1.4.x     1.4.4, 1.4.5, ...    (dist-tag: 1.4.x)
+
+4. Graduating main onto stable
+==============================
+
+Graduation is the moment ``stable`` stops carrying one major and starts carrying
+the next.  It cannot be a fast-forward: by then ``stable`` holds cherry-picked
+backports that are not on the main branch, so the two have diverged.
+
+Rather than reconcile that divergence, graduation retires the branch.  The
+outgoing line keeps its history under its "n.x" name, and ``stable`` is
+re-pointed at the main branch's tip::
+
+  git push origin stable:1.x
+  git push --force origin main:stable
+
+Cutting the "n.x" branch of `3. Stable and maintenance branches`_ and retiring the
+old ``stable`` are therefore the same act.  Since the second push rewrites a
+published branch, graduation is a deliberate push by a maintainer rather than a
+PR.
+
+5. Backports
+============
+
+All changes, including bug fixes and security patches, should target the main
+branch first.  Once merged, they can be backported to ``stable`` and to the
+appropriate maintenance branches.  This "main first" approach ensures that fixes
+are never lost when a new line is cut: they are part of the main branch's
+history and will naturally flow into future releases.
+
+The only foreseeable scenario where a change may land on ``stable`` or a
+maintenance branch without a corresponding change landing on the main branch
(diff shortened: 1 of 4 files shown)


oeps/processes/oep-0010/decisions/0002-frontend-stable-branches.rst
```diff
@@ -0,0 +1,178 @@
...
+###############################
+
+Status
+******
+
+**Accepted**
+
+Context
+*******
+
```
sarina, line 13: I think it would be useful to provide a definition of what a "frontend" repo means - where does this strategy apply? Does this only apply, as well, to frontend-base repos? Might be worth linking to the frontend-base OEP; could be worth (at least temporarily) linking to a tracker as to status of frontend-base conversion?

oeps/processes/oep-0010/decisions/0002-frontend-release-branches.rst
```diff
@@ -0,0 +1,178 @@
...
+
+  release   2.0.0, 2.1.0, ...    (dist-tag: latest)
+  1.x       1.5.0, 1.6.0, ...    (dist-tag: 1.x)
+  1.4.x     1.4.4, 1.4.5, ...    (dist-tag: 1.4.x)
+
+4. Backports
+============
+
+All changes, including bug fixes and security patches, should target the default
+branch first.  Once merged, they can be backported to ``release`` and to the
```
brian-smith-tcril: I'm wondering if using "default branch" throughout this section makes things less clear than just saying `main`. This is more complicated when we have `master` and `frontend-base` branches and `master` is the default (but we want things to land on `frontend-base`)

oeps/processes/oep-0010/decisions/0003-frontend-release-strategy.rst
```diff
@@ -0,0 +1,152 @@
...
+bound by any of this, though one derived from the template will usually want to
+follow the same pattern.
+
+3. The release branch takes patch-level ranges
+==============================================
+
+On that branch, direct dependencies on apps and on ``frontend-base`` are expressed
+as x-ranges anchored to the minor locked at cutoff.  The branch then accepts
+patches from each maintenance line, but declines minors, which cannot be assumed
+to work against that release's backend::
```
brian-smith-tcril: Reading this without context would lead me to question, "Why can't we trust minor version bumps? Shouldn't a feature that requires a new backend capability land as a breaking change?"

I agree this makes sense for Verawood, but saying minor version bumps can't be assumed to work against a release backend feels like us "doing something wrong" and something we should (at least plan to) fix.

brian-smith-tcril (commented):
Overall this looks good, couple small comments.

oeps/processes/oep-0010/decisions/0002-frontend-stable-branches.rst
arbrandes: Yeah, I agree the intro should specify some of this.  But it doesn't just apply to frontend-base repos.  It's any frontend repo that is semantically released to NPM.

The frontend-base distinction is needed only while there are still frontend-app-* repos where frontend-base didn't fully land, yet.  This is probably worth a standalone "Note:" paragraph.

oeps/processes/oep-0010/decisions/0002-frontend-release-branches.rst
arbrandes: Yeah, good call.

oeps/processes/oep-0010/decisions/0003-frontend-release-strategy.rst
arbrandes: Well, here it's less about trust and doing wrong things, and more about just avoiding the perfectly possible and reasonable situation where a minor bump adds a feature the current backend doesn't support.  I'm inclined to leave it as-is.

oeps/processes/oep-0010/decisions/0002-frontend-stable-branches.rst
sarina: Cool - that was exactly the type of clarification I was seeking. Thanks!

oeps/processes/oep-0010/decisions/0003-frontend-release-strategy.rst
brian-smith-tcril: I'd prefer to have the justification be about not having new features suddenly appear in releases. That's a clear decision about _what_ should be in a release (the features that existed at cut), and shows our pinning aligns with that decision.

arbrandes: Ok, cool.  Reworded.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
```diff
@@ -0,0 +1,147 @@
...
+version rather than by a branch cut at release time.
+
+The landing is what lets that be true immediately rather than eventually.  Because
+the merge's first parent is the conversion branch, the prerelease tags published
+from it stay on the main branch's first-parent line, so ``semantic-release`` picks
+up where the conversion branch left off instead of starting the version history
+over.  A ``stable`` branch off that same line is likewise an ancestor of the main
+branch, so ADR 0002's fast-forward backport path is available from the start.  What
+does change is the repository's ``semantic-release`` configuration, which moves its
+prerelease branch from ``frontend-base`` to the main branch.
```
brian-smith-tcril: This block feels a little confusing to me. Parts that stick out as not clear:
* "The landing is what lets that be true immediately rather than eventually."
  * Not fully clear what "the landing" means here. From context I get it means "landing frontend-base on main/master," but it feels odd here.
* "A ``stable`` branch off that same line is likewise an ancestor of the main branch"
  * This feels a bit odd. I guess I understand how `stable` is an ancestor of `main` instead of a descendant because main will include the `stable` commits and _also_ anything that hasn't yet landed on `stable` yet, but it feels like it almost implies "changes from `stable` go into `main`" which isn't the plan.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
```diff
@@ -0,0 +1,147 @@
...
+
+3. The micro-frontend line continues on legacy-mfe
+==================================================
+
+Immediately before the landing, a long-lived ``legacy-mfe`` branch is cut from the
```
brian-smith-tcril: I think we might just need a better term for "the landing." No ideas immediately come to mind but I'll comment if I think of any.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
```diff
@@ -0,0 +1,147 @@
...
+micro-frontend, and releases that still ship it, need a line where it remains
+buildable and patchable, and ``legacy-mfe`` is the only one left.
+
+Consequences
+************
+
+Every converted repository comes out of this with a branch nobody develops on that
+somebody may still have to patch, for as long as a supported release includes its
+micro-frontend.  That is a real cost, and it is the price of not stranding
+operators mid-conversion.
```
brian-smith-tcril: "a branch nobody develops on that somebody may still have to patch" could use clarification. Maybe "a legacy branch that no longer sees active development" or something like that for the "a branch nobody develops on" part, and maybe something about operators still (possibly? likely? definitely?) using it for "that somebody may still have to patch."

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Release tooling needs to know about the split.  A converted repository flips
+``openedx.org/release`` to null per ADR 0003, so nothing should be cut from its main
+branch, but a release that still includes the micro-frontend has to be cut from
+``legacy-mfe`` instead.  Tooling that assumes a repository's release branch comes
+from its default branch will need attention.
```
brian-smith-tcril: Or we just manually cut because this should only last 1 release cycle per `frontend-app` right?

brian-smith-tcril (commented):
Overall this is looking great! Couple notes about sections that felt a little unclear, looking forward to hearing your thoughts!

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
arbrandes: Fair enough.  I think there's simply too many details.  I'll try to keep it more straight to the point, or maybe even remove the paragraph entirely.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
arbrandes: We could, yes, but I think it's a just a matter of pointing the catalog file to "legacy-mfe" until MFEs are properly DEPRed and removed.  Better than having to cut manually again, if the DEPRs don't happen fast enough.

arbrandes: Either way, I'll make it clearer.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
arbrandes: Let's just call it "merging".  "Landing" is too cute.

oeps/architectural-decisions/oep-0065/decisions/0004-landing-frontend-base.rst
arbrandes: Removed the paragraph.  It wasn't necessary.

brian-smith-tcril (approved):

